PR #2, opened 2018-08-02
Basic lifecycle test for GuessNumber App

mvanderh:
Test for happy path of a Counterfactual app:
1. Deploy & fund multisig with A and B as signers
2. Deploy GuessNumber app
3. Deploy StateChannel via Registry
4. `setState` on StateChannel with a terminal state
5. `setResolution` on StateChannel
6. `setNonce` and `finalizeNonce` on NonceRegistry
7. `executeStateChannelConditionalTransfer` on ConditionalTransfer from multisig
8. Confirm that A and B's ETH balances are correct

Misc:
* Helper class for dealing with Multisigs
* New function in Utils: `deployContractViaRegistry`
* Fix security issue in MinimumViableMultisig.setup
* Fix delegatecall bug in `executeStateChannelConditionalTransfer`

TODO: 
- ~Fix tests~
- ~Merge in `develop`~

### DIFF
--- a/packages/contracts/contracts/MinimumViableMultisig.sol
+++ b/packages/contracts/contracts/MinimumViableMultisig.sol
@@ -34,7 +34,8 @@ contract MinimumViableMultisig {
 
   /// @notice Contract constructor
   /// @param owners An array of unique addresses representing the multisig owners
-  constructor(address[] owners) public {
+  function setup(address[] owners) public {
+    require(_owners.length == 0); // Contract hasn't been set up before
     _owners = owners;
     for (uint256 i = 0; i < owners.length; i++) {
       require(owners[i] != 0);
@@ -77,7 +78,7 @@ contract MinimumViableMultisig {
   /// @param value The amount of ETH being sent in the transaction
   /// @param data Any calldata being sent along with the transaction
   /// @param operation An `Operation` referring to the use of `CALL` or `DELEGATECALL`
-  /// @return A unique hash that owners are expected to sign and submit to `execTransaction`
+  /// @return A unique hash that owners are expected to sign and submit to `multisigExecTransaction`
   function getTransactionHash(
     address to,
     uint256 value,
--- a/packages/contracts/contracts/Registry.sol
+++ b/packages/contracts/contracts/Registry.sol
@@ -28,7 +28,9 @@ contract Registry {
   /// @notice Deploy a contract and add a mapping from counterfactual address to deployed address to storage
   /// @param initcode Contract bytecode concatenated with ABI encoded constructor arguments
   /// @param salt Unique salt used in address computation that allows multiple contracts with the same initcode to exist.
-  function deploy(bytes initcode, uint256 salt) public {
+  function deploy(bytes initcode, uint256 salt)
+    public
+  {
     address deployed;
     bytes32 ptr = cfaddress(initcode, salt);
 
--- a/packages/contracts/test/helpers/multisig.ts
+++ b/packages/contracts/test/helpers/multisig.ts
@@ -1,61 +1,103 @@
+import {
+  deployContract,
+  highGasLimit,
+  signMessageBytes
+} from "@counterfactual/test-utils";
 import * as ethers from "ethers";
 
-export const Operation = Object.freeze({
-  Call: 0,
-  Create: 2,
-  Delegatecall: 1
-});
-
-/**
- * Executes a transaction from the multisig to itself according to
- * to the given data.
- *
- * @param data is the contract abi to invoke the correct function with args.
- * @param multisig is the multisig sending and receiving the tx.
- * @param wallets is the array of wallets signing off on the tx.
- * @param op is the Operation to use, i.e., Call, Delegatecall, or Create.
- */
-export async function executeTxData(data, toAddr, multisig, wallets, op) {
-  const value = 0;
-
-  const transactionHash = await multisig.getTransactionHash(
-    toAddr,
-    value,
-    data,
-    op
-  );
-
-  const signatures = sign(transactionHash, wallets);
-  return multisig.execTransaction(
-    toAddr,
-    value,
-    data,
-    op,
-    signatures.v,
-    signatures.r,
-    signatures.s,
-    {
-      gasLimit: 4712388
-    }
-  );
+const enum Operation {
+  Call = 0,
+  Delegatecall = 1
 }
 
-export function sign(data, wallets) {
-  const sortedWallets = wallets.slice().sort((w1, w2) => {
-    return w1.address < w2.address ? -1 : w1.address === w2.address ? 0 : 1;
-  });
-  const v: number[] = [];
-  const r: string[] = [];
-  const s: string[] = [];
-  sortedWallets.forEach(wallet => {
-    const sig = new ethers.SigningKey(wallet.privateKey).signDigest(data);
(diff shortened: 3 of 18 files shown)


packages/contracts/contracts/MinimumViableMultisig.sol
```diff
@@ -34,7 +34,8 @@ contract MinimumViableMultisig {
 
   /// @notice Contract constructor
   /// @param owners An array of unique addresses representing the multisig owners
-  constructor(address[] owners) public {
+  function setup(address[] owners) public {
```
armaniferrante: If we go with this change, we should add a comment explaining how it's meant to be deployed through `ProxyFactory` . Otherwise, we have to incur two transactions to setup the multisig instead of one.

packages/contracts/test/helpers/multisig.ts
```diff
@@ -1,61 +1,103 @@
...
+    if (!this.contract) {
+      throw new Error("Must deploy Multisig contract first");
+    }
+    return this.contract.address;
+  }
+
+  public async deploy(
+    providerOrSigner: ethers.ethers.Wallet | ethers.types.MinimalProvider
+  ) {
+    const contract = await deployContract(
```
armaniferrante: Should we do this through the ProxyFactory instead?

packages/contracts/contracts/Registry.sol
```diff
@@ -28,7 +28,9 @@ contract Registry {
   /// @notice Deploy a contract and add a mapping from counterfactual address to deployed address to storage
   /// @param initcode Contract bytecode concatenated with ABI encoded constructor arguments
   /// @param salt Unique salt used in address computation that allows multiple contracts with the same initcode to exist.
-  function deploy(bytes initcode, uint256 salt) public {
```
armaniferrante: I think we should only break up into multiple lines like this if we go over a line count. This just makes it harder to read when the line is short.

packages/contracts/test/integration/commitReveal.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+    provider
+  ),
+  // 0xb37e49bFC97A948617bF3B63BC6942BB15285715
+  new ethers.Wallet(
+    "0x4ccac8b1e81fb18a98bbaf29b9bfe307885561f71b76bd4680d7aec9d0ddfcfd",
+    provider
+  )
+];
+
+const computeStateHash = (
```
armaniferrante: I'm not sure how others feel, but I think we should avoid this type of syntax for standalone functions and instead use

`function computeStateHash(...) {...}`, passing in `A` and `B` if you want them.

The arrow syntax is useful for inline anonymous functions w/ binding, but closing over global state like this makes things harder to understand when you read it for the first time.

packages/contracts/test/integration/commitReveal.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  beforeEach(async () => {
+    CommitRevealApp.link("StaticCall", StaticCall.address);
+
+    StateChannel.link("Signatures", Signatures.address);
+    StateChannel.link("StaticCall", StaticCall.address);
+    StateChannel.link("Transfer", Transfer.address);
+  });
+
+  it("should complete a full lifecycle", async function() {
```
armaniferrante: This test is ~142 lines long. 

We can break it up and make it more readable and understandable.

Easy hack is to just take  all the numbered comments and make a function out of it where the name of the func is the comment. Then the code is self documenting.

packages/contracts/test/integration/commitReveal.spec.ts
mvanderh: Agreed

packages/contracts/contracts/MinimumViableMultisig.sol
mvanderh: 👍

packages/contracts/test/helpers/multisig.ts
mvanderh: Not necessary for this test IMO

packages/contracts/contracts/Registry.sol
mvanderh: Hmm, wouldn't the linter have fixed it if that were true?

packages/contracts/test/integration/commitReveal.spec.ts
mvanderh: The test is definitely a bit too long. I've written some test tooling that's going to reduce the # of lines here.

I don't like the extract-to-function approach, it removes explicitness.

armaniferrante: Accepting since these aren't critical issues to block on if we disagree, but they are interesting nonetheless,

Can you explain what you mean about explicitness?

The extract-to-function approach is only a heuristic. Let me give some context into how I think about this.

The point is that, when I'm reading tests,  I'm often trying to understand what a particular function or class is doing. I want a high level overview and so I want to operate at one abstraction level, and only step down abstraction levels as needed. I don't want to read the implementation details of how a contract is deployed, or what `setState` means.

I'm often just trying to figure out, what tests are doing? What are they not doing? How do I use this class? 

These are useful questions from the perspective of a new person reading the code. However, when we have monolithic tests like this, it makes answering these questions more difficult.

Take for example, the lifecycle test in the machine:
```
it("should exercise the lifecycle of a state channel", async () => {
  //given
  let [walletA, walletB] = createWallet();

  //when
  await setup(walletA, walletB);
  await makeDeposits(walletA, walletB);
  await playTtt(walletA, walletB);

  // then
  validateSystem(walletA, walletB);
  await gotoChain();
  await validateBlockchain();
}
```

This is a huge test with lots of code, but when reading the test function, you can immediately understand the entirety of what it does. If you care about how TTT is played, you can just go into that function if you want.

Generally, I lean more towards the direction of uncle bob's "CleanCode" style. This makes for a lot of helper functions and classes (via SRP), but makes code read more like English.

packages/contracts/contracts/Registry.sol
armaniferrante: That just means the linter isn't configured to fix this.

packages/contracts/test/integration/commitReveal.spec.ts
mvanderh: I see your point. 

However, to me it feels like there's a context-switching cost to jumping between functions. I find myself constantly having to remember which stage of the test I'm in and which data is currently being operated on. 

By "explicitness" I mean transparency of the code that's executed. More explicit tests might have a steeper learning curve at first, but I find they are better for understanding the logic being tested more deeply. 

That said, I'm not certain. I'll try understandable-at-a-glance-style tests in a later PR, and compare.

packages/contracts/contracts/Registry.sol
mvanderh: It does split up other functions  across multiple lines though. I think it's just a matter of threshold.

packages/contracts/test/integration/commitReveal.spec.ts
armaniferrante: Fwiw, my philosophy here isn't just for test. It applies to all code.

mvanderh: Yeah, and in general i agree with you. But there's a time and a place for both philosophies IMO

PR #9, opened 2018-08-03
Rename contract interface methods

cf19drofxots:
Renames:
- reducer => ~~reduce~~ applyAction
- resolver => resolve

### DIFF
--- a/packages/apps/contracts/CommitRevealApp.sol
+++ b/packages/apps/contracts/CommitRevealApp.sol
@@ -97,7 +97,7 @@ contract CommitRevealApp {
     return abi.encode(nextState);
   }
 
-  function resolver(AppState state, Transfer.Terms terms)
+  function resolve(AppState state, Transfer.Terms terms)
     public
     pure
     returns (Transfer.Details)
--- a/packages/apps/contracts/CommitRevealApp.sol
+++ b/packages/apps/contracts/CommitRevealApp.sol
@@ -58,7 +58,7 @@ contract CommitRevealApp {
     return Player.CHOOSING;
   }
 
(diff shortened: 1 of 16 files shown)


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
note that this also renames the `resolver` from the registry, which is (unfortunately) different from the resolver used in apps

cf19drofxots:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll good catch - reverted

packages/apps/contracts/CommitRevealApp.sol
```diff
@@ -58,7 +58,7 @@ contract CommitRevealApp {
     return Player.CHOOSING;
   }
 
-  function reducer(AppState state, Action action)
+  function reduce(AppState state, Action action)
```
mvanderh: Is there a reason we don't just call this `applyAction` or something? Is it just to use Redux terminology?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I prefer `applyAction` over `reduce`

mvanderh: @snario @armaniferrante @ebryn

armaniferrante: As mentioned offline, I prefer `applyAction` over reduce as well. Reduce/reducer confused me when I first read it.

cf19drofxots:
@mvanderh @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll @armaniferrante any other concerns with this renaming?

PR #11, opened 2018-08-03
Tooling for tests v1 (StructAbiEncoder, computeHash functions)

mvanderh:
- New convenience class: StructAbiEncoder
- Extract helper functions: computeStateHash and computeActionHash.
- utils.ts: Rename some constants and remove some unused functions.
- Migrated commitReveal.spec.ts


packages/test-utils/src/structAbiEncoder.ts
```diff
@@ -0,0 +1,33 @@
...
+      }
+      const parts = definition.split(" ");
+      if (parts.length !== 2) {
+        throw new Error(
+          `Invalid struct field. Expected '[type] [name]', got '${definition}'`
+        );
+      }
+      definitions.push(definition);
+    }
+    const encoding = `tuple(${definitions.join(", ")})`;
```
cf19drofxots: It would be cool to extend this in the future to support encodings with nested tuples.

packages/contracts/test/integration/commitReveal.spec.ts
```diff
@@ -104,27 +82,25 @@ describe("CommitRevealApp", async () => {
...
+    const startBalanceB = await bob.getBalance();
+
     // 1. Deploy & fund multisig
-    const multisig = new Multisig([A.address, B.address]);
+    const multisig = new Multisig([alice.address, bob.address]);
     await multisig.deploy(masterAccount);
     await masterAccount.sendTransaction({
       to: multisig.address,
-      value: Utils.UNIT_ETH.mul(2)
+      value: ethers.utils.parseEther("2")
```
cf19drofxots: Passing thought: maybe we can have a wrapper around this to abstractly indicate 2 eth being used instead of having a utility function needing to parse it?

packages/test-utils/src/utils.ts
```diff
@@ -91,7 +107,7 @@ export const setupTestEnv = (web3: any) => {
   return { provider, unlockedAccount };
 };
 
-export function signMessage(message, wallet): [number, string, string] {
+function signMessage(message, wallet): [number, string, string] {
```
cf19drofxots: Why privatize these? Are they not being used anywhere else?

mvanderh: Yup, they're nont

packages/contracts/test/integration/commitReveal.spec.ts
mvanderh: What do you mean? `EtherAmount(2)`?

packages/test-utils/src/structAbiEncoder.ts
mvanderh: Yeah, shouldn't be hard to do

packages/contracts/test/integration/commitReveal.spec.ts
cf19drofxots: Yeah something like that - def not a blocker by any means tho

mvanderh: I'm not sure that's an abstraction worth doing. Esp if all the other Ethereum tooling uses raw integers.

cf19drofxots: Yeah it may add unneeded overhead to existing usable patterns. Again, just a passing thought - not worth reading too much into

snario: Maybe just a shorthand util `const str2eth = x => ethers.utils.parseEther`

packages/contracts/test/integration/commitReveal.spec.ts
```diff
@@ -104,27 +82,25 @@ describe("CommitRevealApp", async () => {
...
+      token: ZERO_ADDRESS
     };
     // 2. Deploy CommitRevealApp app
-    const appContract = await Utils.deployContract(
-      CommitRevealApp,
-      masterAccount
-    );
-    app = {
+    const appContract = await deployContract(CommitRevealApp, masterAccount);
+    const app = {
```
snario: We could write a type for this. addr is a 20 byte string, reduce could be a ethers.Interface.FunctionDefinition etc

mvanderh: Already on it

PR #14, opened 2018-08-07
New contract tools: Abstract Contract and Contract

mvanderh:
- Simple `Contract` class which wraps `ethers.Contract` with added fields for Counterfactual deployment.
- `AbstractContract` class which handles: load Truffle artifact & link, retrieve singleton instance, deploy contract (+via registry), connect to contract.

ALSO:
- Created index.ts in test-utils for exporting
- Moved Multisig to test-utils
- Updated `ethers` to `4.0.0-beta.11` 
- Migrated commitReveal test


packages/contracts/test/unit/conditionalTransfer.spec.ts
```diff
@@ -64,21 +64,21 @@ contract("ConditionalTransfer", (accounts: string[]) => {
...
+            data: Utils.ZERO_ADDRESS
           }
         ]
       );
 
       await delegateProxy.functions.delegate(
         ct.address,
         tx,
-        Utils.highGasLimit
+        Utils.HIGH_GAS_LIMIT
```
snario: Any way to avoid using this so often?

snario:
On first glance, `AbstractContract` looks an awful lot like `TruffleContract`. What's different about it except that it uses the registry?

packages/contracts/test/unit/conditionalTransfer.spec.ts
mvanderh: The new version of Ethers uses `estimateGas` so we can remove HIGH_GAS_LIMIT in most places. I've found we still need it for special cases though e.g. deploying via registry.

mvanderh:
AbstractContract is an idea that @ricmoo and i discussed. It means just the result of contract compilation i.e. the ABI and bytecode. This is different from an on-chain instance of a smart contract.

I like it because it distinguishes between the two definitions of "contract".

I thought I'd give it a shot and see if it makes sense, but you might be right that it's not a worthwhile abstraction of TruffleContract.

packages/contracts/test/helpers/cfhelpers.ts
```diff
@@ -1,221 +0,0 @@
-import * as ethers from "ethers";
```
armaniferrante: 👍 for getting rid of this file.

packages/test-utils/src/contract.ts
```diff
@@ -0,0 +1,111 @@
...
+    let { bytecode } = truffleContract;
+    if (links) {
+      for (const name of Object.keys(links)) {
+        const library = links[name];
+        if (!library.deployedAddress) {
+          throw new Error("Library must have a deployed address");
+        }
+        const regex = new RegExp(`__${name}_+`, "g");
+        const addressHex = library.deployedAddress.replace("0x", "");
+        bytecode = bytecode.replace(regex, addressHex);
```
mvanderh: Truffle-contract doesn't expose the linkBytecode util function, so I had to copy the source code. 

I can use the `link` method on the Truffle artifact, but that would couple AbstractContract tightly with TruffleContract, which we don't want.

packages/test-utils/src/stateChannel.ts
```diff
@@ -1,6 +1,26 @@
...
+      contract.interface.functions.applyAction.sighash,
+      contract.interface.functions.resolve.sighash,
+      contract.interface.functions.turnTaker.sighash,
+      contract.interface.functions.isStateFinal.sighash
+    );
+  }
+
+  constructor(
+    readonly address: string,
+    readonly applyAction: string,
```
snario: Consider storing the `ethers.Interface.FunctionDefinition` instead of the sighash string here?

packages/test-utils/src/utils.ts
```diff
@@ -23,7 +23,7 @@ export {
 
 // https://github.com/ethers-io/ethers.js/pull/225
 // @ts-ignore
-ethers.types.BigNumber.prototype.equals = function(x): boolean {
+ethers.utils.BigNumber.prototype.equals = function(x): boolean {
```
snario: lol @ricmoo stop moving this around

packages/contracts/test/integration/commitReveal.spec.ts
```diff
@@ -99,19 +104,20 @@ describe("CommitReveal", async () => {
...
-      turnTaker: appContract.interface.functions.getTurnTaker.sighash,
+      turnTaker: appContract.interface.functions.turnTaker.sighash,
       isStateFinal: appContract.interface.functions.isStateFinal.sighash
     };
+    const registry = Registry.getSingleton(masterAccount);
 
     // 3. Deploy StateChannel
-    const { cfAddr, contract: stateChannel } = await deployContractViaRegistry(
-      StateChannel,
+    const stateChannel = await StateChannel.deployViaRegistry(
```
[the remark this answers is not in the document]
mvanderh: "Via" is defined as "by way of; through.", so it means the same as "through". I prefer them both to "using" though.

mvanderh:
@ricmoo AbstractContracts have a "deployed" (non-Counterfactual) address when deployed in a Truffle migration. It's a bit messy because this makes AbstractContracts less... abstract. I'm wondering if that's a problem and if so, what a good solution would be.

The problem isn't an inaccurate gas price but rather gas limit i.e. it's hard to figure out how much gas a `delegatecall` is going to cost.

Also I agree with your case against a global singleton for the Registry. It would make testing complicated.

packages/test-utils/src/contract.ts
```diff
@@ -0,0 +1,111 @@
...
+  readonly contractName: string;
+  readonly abi: any[];
+  readonly bytecode: string;
+  readonly networks: { [networkName: string]: { address: string } };
+}
+
+export class Contract extends ethers.Contract {
+  public salt?: string;
+  public cfAddress?: string;
+  public registry?: Contract;
```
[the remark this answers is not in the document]
mvanderh: Refer to ricmoo's comment

packages/test-utils/src/contract.ts
```diff
@@ -0,0 +1,111 @@
...
+import { HIGH_GAS_LIMIT } from "./utils";
+
+const { solidityKeccak256 } = ethers.utils;
+
+export interface TruffleContract {
+  readonly contractName: string;
+  readonly abi: any[];
+  readonly bytecode: string;
+  readonly networks: { [networkName: string]: { address: string } };
+}
```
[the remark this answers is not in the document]
mvanderh: It seems `typescript-typings` hasn't been updated since the move to monorepo.

packages/test-utils/src/stateChannel.ts
mvanderh: That precludes the ability to pass the App object verbatim into a contract call, with no great benefit IMO e.g. `stateChannel.functions.setResolution(app, state, terms)`

packages/test-utils/src/contract.ts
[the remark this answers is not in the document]
snario: I mean, we should add this into that package and import the typing for `TruffleContract` vs declaring it here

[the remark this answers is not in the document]
mvanderh: What i meant to say was it broke because it hasn't been updated since move to monorepo, sorry for confusion 😅

PR #22, opened 2018-08-13
StateChannel convenience class (and other conveniences)

mvanderh:
StateChannel convenience class
- Instantiate from signers, multisig, app and terms
- keeps track of appStateNonce
- setState() method
- setResolution() method

**ALSO**
+ Move contracts-specific test tools from `test-utils/` to `contracts/` (avoids circular dependency, and makes more sense)
+ Change `signMessageBytes`  to accept vararg of signers instead of array
+ App contract interface
   + Rename all App functions to get turn taker to `getTurnTaker`
   + Rename App function `isStateFinal` to `isStateTerminal`
   + Enforce App function names by hardcoding 
+ Remove "App" convenience class, make it an interface instead
+ Remove `StructAbiEncoder` class, instead just expose helper functions.
+ Make CommitReveal test more readable by extracting functions


packages/contracts/test/integration/commitReveal.spec.ts
```diff
@@ -61,162 +62,176 @@ enum Player {
...
-  const StateChannel = AbstractContract.loadBuildArtifact("StateChannel", {
-    StaticCall,
-    Signatures,
-    Transfer
+async function createMultisig() {
+  const multisig = new Multisig([alice.address, bob.address]);
+  await multisig.deploy(masterAccount);
+  await masterAccount.sendTransaction({
+    to: multisig.address,
+    value: parseEther("2")
```
cf19drofxots: Maybe this helper function can be generalized to take an eth amount?

cf19drofxots: Err I meant this comment for `createMultisig` so that the eth amount can be passed into `parseEther`

armaniferrante: Yea we could do that. The caller can pass in the variable, and we can use the same one to check `endBalanceA.sub(startBalanceA).should.be.bignumber.eq(parseEther("2"));` at the end of the test.

armaniferrante: In addition, it'd be changed the function signature to be:

`async function createMultisig(alice: PeerBalance, bob: PeerBalance)`, where

`interface PeerBalance { peer: Address; balance: number}`.

Then we could reuse this in other tests, e.g., in the machine.

cf19drofxots: Yeah exactly

packages/contracts/utils/structEncoding.ts
```diff
@@ -0,0 +1,24 @@
+import * as ethers from "ethers";
+
+export function structDefinitionToEncoding(structDefinition: string): string {
```
armaniferrante: The name of this is a little confusing when coupled with  `encodeStruct`, since this function doesn't do the ABI encoding.

Maybe we should change the name to `canonicalizeStruct` or something else.

packages/contracts/test/integration/commitReveal.spec.ts
mvanderh: Good suggestions. 

Side note: I believe every test should be idempotent, and therefore that each test should generate new Ethereum wallets with balance zero. This would make `startBalance*` obsolete. Likely I'll do this in a later PR.

packages/contracts/utils/stateChannel.ts
```diff
@@ -0,0 +1,191 @@
...
+  ERC20
+}
+
+export interface TransferTerms {
+  assetType: AssetType;
+  limit: ethers.utils.BigNumber;
+  token?: string;
+}
+
+export interface App {
```
armaniferrante: We should really start sharing more code between the machine and this. See https://github.com/counterfactual/machine/blob/master/src/middleware/cf-operation/types.ts#L39

armaniferrante: I also think using explicit types instead of `string` everywhere helps readability, a lot, especially for new readers.

mvanderh: I agree and I agree. This is why @cf19drofxots' work on linking packages is so important.

packages/contracts/utils/structEncoding.ts
mvanderh: Yup, don't like the name either. `abiEncodingForStruct`?

PR #30, opened 2018-08-23
Exportable contracts step 1

mvanderh:
AbstractContract:
- Keep record of contracts deployed on each network
- Load networkId from Ethereum provider
- User networkId to read deployed contract address
- Link libraries at deploy time

Need this to reference contracts deployed on different networks e.g. testnet, mainnet, local dev net.

Still figuring out what makes the most sense for deploying contracts to local test environments in external packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   "devDependencies": {
     "@counterfactual/configs": "https://github.com/counterfactual/configs.git",
     "@counterfactual/run-with-ganache": "https://github.com/counterfactual/run-with-ganache.git",
-    "@counterfactual/typescript-typings": "../typescript-typings",
+    "@counterfactual/typescript-typings": "https://github.com/counterfactual/typescript-typings.git",
     "@types/bignumber.js": "4.0.2",
     "@types/mocha": "5.2.0",
     "@types/node": "10.1.3",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
   "devDependencies": {
     "@counterfactual/configs": "https://github.com/counterfactual/configs.git",
     "@counterfactual/run-with-ganache": "https://github.com/counterfactual/run-with-ganache.git",
-    "@counterfactual/typescript-typings": "https://github.com/counterfactual/typescript-typings.git",
+    "@counterfactual/typescript-typings": "../typescript-typings",
```
mvanderh: For testing, will remove

utils/deploy.ts
```diff
@@ -0,0 +1,30 @@
+export function deployTruffleArtifacts(
```
armaniferrante: Nice I assume this is what we'll use in the machine to deploy the contracts?

mvanderh: Something like this, but with a `provider` instead of a `deployer` akin to https://github.com/0xProject/0x-monorepo/blob/a21bfc5f2c9e7e8b660f9ea6671e1dfd229ab75d/packages/contract-wrappers/test/global_hooks.ts

`deployTruffleArtifacts` was mostly an attempt for me to figure out whether this worked at all

PR #35, opened 2018-09-02
rename nonce -> expectedNonce in isFinalized

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR renames the `nonce` parameter in `NonceRegistry::isFinalized` to be more specific.


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Because the previous check (`table[key].finalizesAt <= block.number`) throws an error as well. I see this function as a conjunction of two checks

- `table[key].finalizesAt <= block.number`
- `nonce == expectedNonce`

so we should use the same behaviour for both

PR #36, opened 2018-09-02
Support packed nonce

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR allows dependency nonces to be treated in a "packed" format (https://gist.github.com/IIIIllllIIIIllllIIIIllllIIIIllllIIIIll/af044ce89cdd6839453afcffa6c43302).

### DIFF
--- a/contracts/NonceRegistry.sol
+++ b/contracts/NonceRegistry.sol
@@ -36,12 +36,69 @@ contract NonceRegistry {
     return state.nonce == nonce;
   }
 
+  /// @notice Return the N highest-order bits from the input.
+  /// @param input A uint256 treated as a bitfield from which to get the high-order bits
+  /// @param N the number of bits to get from input
+  /// @return A uint256 where the N highest-order bits are the same as in input, and
+  /// the other bits are 0
+  function getFirstNBits(uint256 input, uint8 N) public pure returns (uint256) {
+    uint256 nOnes = uint256(2) ** N - 1;
+    uint256 mask = nOnes << (uint8(256) - N); // uint8(256) == uint8(0)
+    return uint256(bytes32(input) & bytes32(mask));
+  }
+
+
+  /// @return Whether the Nth highest-order bit in input is set
+  function bitSet(uint256 self, uint8 index) public pure returns (bool) {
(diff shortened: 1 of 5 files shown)


contracts/NonceRegistry.sol
```diff
@@ -33,28 +33,81 @@ contract NonceRegistry {
...
-    return state.nonce == nonce;
+    require(
+      state.nonce == expectedNonce,
+      "Nonce is not equal to expectedNonce"
+    );
+    return true;
   }
 
-  /// @notice Set a nonce in the mapping and triggers the timeout period to begin
+  /// "Get First N Bits" from
```
snario: Can you stick with the natstring format

test/unit/nonceRegistry.spec.ts
```diff
@@ -17,6 +27,28 @@ contract("NonceRegistry", accounts => {
...
+    );
+  });
+
+  it("getFirstNBits works for 9", async () => {
+    const ret = await registry.functions.getFirstNBits(
+      "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+      9
+    );
+    // 0x8 == 0b1000
+    console.log(ret);
```
snario: Shouldn't be included

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
When was nonce finalization without a timeout being used?

armaniferrante:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll I don't think nonce finalization without a timeout was being used anywhere. However, to get a POC of #59 to work, we added it to the protocols just so that the multisend transaction would go through on-chain. 

@snario is this what you were referring to?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
closed by accident

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
If we added the finalization as a hack with a known reason, let's revert the removal from this PR and remove it / discuss its removal separately. The purpose of a dependency nonce implies that its unanimous finalization is not needed in an install and uninstall protocol.

PR #38, opened 2018-09-06
[WIP] Global StateChannel Adjudicator Contract

snario:
To save on gas costs, we can have a global contract for dispute resolution. This is the first step towards that. Currently a work in progress due to this error which is expected.

```console
contracts/StateChannelAdjudicator.sol:255:26:
CompilerError: Stack too deep, try removing local variables.
        computeStateHash(_id, keccak256(appState), nonce, timeout),
```


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Can you rename it to `StateChannelAppAdjudicator`

PR #43, opened 2018-09-21
nonceIsSet

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
add a function to check if a nonce has been set

### DIFF
--- a/contracts/NonceRegistry.sol
+++ b/contracts/NonceRegistry.sol
@@ -94,6 +94,10 @@ contract NonceRegistry {
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


contracts/NonceRegistry.sol
```diff
@@ -94,6 +94,10 @@ contract NonceRegistry {
     return true;
   }
 
+  function nonceIsSet(bytes32 key) public returns (bool) {
```
snario: Preferred naming would be `isNonceSet`

snario: Also should be a `view` function

PR #48, opened 2018-09-27
Rename `StateChannel.sol`

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Proposed new names:

- `AppDefinition`: a contract with `applyAction, resolve, getTurnTaker, isStateTerminal`
- `AppInstance`: a contract that holds app state storage in a dispute. The `AppInstance.sol` contract also happens to contain framework code for dispute resolution
- "app": either an `AppDefinition` or `AppInstance` depending on context


utils/multisig.ts
```diff
@@ -85,6 +85,9 @@ export class Multisig {
     args: any[],
     signers: ethers.Wallet[]
   ): Promise<ethers.types.Transaction> {
+    if (toContract.interface.functions[funcName] === undefined) {
+      throw new Error(`Tried to execute delegateCall to ${funcName} but no such function exists on the target contract`)
```
snario: The docstring & this comment should be consistent in using the word call vs delegatecall.

contracts/AppInstance.sol
```diff
@@ -6,12 +6,12 @@ import "./lib/StaticCall.sol";
 import "./lib/Transfer.sol";
 
 
-/// @title StateChannel - A generalized state channel application contract
+/// @title AppInstance - A contract that defines a state channel application
```
cf19drofxots: So by this terminology, deploying a state channel application gives an instance of an `AppInstance`?

Why not just name it `App` then? Then it's more intuitive to have an _instance_ of an app?

contracts/delegateTargets/ConditionalTransfer.sol
```diff
@@ -26,17 +26,17 @@ contract ConditionalTransfer is Conditional {
     tx.execute();
   }
 
-  /// @notice Execute a fund transfer for a state channel in a finalized state
+  /// @notice Execute a fund transfer for a state channel app in a finalized state
   /// @param key The key in the nonce registry
   /// @param expectedNonce The expected nonce in the nonce registry
-  /// @param channelCfAddress Counterfactual address of the state channel contract
+  /// @param appCfAddress Counterfactual address of the state channel contract
```
cf19drofxots: `...of the app's contract`?

contracts/AppInstance.sol
mvanderh: I think it's to distinguish between ambiguous meanings for "App": Instance and Definition

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Yeap what @mvanderh said, I feel like having separate `App` and a `AppDefinition` contracts is more confusing

cf19drofxots: Fair enough

mvanderh:
I like this a lot! Big improvement on earlier naming.

Random idea: Maybe we should name the "AppInstance" something with "Dispute" in it? Since it's only ever used to resolve disputes on chain that would make sense to me.

PR #49, opened 2018-09-28
disallow signatures on message-based authentication

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
this is mainly a sanity check - callers shouldn't used both at the same time


snario:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll Can you apply the changes to pass the tests?

PR #51, opened 2018-10-03
Update to ethers 4.0.1

snario:


utils/multisig.ts
```diff
@@ -84,9 +86,11 @@ export class Multisig {
     funcName: string,
     args: any[],
     signers: ethers.Wallet[]
-  ): Promise<ethers.types.Transaction> {
+  ): Promise<any> {
```
cf19drofxots: Why `any` here? And below on line 110.

snario: There isn't an exposed ethers type for transacations now

PR #52, opened 2018-10-04
cleanup

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- rename `appsDefinitions` -> `appDefinitions`
- move `ETHBalanceRefundApp.sol` to `appDefinitions`
- rename `tx` to `txn` since `tx` is a global variable in solidity
- rename `itx` to `txn`


contracts/lib/Transfer.sol
```diff
@@ -50,19 +50,19 @@ library Transfer {
 
   /// @notice Verifies whether or not a `Transfer.Transaction` meets the terms set by a
   /// `Transfer.Terms` object based on the limit information of how much can be transferred
-  /// @param tx A `Transfer.Transaction` struct
+  /// @param txn A `Transfer.Transaction` struct
```
cf19drofxots: This also appears in https://github.com/counterfactual/contracts/blob/87397d6cd8ffbcb468c78c92d83d51d8c0f09daf/contracts/delegateTargets/ConditionalTransfer.sol#L21

snario: I had used `itx` below (kind of like "Internal Transaction").

PR #58, opened 2018-10-11
[PoC] contract design with uninstall fix and root nonce 

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


contracts/delegateTargets/ConditionalTransfer.sol
```diff
@@ -27,23 +27,29 @@ contract ConditionalTransfer is Conditional {
...
+    uint256 rootNonceExpectedValue,
     bytes32 appCfAddress,
     Transfer.Terms terms
   )
     public
   {
     require(
-      NonceRegistry(nonceRegistry).isFinalized(key, expectedNonce),
-      "State Channel nonce is either not finalized or finalized at an incorrect nonce"
+      !NonceRegistry(nonceRegistry).isFinalized(uninstallKey, 1),
```
snario: 1 is being used to represent "uninstalled" correct? `null` is uninitialized, `0` is open, `1` is closed? If so should this be a `bool`?

contracts/delegateTargets/ConditionalTransfer.sol
```diff
@@ -27,23 +27,29 @@ contract ConditionalTransfer is Conditional {
...
+      "App has been uninstalled"
+    );
+
+    require(
+      NonceRegistry(nonceRegistry).isFinalized(rootNonceKey, rootNonceExpectedValue),
```
snario: We're using this twice; perhaps we should fetch the value and do the check in this contract call versus two internal transactions to save on gas?

contracts/delegateTargets/ConditionalTransfer.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `isFinalized` does return a bool

PR #150, opened 2018-10-31
Removes unnecessary condition checking, makes proper use of Map type for Observable

joelalejandro:
By adding a private `getObservers(type)` function that guarantees an always-defined Array return value, we are able to remove any `if` blocks when registering or executing observers.


joelalejandro:
I'm closing this, as I've just realized there's a second, identical copy of the Observable mixin baked into the cf.js package. I'm thinking we should bring in a dependency for this and not worry about our own implementation.

PR #163, opened 2018-11-05
[WIP] Add AppInstanceAdjudicator

snario:
To save on gas costs, we can have a global contract for dispute resolution. This is the first step towards that. Currently a work in progress due to this error which is expected.

```console
contracts/StateChannelAdjudicator.sol:255:26:
CompilerError: Stack too deep, try removing local variables.
        computeStateHash(_id, keccak256(appState), nonce, timeout),
```

### DIFF
--- a/packages/contracts/contracts/AppInstanceAdjudicator.sol
+++ b/packages/contracts/contracts/AppInstanceAdjudicator.sol
@@ -248,6 +248,10 @@ contract AppInstanceAdjudicator {
     public
     onlyWhenChannelOpen(_id)
   {
+
+    // FIXME: Before merging, the timeout variable has been removed to get
+    // a stack too deep error to go away ... need to refactor function.
+
     AppInstance storage c = apps[_id];
 
     require(
--- a/packages/contracts/contracts/AppInstanceAdjudicator.sol
+++ b/packages/contracts/contracts/AppInstanceAdjudicator.sol
@@ -1,0 +1,575 @@
+pragma solidity ^0.4.24;
+pragma experimental "ABIEncoderV2";
+
+import "./lib/Signatures.sol";
+import "./lib/StaticCall.sol";
+import "./lib/Transfer.sol";
+
+
+/// @title StateChannelAdjudicator - A generalized state channel application contract
+/// @author Liam Horne - <liam@l4v.io>
+/// @notice Supports the adjudication and timeout guarantees required by state channel
+/// applications to be secure in a gas and storage-optimized manner. Resolves to a
+/// `Transfer.Transaction` when the channel is closed.
+contract StateChannelAdjudicator {
+
+  using Transfer for Transfer.Transaction;
+  using StaticCall for address;
+  using Signatures for bytes;
+
+  event DisputeStarted(
+    bytes32 indexed _id,
+    address sender,
+    uint256 disputeCounter,
+    bytes32 appStateHash,
+    uint256 nonce,
+    uint256 finalizesAt
+  );
+
+  event DisputeProgressed(
+    bytes32 indexed _id,
+    address sender,
+    bytes appState,
+    bytes action,
+    bytes toState,
+    uint256 disputeNonce,
+    uint256 finalizesAt
+  );
+
+  event DisputeFinalized(
+    bytes32 indexed _id,
+    address sender,
+    bytes finalState
+  );
+
+  event DisputeCancelled(
+    bytes32 indexed _id,
+    address sender
+  );
+
+  enum Status {
+    ON,
+    DISPUTE,
+    OFF
+  }
+
+  struct Auth {
+    address owner;
+    address[] signingKeys;
+  }
+
+  struct App {
+    address addr;
+    bytes4 applyAction;
+    bytes4 resolve;
+    bytes4 getTurnTaker;
+    bytes4 isStateTerminal;
+  }
+
+  struct State {
+    Status status;
+    bytes32 appStateHash;
+    address latestSubmitter;
+    uint256 nonce;
+    uint256 disputeNonce;
+    uint256 finalizesAt;
+    uint256 disputeCounter;
+  }
+
+  struct StateChannel {
+    Auth auth;
+    State state;
+    Transfer.Transaction resolution;
+    bytes32 appHash;
+    bytes32 termsHash;
+    uint256 defaultTimeout;
+  }
+
+  mapping (bytes32 => StateChannel) channels;
+
+  modifier onlyWhenChannelOpen(bytes32 _id) {
+    require(
+      !isStateTerminal(channels[_id].state),
+      "State has already been settled"
+    );
+    _;
+  }
+
+  modifier onlyWhenChannelDispute(bytes32 _id) {
+    require(
+      channels[_id].state.status == Status.DISPUTE,
+      "State is not being disputed"
+    );
+    _;
+  }
+
+  modifier onlyWhenChannelClosed(bytes32 _id) {
+    require(isStateTerminal(channels[_id].state), "State is still unsettled");
+    _;
+  }
+
+  /// @notice Contract constructor
+  /// @param owner An address which is allowed to set state. Typically an address off a multisig wallet.
+  /// @param signingKeys An array of addresses which can set state with unanimous consent
+  /// @param app Hash of the application's interface
+  /// @param terms Hash of a `Transfer.Terms` object commiting to the terms of the app
+  /// @param timeout The default timeout (in blocks) in case of dispute
+  function registerChannel(
+    address owner,
+    address[] signingKeys,
+    bytes32 app,
+    bytes32 terms,
+    uint256 timeout
+  )
+    external
+  {
+    bytes32 _id = keccak256(
+      abi.encodePacked(
+        owner, signingKeys, app, terms, timeout
+      )
+    );
+
+    StateChannel storage c = channels[_id];
+
+    c.auth.owner = owner;
+    c.auth.signingKeys = signingKeys;
+    c.termsHash = terms;
+    c.appHash = app;
+    c.defaultTimeout = timeout;
+  }
+
+  /// @notice A getter function for the owner of the state channel
+  /// @return The address of the `owner`
+  function getOwner(bytes32 _id) external view returns (address) {
+    return channels[_id].auth.owner;
+  }
+
+  /// @notice A getter function for the signing keys of the state channel
+  /// @return The addresses of the `signingKeys`
+  function getSigningKeys(bytes32 _id) external view returns (address[]) {
+    return channels[_id].auth.signingKeys;
+  }
+
+  /// @notice A getter function for the latest agreed nonce of the state channel
+  /// @return The uint value of the latest agreed nonce
+  function latestNonce(bytes32 _id) external view returns (uint256) {
+    return channels[_id].state.nonce;
+  }
+
+  /// @notice A helper method to determine whether or not the channel is closed
+  /// @return A boolean representing whether or not the state channel is closed
+  function isClosed(bytes32 _id) external view returns (bool) {
+    return isStateTerminal(channels[_id].state);
+  }
+
+  /// @notice A getter function for the resolution if one is set
+  /// @return A `Transfer.Transaction` object representing the resolution of the channel
+  function getResolution(bytes32 _id) public view returns (Transfer.Transaction) {
+    return channels[_id].resolution;
+  }
+
+  /// @notice Set the application state to a given value.
+  /// This value must have been signed off by all parties to the channel, that is,
+  /// this must be called with the correct msg.sender or signatures must be provided.
+  /// @param appStateHash The hash of the agreed upon state. Typically this is the latest signed state.
+  /// @param nonce The nonce of the agreed upon state
+  /// @param timeout A dynamic timeout value representing the timeout for this state
+  /// @param signatures A sorted bytes string of concatenated signatures of each signingKey
+  /// @dev This function is only callable when the state channel is in an ON state.
+  function setState(
+    bytes32 _id,
+    bytes32 appStateHash,
+    uint256 nonce,
+    uint256 timeout,
+    bytes signatures
+  )
+    public
+    onlyWhenChannelOpen(_id)
+  {
+    StateChannel storage c = channels[_id];
+
+    if (msg.sender != c.auth.owner) {
+      bytes32 h = computeStateHash(_id, appStateHash, nonce, timeout);
+      require(
+        signatures.verifySignatures(h, c.auth.signingKeys),
+        "Invalid signatures"
+      );
+    }
+
+    if (timeout > 0) {
+      require(
+        nonce > c.state.nonce,
+        "Tried to set state with non-new state"
+      );
+      c.state.status = Status.DISPUTE;
+    } else {
+      require(
+        nonce >= c.state.nonce,
+        "Tried to finalize state with stale state"
+      );
+      c.state.status = Status.OFF;
+    }
+
+    c.state.appStateHash = appStateHash;
+    c.state.nonce = nonce;
+    c.state.disputeNonce = 0;
+    c.state.finalizesAt = block.number + timeout;
+    c.state.disputeCounter += 1;
+    c.state.latestSubmitter = msg.sender;
+  }
+
+  /// @notice Create a dispute regarding the latest signed state and immediately after,
+  /// performs a unilateral action to update it.
+  /// @param app An `App` struct specifying the application logic
+  /// @param appState The ABI encoded application state
+  /// @param nonce The nonce of the agreed upon state
+  /// @param timeout A dynamic timeout value representing the timeout for this state
+  /// @param action The ABI encoded action the submitter wishes to take
+  /// @param appStateSignatures A sorted bytes string of concatenated signatures on the
+  /// `appState` state, signed by all `signingKeys`
+  /// @param actionSignature A bytes string of a single signature by the address of the
+  /// signing key for which it is their turn to take the submitted `action`
+  /// @param claimFinal A boolean representing a claim by the caller that the action
+  /// progresses the state of the application to a terminal / finalized state
+  /// @dev Note this function is only callable when the state channel is in an ON state
+  function createDispute(
+    bytes32 _id,
+    App app,
+    bytes appState,
+    uint256 nonce,
+    uint256 timeout,
+    bytes action,
+    bytes appStateSignatures,
+    bytes actionSignature,
+    bool claimFinal
+  )
+    public
+    onlyWhenChannelOpen(_id)
+  {
+    StateChannel storage c = channels[_id];
+
+    require(
+      nonce > c.state.nonce,
+      "Tried to create dispute with outdated state"
+    );
+
+    bytes32 appStateHash = keccak256(appState);
+    require(
+      appStateSignatures.verifySignatures(
+        computeStateHash(_id, appStateHash, nonce, timeout),
+        c.auth.signingKeys
+      ),
+      "Invalid signatures"
+    );
+
+    address turnTaker = getAppTurnTaker(_id, app, appState);
+
+    bytes32 actionHash = computeActionHash(
+      turnTaker,
+      keccak256(appState),
+      action,
+      nonce,
+      c.state.disputeNonce
+    );
+    require(
+      turnTaker == actionSignature.recoverKey(actionHash, 0),
+      "Action must have been signed by correct turn taker"
+    );
+
+    emit DisputeStarted(
+      _id,
+      msg.sender,
+      c.state.disputeCounter + 1,
+      appStateHash,
+      nonce,
+      block.number + timeout
+    );
+
+    bytes memory newAppState = executeAppApplyAction(app, appState, action);
+
+    c.state.appStateHash = keccak256(newAppState);
+    c.state.nonce = nonce;
+    c.state.disputeNonce = 0;
+    c.state.disputeCounter += 1;
+    c.state.latestSubmitter = msg.sender;
+
+    if (claimFinal) {
+      require(isAppStateTerminal(app, newAppState));
+      c.state.finalizesAt = block.number;
+      c.state.status = Status.OFF;
+
+      emit DisputeFinalized(_id, msg.sender, newAppState);
+    } else {
+      c.state.finalizesAt = block.number + timeout;
+      c.state.status = Status.DISPUTE;
+
+      emit DisputeProgressed(
+        _id,
+        msg.sender,
+        appState,
+        action,
+        newAppState,
+        c.state.disputeNonce,
+        block.number + timeout
+      );
+    }
+  }
+
+  /// @notice Respond to a dispute with a valid action
+  /// @param app An `App` struct specifying the application logic
+  /// @param appState The ABI encoded latest signed application state
+  /// @param action The ABI encoded action the submitter wishes to take
+  /// @param actionSignature A bytes string of a single signature by the address of the
+  /// signing key for which it is their turn to take the submitted `action`
+  /// @param claimFinal If set, the caller claims that the action progresses the state
+  /// to a terminal / finalized state
+  /// @dev This function is only callable when the state channel is in a DISPUTE state
+  function progressDispute(
+    bytes32 _id,
+    App app,
+    bytes appState,
+    bytes action,
+    bytes actionSignature,
+    bool claimFinal
+  )
+    public
+    onlyWhenChannelDispute(_id)
+  {
+    StateChannel storage c = channels[_id];
+
+    require(
+      keccak256(appState) == c.state.appStateHash,
+      "Invalid state submitted"
+    );
+
+    require(
+      keccak256(abi.encode(app)) == c.appHash,
+      "Tried to resolve dispute with non-agreed upon app"
+    );
+
+    address turnTaker = getAppTurnTaker(_id, app, appState);
+
+    require(
+      turnTaker == actionSignature.recoverKey(keccak256(action), 0),
+      "Action must have been signed by correct turn taker"
+    );
+
+    bytes memory newAppState = executeAppApplyAction(app, appState, action);
+
+    c.state.appStateHash = keccak256(newAppState);
+    c.state.disputeNonce += 1;
+    c.state.latestSubmitter = msg.sender;
+
+    if (claimFinal) {
+      require(isAppStateTerminal(app, newAppState));
+      c.state.finalizesAt = block.number;
+      c.state.status = Status.OFF;
+
+      emit DisputeFinalized(_id, msg.sender, newAppState);
+    } else {
+      c.state.status = Status.DISPUTE;
+      c.state.finalizesAt = block.number + c.defaultTimeout;
+
+      emit DisputeProgressed(
+        _id,
+        msg.sender,
+        appState,
+        action,
+        newAppState,
+        c.state.disputeNonce,
+        block.number + c.defaultTimeout
+      );
+    }
+  }
+
+  /// @notice Unanimously agree to cancel a dispute
+  /// @param signatures Signatures by all signing keys of the currently latest disputed
+  /// state; an indication of agreement of this state and valid to cancel a dispute
+  /// @dev Note this function is only callable when the state channel is in a DISPUTE state
+  function cancelDispute(bytes32 _id, bytes signatures)
+    public
+    onlyWhenChannelDispute(_id)
+  {
+    StateChannel storage c = channels[_id];
+
+    bytes32 stateHash = computeStateHash(
+      _id,
+      c.state.appStateHash,
+      c.state.nonce,
+      c.defaultTimeout
+    );
+
+    require(
+      signatures.verifySignatures(stateHash, c.auth.signingKeys),
+      "Invalid signatures"
+    );
+
+    c.state.disputeNonce = 0;
+    c.state.finalizesAt = 0;
+    c.state.status = Status.ON;
+    c.state.latestSubmitter = msg.sender;
+
+    emit DisputeCancelled(_id,msg.sender);
+  }
+
+  /// @notice Fetch and store the resolution of a state channel application
+  /// @param app An `App` struct including all information relevant to interface with an app
+  /// @param finalState The ABI encoded version of the finalized application state
+  /// @param terms The ABI encoded version of the already agreed upon terms
+  /// @dev Note this function is only callable when the state channel is in an OFF state
+  function setResolution(bytes32 _id, App app, bytes finalState, bytes terms)
+    public
+    onlyWhenChannelClosed(_id)
+  {
+    StateChannel storage c = channels[_id];
+
+    require(
+      keccak256(finalState) == c.state.appStateHash,
+      "Tried to set resolution with incorrect final state"
+    );
+
+    require(
+      keccak256(terms) == c.termsHash,
+      "Tried to set resolution with non-agreed upon terms"
+    );
+
+    require(
+      keccak256(abi.encode(app)) == c.appHash,
+      "Tried to set resolution with non-agreed upon app"
+    );
+
+    c.resolution = getAppResolution(app, finalState, terms);
+  }
+
+  /// @notice A helper method to check if the state of the channel is final by
+  /// doing a check on the submitted state and comparing to the current block number
+  /// @param s A state wrapper struct including the status and finalization time
+  /// @return A boolean indicating if the state is final or not
+  function isStateTerminal(State s) public view returns (bool) {
+    if (s.status == Status.ON) {
+      return false;
+    } else if (s.status == Status.DISPUTE) {
+      return block.number >= s.finalizesAt;
+    } else if (s.status == Status.OFF) {
+      return true;
+    }
+  }
+
+  /// @notice Compute a unique hash for a state of this state channel and application
+  /// @param stateHash The hash of a state to be signed
+  /// @param nonce The nonce corresponding to the version of the state
+  /// @param timeout A dynamic timeout value representing the timeout for this state
+  /// @return A bytes32 hash of the arguments encoded with the signing keys for the channel
+  function computeStateHash(bytes32 _id, bytes32 stateHash, uint256 nonce, uint256 timeout)
+    internal
+    view
+    returns (bytes32)
+  {
+    return keccak256(
+      abi.encodePacked(
+        byte(0x19),
+        channels[_id].auth.signingKeys,
+        nonce,
+        timeout,
+        stateHash
+      )
+    );
+  }
+
+  /// @notice Compute a unique hash for an action used in this channel application
+  /// @param turnTaker The address of the user taking the action
+  /// @param previousState The hash of a state this action is being taken on
+  /// @param action The ABI encoded version of the action being taken
+  /// @param setStateNonce The nonce of the state this action is being taken on
+  /// @param disputeNonce A nonce corresponding to how many actions have been taken on the
+  /// state since a new state has been unanimously agreed upon by all signing keys.
+  /// @return A bytes32 hash of the arguments
+  function computeActionHash(
+    address turnTaker,
+    bytes32 previousState,
+    bytes action,
+    uint256 setStateNonce,
+    uint256 disputeNonce
+  )
+    internal
+    pure
+    returns (bytes32)
+  {
+    return keccak256(
+      abi.encodePacked(
+        byte(0x19),
+        turnTaker,
+        previousState,
+        action,
+        setStateNonce,
+        disputeNonce
+      )
+    );
+  }
+
+  /// @notice A helper method to check if the state of an application is terminal or not
+  /// @param app An `App` struct including all information relevant to interface with an app
+  /// @param appState The ABI encoded version of some application state
+  /// @return A boolean indicating if the application state is terminal or not
+  function isAppStateTerminal(App app, bytes appState)
+    private
+    view
+    returns (bool)
+  {
+    return app.addr.staticcall_as_bool(
+      abi.encodePacked(app.isStateTerminal, appState)
+    );
+  }
+
+  /// @notice A helper method to get the turn taker for an app
+  /// @param app An `App` struct including all information relevant to interface with an app
+  /// @param appState The ABI encoded version of some application state
+  /// @return An address representing the turn taker in the `signingKeys`
+  function getAppTurnTaker(bytes32 _id, App app, bytes appState)
+    private
+    view
+    returns (address)
+  {
+    uint256 idx = app.addr.staticcall_as_uint256(
+      abi.encodePacked(app.getTurnTaker, appState)
+    );
+
+    require(
+      channels[_id].auth.signingKeys[idx] != address(0),
+      "Application returned invalid turn taker index"
+    );
+
+    return channels[_id].auth.signingKeys[idx];
+  }
+
+  /// @notice Execute the application's applyAction function to compute new state
+  /// @param app An `App` struct including all information relevant to interface with an app
+  /// @param appState The ABI encoded version of some application state
+  /// @param action The ABI encoded version of some application action
+  /// @return A bytes array of the ABI encoded newly computed application state
+  function executeAppApplyAction(App app, bytes appState, bytes action)
+    private
+    view
+    returns (bytes)
+  {
+    return app.addr.staticcall_as_bytes(
+      abi.encodePacked(app.applyAction, appState, action)
+    );
+  }
+
+  /// @notice Execute the application's resolve function to compute a resolution
+  /// @param app An `App` struct including all information relevant to interface with an app
+  /// @param appState The ABI encoded version of some application state
+  /// @param terms The ABI encoded version of the transfer terms
+  /// @return A `Transfer.Transaction` struct with all encoded information of the resolution
+  function getAppResolution(App app, bytes appState, bytes terms)
+    private
+    view
+    returns (Transfer.Transaction)
+  {
+    return app.addr.staticcall_as_TransferDetails(
+      abi.encodePacked(app.resolve, appState, terms)
+    );
+  }
+
+}
--- a/packages/contracts/contracts/AppInstanceAdjudicator.sol
+++ b/packages/contracts/contracts/AppInstanceAdjudicator.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.4.25;
 pragma experimental "ABIEncoderV2";
 
 import "./lib/Signatures.sol";
@@ -6,12 +6,12 @@ import "./lib/StaticCall.sol";
 import "./lib/Transfer.sol";
 
 
-/// @title StateChannelAdjudicator - A generalized state channel application contract
+/// @title AppInstanceAdjudicator - A generalized state channel application contract
 /// @author Liam Horne - <liam@l4v.io>
 /// @notice Supports the adjudication and timeout guarantees required by state channel
 /// applications to be secure in a gas and storage-optimized manner. Resolves to a
 /// `Transfer.Transaction` when the channel is closed.
-contract StateChannelAdjudicator {
+contract AppInstanceAdjudicator {
 
   using Transfer for Transfer.Transaction;
   using StaticCall for address;
@@ -106,6 +106,10 @@ contract StateChannelAdjudicator {
   modifier onlyWhenChannelClosed(bytes32 _id) {
     require(isStateTerminal(channels[_id].state), "State is still unsettled");
     _;
+  }
+
+  function getChannel(bytes32 _id) external view returns (StateChannel) {
+    return channels[_id];
   }
 
   /// @notice Contract constructor
@@ -223,7 +227,6 @@ contract StateChannelAdjudicator {
   /// @param app An `App` struct specifying the application logic
   /// @param appState The ABI encoded application state
   /// @param nonce The nonce of the agreed upon state
-  /// @param timeout A dynamic timeout value representing the timeout for this state
   /// @param action The ABI encoded action the submitter wishes to take
   /// @param appStateSignatures A sorted bytes string of concatenated signatures on the
   /// `appState` state, signed by all `signingKeys`
@@ -237,7 +240,6 @@ contract StateChannelAdjudicator {
     App app,
     bytes appState,
     uint256 nonce,
-    uint256 timeout,
     bytes action,
     bytes appStateSignatures,
     bytes actionSignature,
@@ -253,10 +255,9 @@ contract StateChannelAdjudicator {
       "Tried to create dispute with outdated state"
     );
 
-    bytes32 appStateHash = keccak256(appState);
     require(
       appStateSignatures.verifySignatures(
-        computeStateHash(_id, appStateHash, nonce, timeout),
+        computeStateHash(_id, keccak256(appState), nonce, 1337),
         c.auth.signingKeys
       ),
       "Invalid signatures"
@@ -264,15 +265,17 @@ contract StateChannelAdjudicator {
 
     address turnTaker = getAppTurnTaker(_id, app, appState);
 
-    bytes32 actionHash = computeActionHash(
-      turnTaker,
-      keccak256(appState),
-      action,
-      nonce,
-      c.state.disputeNonce
-    );
     require(
-      turnTaker == actionSignature.recoverKey(actionHash, 0),
+      turnTaker == actionSignature.recoverKey(
+        computeActionHash(
+          turnTaker,
+          keccak256(appState),
+          action,
+          nonce,
+          c.state.disputeNonce
+        ),
+        0
+      ),
       "Action must have been signed by correct turn taker"
     );
 
@@ -280,9 +283,9 @@ contract StateChannelAdjudicator {
       _id,
       msg.sender,
       c.state.disputeCounter + 1,
-      appStateHash,
+      keccak256(appState),
       nonce,
-      block.number + timeout
+      block.number + 1337
     );
 
     bytes memory newAppState = executeAppApplyAction(app, appState, action);
@@ -300,7 +303,7 @@ contract StateChannelAdjudicator {
 
       emit DisputeFinalized(_id, msg.sender, newAppState);
     } else {
-      c.state.finalizesAt = block.number + timeout;
+      c.state.finalizesAt = block.number + 1337;
       c.state.status = Status.DISPUTE;
 
       emit DisputeProgressed(
@@ -310,7 +313,7 @@ contract StateChannelAdjudicator {
         action,
         newAppState,
         c.state.disputeNonce,
-        block.number + timeout
+        block.number + 1337
       );
     }
   }
--- a/packages/contracts/contracts/AppInstanceAdjudicator.sol
+++ b/packages/contracts/contracts/AppInstanceAdjudicator.sol
@@ -412,7 +412,7 @@ contract AppInstanceAdjudicator {
     c.state.status = Status.ON;
     c.state.latestSubmitter = msg.sender;
 
-    emit DisputeCancelled(_id,msg.sender);
+    emit DisputeCancelled(_id, msg.sender);
   }
 
   /// @notice Fetch and store the resolution of a state channel application
--- a/packages/contracts/contracts/AppInstanceAdjudicator.sol
+++ b/packages/contracts/contracts/AppInstanceAdjudicator.sol
@@ -76,7 +76,7 @@ contract AppInstanceAdjudicator {
     uint256 disputeCounter;
   }
 
-  struct StateChannel {
+  struct AppInstance {
     Auth auth;
     State state;
     Transfer.Transaction resolution;
@@ -85,11 +85,11 @@ contract AppInstanceAdjudicator {
     uint256 defaultTimeout;
   }
 
-  mapping (bytes32 => StateChannel) channels;
+  mapping (bytes32 => AppInstance) apps;
 
   modifier onlyWhenChannelOpen(bytes32 _id) {
     require(
-      !isStateTerminal(channels[_id].state),
+      !isStateTerminal(apps[_id].state),
       "State has already been settled"
     );
     _;
@@ -97,19 +97,19 @@ contract AppInstanceAdjudicator {
 
   modifier onlyWhenChannelDispute(bytes32 _id) {
     require(
-      channels[_id].state.status == Status.DISPUTE,
+      apps[_id].state.status == Status.DISPUTE,
       "State is not being disputed"
     );
     _;
   }
 
   modifier onlyWhenChannelClosed(bytes32 _id) {
-    require(isStateTerminal(channels[_id].state), "State is still unsettled");
+    require(isStateTerminal(apps[_id].state), "State is still unsettled");
     _;
   }
 
-  function getChannel(bytes32 _id) external view returns (StateChannel) {
-    return channels[_id];
+  function getAppInstance(bytes32 _id) external view returns (AppInstance) {
+    return apps[_id];
   }
 
   /// @notice Contract constructor
@@ -118,7 +118,7 @@ contract AppInstanceAdjudicator {
   /// @param app Hash of the application's interface
   /// @param terms Hash of a `Transfer.Terms` object commiting to the terms of the app
   /// @param timeout The default timeout (in blocks) in case of dispute
-  function registerChannel(
+  function registerAppInstance(
     address owner,
     address[] signingKeys,
     bytes32 app,
@@ -133,7 +133,7 @@ contract AppInstanceAdjudicator {
       )
     );
 
-    StateChannel storage c = channels[_id];
+    AppInstance storage c = apps[_id];
 
     c.auth.owner = owner;
     c.auth.signingKeys = signingKeys;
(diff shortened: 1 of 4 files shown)


snario:
Tests currently pass for me with the recent changes but there are some caveats...

- [ ] Haven't really carefully evaluated security properties of this approach yet, still TODO
- [ ] Hard-coded in a 1337 timeout versus using a variable in `createDispute` because of a `stack too deep` error. Need to find a workaround still.

Still would appreciate any extra feedback @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll

packages/contracts/contracts/AppInstanceAdjudicator.sol
```diff
@@ -0,0 +1,578 @@
...
+    );
+    _;
+  }
+
+  modifier onlyWhenChannelClosed(bytes32 _id) {
+    require(isStateTerminal(channels[_id].state), "State is still unsettled");
+    _;
+  }
+
+  function getChannel(bytes32 _id) external view returns (StateChannel) {
```
snario: Reminder for myself... need to add docstring here.

packages/contracts/contracts/AppInstanceAdjudicator.sol
```diff
@@ -0,0 +1,578 @@
...
+  /// @return A bytes32 hash of the arguments encoded with the signing keys for the channel
+  function computeStateHash(bytes32 _id, bytes32 stateHash, uint256 nonce, uint256 timeout)
+    internal
+    view
+    returns (bytes32)
+  {
+    return keccak256(
+      abi.encodePacked(
+        byte(0x19),
+        channels[_id].auth.signingKeys,
```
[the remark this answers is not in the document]
snario: Good point, will update tomorrow

packages/contracts/contracts/AppInstanceAdjudicator.sol
```diff
@@ -0,0 +1,582 @@
...
+  struct AppInstance {
+    Auth auth;
+    State state;
+    Transfer.Transaction resolution;
+    bytes32 appHash;
+    bytes32 termsHash;
+    uint256 defaultTimeout;
+  }
+
+  mapping (bytes32 => AppInstance) apps;
```
mvanderh: `instances` is a better name here IMO.

packages/contracts/contracts/AppInstanceAdjudicator.sol
```diff
@@ -0,0 +1,582 @@
...
+  )
+    external
+  {
+    bytes32 _id = keccak256(
+      abi.encodePacked(
+        owner, signingKeys, app, terms, timeout
+      )
+    );
+
+    AppInstance storage c = apps[_id];
```
mvanderh: Rename "c" to "inst" or "instance" or something.

packages/contracts/contracts/AppInstanceAdjudicator.sol
```diff
@@ -0,0 +1,582 @@
...
+    bytes appStateSignatures,
+    bytes actionSignature,
+    bool claimFinal
+  )
+    public
+    onlyWhenChannelOpen(_id)
+  {
+
+    // FIXME: Before merging, the timeout variable has been removed to get
+    // a stack too deep error to go away ... need to refactor function.
```
mvanderh: Is solving this error just a matter of adding an extra function?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
-1 on the name `AppInstanceAdjudicator`; maybe `AppInstanceRegistry`?

packages/contracts/contracts/AppInstanceAdjudicator.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `appInstances` is even better

mvanderh:
I prefer "adjudicator" to "registry" tbh. It describes the raison d'etre of the contract, which is to adjudicate disputes on-chain.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
@mvanderh that explanation doesn't support the `Instance` in the name though

packages/contracts/contracts/AppInstanceAdjudicator.sol
snario: That would be an unfortunate way of solving it. Seems like I can place more local variable inside structs to solve it.

mvanderh:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll Why not? The contract adjudicates app instances.

cf19drofxots:
@snario Status update?

snario:
> @snario Status update?

In progress.

packages/contracts/networks/4447.json
```diff
@@ -0,0 +1,242 @@
+[
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: is this meant to be committed?

snario: Nope

package.json
```diff
@@ -17,32 +17,15 @@
     "clean": "lerna run clean --parallel --no-bail",
     "test": "lerna run --stream --concurrency 1 test",
     "test:ci-non-playground": "lerna run --ignore @counterfactual/playground --ignore @counterfactual/dapp-high-roller --stream test",
-    "ganache": "ganache-cli --networkId ${npm_package_config_ganacheNetworkID} --verbose --gasLimit ${npm_package_config_ganacheGasLimit} --gasPrice ${npm_package_config_ganacheGasPrice} --port ${npm_package_config_ganachePort} --deterministic --account=\"${npm_package_config_unlockedAccount0},${npm_package_config_etherBalance}\" --account=\"${npm_package_config_unlockedAccount1},${npm_package_config_etherBalance}\" --account=\"${npm_package_config_unlockedAccount2},${npm_package_config_etherBalance}\" --account=\"${npm_package_config_unlockedAccount3},${npm_package_config_etherBalance}\" &> /dev/null &",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: do we not get top-level ganache anymore? how do some of the machine tests which run against an EVM work? can you reflect these in the README and circleci?

packages/contracts/test/utils/abi-encoder-v2.js.map
```diff
@@ -0,0 +1 @@
+{"version":3,"file":"abi-encoder-v2.js","sourceRoot":"","sources":["abi-encoder-v2.ts"],"names":[],"mappings":";;AAAA,wCAA+C;AAE/C,SAAgB,oBAAoB,CAAC,gBAAwB;IAC3D,MAAM,WAAW,GAAa,EAAE,CAAC;IACjC,MAAM,KAAK,GAAG,gBAAgB,CAAC,KAAK,CAAC,GAAG,CAAC,CAAC;IAC1C,KAAK,MAAM,IAAI,IAAI,KAAK,EAAE;QACxB,MAAM,UAAU,GAAG,IAAI,CAAC,IAAI,EAAE,CAAC;QAC/B,IAAI,UAAU,CAAC,MAAM,KAAK,CAAC,EAAE;YAC3B,SAAS;SACV;QACD,MAAM,KAAK,GAAG,UAAU,CAAC,KAAK,CAAC,GAAG,CAAC,CAAC;QACpC,IAAI,KAAK,CAAC,MAAM,KAAK,CAAC,EAAE;YACtB,MAAM,IAAI,KAAK,CACb,wDAAwD,UAAU,GAAG,CACtE,CAAC;SACH;QACD,WAAW,CAAC,IAAI,CAAC,UAAU,CAAC,CAAC;KAC9B;IACD,OAAO,SAAS,WAAW,CAAC,IAAI,CAAC,IAAI,CAAC,GAAG,CAAC;AAC5C,CAAC;AAjBD,oDAiBC;AAED,SAAgB,YAAY,CAAC,QAAgB,EAAE,MAAc;IAC3D,OAAO,uBAAe,CAAC,MAAM,CAAC,CAAC,QAAQ,CAAC,EAAE,CAAC,MAAM,CAAC,CAAC,CAAC;AACtD,CAAC;AAFD,oCAEC"}
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: is this `.map` file meant to be committed?

packages/contracts/contracts/state-deposit-holders/MinimumViableMultisig.sol
```diff
@@ -11,9 +11,9 @@ import "./lib/Signatures.sol";
 /// (b) Requires n-of-n unanimous consent
 /// (c) Does not use on-chain address for signature verification
 /// (d) Uses hash-based instead of nonce-based replay protection
-contract MinimumViableMultisig {
+contract MinimumViableMultisig is LibSignature {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why change `Signature` to a contract?

packages/common-types/src/node-protocol.ts
```diff
@@ -13,6 +13,17 @@ export interface INodeProvider {
...
+    // Protocol
+    ConditionalTransaction: Address;
+    MultiSend: Address;
+    NonceRegistry: Address;
+    AppRegistry: Address;
+    // App-specific
+    PaymentApp: Address;
+    ETHBalanceRefund: Address;
+  };
+
```
snario: needs to be updated

package.json
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Is the rewrite part of this PR? It was a pretty useful test

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
There is a current name clash between `AppInstanceId` defined in the PR body here and in the cf.js API docs; need to differentiate the names and implement the rename

packages/apps/contracts/TicTacToeApp.sol
```diff
@@ -1,10 +1,10 @@
 pragma solidity 0.4.25;
 pragma experimental "ABIEncoderV2";
 
-import "../lib/Transfer.sol";
+import "@counterfactual/contracts/contracts/libs/Transfer.sol";
 
 
-contract TicTacToe {
+contract TicTacToeApp {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I think if we add a suffix, it should be `AppDefinition`

packages/cf.js/package.json
```diff
@@ -12,12 +12,12 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "clean": "rm -rf .rpt2_cache dist",
+    "clean": "git clean -Xdf",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I don't think we need a thin alias for `git clean`. Better to just remove the `clean` command

packages/contracts/test/conditional-transaction.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+import { AddressZero, HashZero, WeiPerEther, Zero } from "ethers/constants";
+import { JsonRpcSigner, Web3Provider } from "ethers/providers";
+import {
+  defaultAbiCoder,
+  hexlify,
+  randomBytes,
+  solidityKeccak256
+} from "ethers/utils";
+
+import { expect } from "./utils";
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: - seems better to me to inline this 
- the definition of `expect` varies through this codebase; is it possible to have it be the same thing everywhere?

package.json
snario: I can add one back in but I think it would be better to do it in a new PR

packages/contracts/test/utils/abi-encoder-v2.js.map
snario: no

snario:
> There is a current name clash between `AppInstanceId` defined in the PR body here and in the cf.js API docs; need to differentiate the names and implement the rename

Agree. @ebryn @mvanderh

packages/contracts/test/conditional-transaction.spec.ts
snario: It's necessarily different in test files that use Jest vs Mocha / Jasmine

packages/cf.js/package.json
snario: We have multiple layers of gitignore files and so `git clean` doesn't work the same way at the root level which is why we "dispatch" it to the packages

snario: Actually nevermind i just hadn't tried it .. https://stackoverflow.com/questions/9730486/can-you-have-additional-gitignore-per-directory-within-a-single-repo

packages/apps/contracts/TicTacToeApp.sol
snario: Probably nicer if we don't require prefixes. In Solidity 0.5 we'll be able to enforce that the programmer inherits from some interface which we can call `interface AppDefinition { ... }` so it looks like `contract TicTacToe is AppDefinition { ... }`

packages/contracts/test/utils/multisig.ts
```diff
@@ -0,0 +1,143 @@
+// import { ethers } from "ethers";
```
mvanderh: What's all this?

packages/contracts/test/app-registry.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+  beforeEach(async () => {
+    const appInstance = new AppInstance(
+      accounts[0],
+      [ALICE.address, BOB.address],
+      new AppInterface(
+        ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+        ethers.utils.hexlify(ethers.utils.randomBytes(4)),
+        ethers.utils.hexlify(ethers.utils.randomBytes(4)),
+        ethers.utils.hexlify(ethers.utils.randomBytes(4)),
+        ethers.utils.hexlify(ethers.utils.randomBytes(4))
```
mvanderh: Prefer these to be constant and not randomized.

packages/contracts/test/app-registry.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    });
+  });
+
+  describe("waiting for timeout", async () => {
+    it("should block updates after the timeout", async () => {
+      expect(await isStateFinalized()).to.be.false;
+
+      await setStateAsOwner(1);
+
+      for (const _ of Array(TIMEOUT + 1)) {
```
mvanderh: This works? 😮

packages/contracts/contracts/AppRegistry.sol
```diff
@@ -0,0 +1,20 @@
...
+pragma experimental "ABIEncoderV2";
+
+import "./mixins/MixinAppRegistryCore.sol";
+import "./mixins/MixinCancelChallenge.sol";
+import "./mixins/MixinSetResolution.sol";
+import "./mixins/MixinSetState.sol";
+import "./mixins/MixinSetStateWithAction.sol";
+
+
+/// @dev Base contract implmenting all logic needed for full-featured App registry
```
mvanderh: typo

packages/contracts/contracts/mixins/MixinAppRegistryCore.sol
```diff
@@ -0,0 +1,52 @@
...
+  /// @return A `AppChallenge` object representing the state of the on-chain challenge
+  function getAppChallenge(bytes32 _id)
+    external
+    view
+    returns (LibStateChannelApp.AppChallenge)
+  {
+    return appStates[_id];
+  }
+
+  // TODO:
```
mvanderh: Todo what?

packages/contracts/contracts/mixins/MixinSetState.sol
```diff
@@ -0,0 +1,97 @@
...
+  /// This value must have been signed off by all parties to the channel, that is,
+  /// this must be called with the correct msg.sender or signatures must be provided.
+  /// @param appIdentity an AppIdentity struct with all information encoded within
+  ///        it to represent which particular app is having state submitted
+  /// @param req An object containing the update to be applied to the
+  ///        applications state including the signatures of the users needed
+  /// @dev This function is only callable when the state channel is in an ON state.
+  function setState(
+    AppIdentity appIdentity,
+    SignedStateHashUpdate req
```
mvanderh: Weird name for a "SignedStateHashUpdate"

packages/contracts/contracts/libs/LibStaticCall.sol
```diff
@@ -24,7 +25,7 @@ library StaticCall {
   {
     require(to.isContract(), "StaticCall to address is not a contract.");
     assembly {
-      let success := staticcall(gas, to, add(data, 0x20), mload(data), 0, 0)
+      success := staticcall(gas, to, add(data, 0x20), mload(data), 0, 0)
```
mvanderh: Out of curiosity, what's the difference between these assembly stmts?

packages/contracts/contracts/libs/LibStateChannelApp.sol
```diff
@@ -0,0 +1,54 @@
...
+  // A minimal structure that uniquely identifies a single instance of an App
+  struct AppIdentity {
+    address owner;
+    address[] signingKeys;
+    bytes32 appInterfaceHash;
+    bytes32 termsHash;
+    uint256 defaultTimeout;
+  }
+
+  // A minimal structure representing the state of an App from POV of the blockchain
```
mvanderh: "App Instance" instead of "App"

packages/contracts/contracts/state-deposit-holders/MinimumViableMultisig.sol
```diff
@@ -57,7 +57,7 @@ contract MinimumViableMultisig {
     bytes32 transactionHash = getTransactionHash(to, value, data, operation);
 
     require(
-      signatures.verifySignatures(transactionHash, _owners),
+      verifySignatures(signatures, transactionHash, _owners),
```
mvanderh: +1 this change

packages/contracts/test/app-registry-new.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      status,
+      latestSubmitter,
+      appStateHash,
+      disputeCounter,
+      disputeNonce,
+      finalizesAt,
+      nonce
+    } = await appRegistry.functions.appStates(appInstance.id);
+
+    console.log({
```
mvanderh: Remove this log pls

packages/machine/src/models/state-channel.ts
```diff
@@ -0,0 +1,221 @@
...
+    );
+
+    const freeBalanceAppIndexes = new Map<AssetType, string>(
+      this.freeBalanceAppIndexes.entries()
+    );
+
+    return new StateChannel(
+      this.multisigAddress,
+      this.multisigOwners,
+      appInstances
```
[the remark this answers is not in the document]
snario: Sorry what is dot train ?

packages/machine/src/protocol-types-tbd.ts
```diff
@@ -0,0 +1,50 @@
...
+  toAddress: string;
+  seq: number;
+  signature?: Signature;
+};
+
+export type SetupData = {};
+
+export type UpdateData = {
+  appInstanceId: string;
+  newState: object;
```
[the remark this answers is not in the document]
snario: tbh I don’t know what this type is I think this is unused and should be removed

packages/contracts/contracts/libs/LibStaticCall.sol
snario: I’m not sure how the first one passed since the success variable name is already in scope

packages/apps/test/nim.spec.ts
```diff
@@ -0,0 +1,117 @@
+import chai from "chai";
+import * as waffle from "ethereum-waffle";
+import { ethers } from "ethers";
```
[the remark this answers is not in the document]
snario: Hm this one was missed, will fix

packages/machine/src/ethereum/utils/multisend-decoder.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export function decodeMultisendCalldata(txs: string) {
+  const ret: [number, string, number, string][] = [];
+
+  let i = 0;
+
+  while (i < hexDataLength(txs)) {
+    // We expect 0x80 to be a hard-coded pointer to the `data` location. Refer to
+    // https://solidity.readthedocs.io/en/v0.5.0/abi-spec.html for ABI specification
+    const ptr = hexDataSlice(txs, i + 0x60, i + 0x80);
```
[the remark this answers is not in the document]
snario: I think as-is is precisely as meaningful as it can be — it represents locations in memory

packages/machine/src/ethereum/utils/types.ts
```diff
@@ -0,0 +1,30 @@
+import { Signature } from "ethers/utils";
+
+export abstract class EthereumCommitment {
+  public abstract hashToSign(): string;
+  public abstract transaction(sigs: Signature[]): Transaction;
+}
+
+export enum MultisigOperation {
+  Call = 0,
+  Delegatecall = 1,
```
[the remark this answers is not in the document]
snario: Not really any particular standard

packages/machine/src/models/app-instance.ts
```diff
@@ -0,0 +1,231 @@
...
+import { Memoize } from "typescript-memoize";
+
+import { appIdentityToHash } from "../ethereum/utils/app-identity";
+import { APP_INTERFACE, TERMS } from "../ethereum/utils/encodings";
+
+/**
+ * Represenation of the values a dependency nonce can take on.
+ */
+export enum DependencyValue {
+  NOT_UNINSTALLED = 0,
```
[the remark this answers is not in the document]
snario: I did it this way because the _default_ status of the value is that an app has not explicitly been uninstalled yet. There is no step that changes this value from “nothing” to “installed”. There is only a step from “not uninstalled” to “uninsyalled”

packages/machine/src/protocol-types-tbd.ts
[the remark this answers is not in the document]
snario: This is literally arbitrary state as defined by the developer writing their application — it can be anything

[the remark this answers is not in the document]
snario: Actually I see your point — it can be nested though

packages/machine/src/protocol/index.ts
```diff
@@ -0,0 +1,32 @@
...
+import { UPDATE_PROTOCOL } from "./update";
+
+export function getProtocolFromName(protocolName: Protocol) {
+  switch (protocolName) {
+    case Protocol.Setup:
+      return SETUP_PROTOCOL;
+    case Protocol.Install:
+      return INSTALL_PROTOCOL;
+    case Protocol.SetState:
+      return UPDATE_PROTOCOL;
```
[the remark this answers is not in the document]
snario: Yeah........ @ldct thoughts?

packages/machine/test/unit/models/state-channel/uninstall-app.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { AssetType, ETHBucketAppState } from "@counterfactual/types";
+import { AddressZero, Zero } from "ethers/constants";
+import { bigNumberify, getAddress, hexlify, randomBytes } from "ethers/utils";
+
+import { AppInstance, StateChannel } from "../../../../src/models";
+
+describe("StateChannel::uninstallApp", () => {
+  const networkContext = {
+    ETHBucket: getAddress(hexlify(randomBytes(20))),
```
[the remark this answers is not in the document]
snario: Yeah I’m being lazy here but essentially it’s just important that every value is unique every time I run the random function so the tests integrity is solid

packages/contracts/contracts/mixins/MixinSetStateWithAction.sol
```diff
@@ -0,0 +1,153 @@
...
+    bytes encodedAction;
+    bytes signature;
+    bool checkForTerminal;
+  }
+
+  /// @notice Create a dispute regarding the latest signed state and immediately after,
+  /// performs a unilateral action to update it.
+  // @param app An `App` struct specifying the application logic
+  // TODO: Docs
+  /// @dev Note this function is only callable when the state channel is in an ON state
```
cf19drofxots: Is there a modifier enforcing this?

packages/contracts/package.json
```diff
@@ -9,22 +9,20 @@
...
   ],
   "scripts": {
-    "build": "truffle compile --all && tsc -b",
+    "build": "truffle compile --all",
     "migrate": "truffle migrate --reset",
-    "test": "tsc -b && truffle test --network ganache $npm_package_config_testFiles",
-    "test:windows": "tsc -b && truffle test --network ganache %npm_package_config_testFiles%",
-    "clean": "rm -rf dist build networks/7777777.json",
+    "test": "tsc -b && yarn ganache:start && yarn truffle:test && yarn ganache:stop",
+    "test:windows": "tsc -b && truffle test",
```
cf19drofxots: Is the ganache bootstrap reflected in the windows script?

packages/contracts/package.json
```diff
@@ -33,7 +31,9 @@
     "solidity"
   ],
   "devDependencies": {
+    "@counterfactual/types": "1.0.0",
```
cf19drofxots: Versioning that package at `1.0.0` seems premature

packages/contracts/test/app-registry-new.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    ).deploy({ gasLimit: 6e9 });
+
+    await appRegistry.deployed();
+  });
+
+  it("is possible to call setState to put state on-chain", async () => {
+    // Test AppInterface
+    const appInterface = new AppInterface(
+      AddressZero,
+      hexlify(randomBytes(4)),
```
cf19drofxots: In other places we avoid using random values. We should be consistent here too.

packages/contracts/test/app-registry.spec.ts
```diff
@@ -0,0 +1,230 @@
+import { ethers } from "ethers";
```
cf19drofxots: Import necessary sub-modules.

packages/contracts/test/app-registry.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+  AssetType,
+  computeStateHash,
+  expect,
+  Terms
+} from "./utils";
+
+import { ALICE, BOB } from "./constants";
+
+// HELPER DATA
+const TIMEOUT = 30;
```
cf19drofxots: What is this timeout referring to? In the context of contracts and an async testing suite, it can be confusing to infer.

packages/contracts/test/app-registry.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+      });
+
+      it("should work with much higher nonce", async () => {
+        expect(await latestNonce()).to.eq(0);
+        await setStateAsOwner(1000);
+        expect(await latestNonce()).to.eq(1000);
+      });
+
+      it("shouldn't work with an equal nonce", async () => {
+        // @ts-ignore
```
cf19drofxots: Ideal to not include ignores.

packages/contracts/test/conditional-transaction.spec.ts
cf19drofxots: Yeah, truffle forces usage of Mocha :(

packages/contracts/test/constants.ts
```diff
@@ -0,0 +1,13 @@
+import { ethers } from "ethers";
```
cf19drofxots: Import style.

packages/contracts/test/utils/index.ts
```diff
@@ -0,0 +1,205 @@
...
+
+export const expect = chai.use(solidity).expect;
+
+export enum AssetType {
+  ETH,
+  ERC20,
+  ANY
+}
+
+export class App {
```
cf19drofxots:
```suggestion
export class AppInstance {
```

packages/contracts/test/utils/index.ts
```diff
@@ -0,0 +1,205 @@
...
+            token: this.token
+          }
+        ]
+      )
+    );
+  }
+}
+
+// TS version of MAppRegistryCore::computeStateHash
+// export const;
```
cf19drofxots: `// export const` ?

packages/contracts/test/utils/index.ts
```diff
@@ -0,0 +1,205 @@
...
+            uint256 defaultTimeout
+          )`
+        ],
+        [this.appIdentity]
+      )
+    );
+  }
+}
+
+// export class OffChainState {
```
cf19drofxots: Remove if not needed.

yarn.lock
```diff
@@ -594,6 +594,254 @@
   resolved "https://registry.yarnpkg.com/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz#2b5a3ab3f918cca48a8c754c08168e3f03eba61b"
   integrity sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw==
 
+<<<<<<< HEAD
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: incorrectly resolved merge conflict

packages/machine/src/protocol-types-tbd.ts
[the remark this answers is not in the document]
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: maybe we can do

```
interface Json {
	[x: string]: string | number | boolean | Json | JsonArray;
}

interface JsonArray extends Array<string | number | boolean | Json | JsonArray> { }
```

IIRC there is a constraint similar to this anyway since `newState` is sent over the wire

modified from https://github.com/Microsoft/TypeScript/issues/1897#issuecomment-271146964

packages/machine/src/models/state-channel.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  public get numberOfApps() {
+    return this.appInstances.size;
+  }
+
+  public getAppInstance(appInstanceId: string) {
+    const appInstance = this.appInstances.get(appInstanceId);
+
+    if (appInstance === undefined) throw Error(Errors.APP_DOES_NOT_EXIST);
```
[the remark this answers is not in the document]
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `${Errors.APP_DOES_NOT_EXIST}({appInstanceId})` should work

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

package.json
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why can't they just be added back?

packages/machine/src/models/state-channel.ts
```diff
@@ -0,0 +1,221 @@
...
+      AppInstance
+    >([]),
+    private readonly freeBalanceAppIndexes: Readonly<
+      Map<AssetType, string>
+    > = new Map<AssetType, string>([]),
+    private readonly monotonicallyIncreasingSeqNo: number = 0
+  ) {}
+
+  public get sequenceNumber() {
+    return this.monotonicallyIncreasingSeqNo;
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: IMO the simplest naming scheme is:

- `numActiveApps` - number of active apps - increases when an app is installed, decreases when an app is uninstalled
- `numInstalledApps` - increases when app is installed, does not decrease when app is uninstalled

the second one could also plausibly be called `monotonicNumInstalledApps`

packages/machine/src/models/state-channel.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export class StateChannel {
+  constructor(
+    public readonly multisigAddress: string,
+    public readonly multisigOwners: string[],
+    private readonly appInstances: Readonly<Map<string, AppInstance>> = new Map<
+      string,
+      AppInstance
+    >([]),
+    private readonly freeBalanceAppIndexes: Readonly<
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: AFAICT `Readonly<Map<T>` is equivalent to `Map<T>` (ref: https://github.com/Microsoft/TypeScript/blob/v2.8.1/lib/lib.es6.d.ts#L1351), which is different from `ReadonlyMap<T>` (ref: https://github.com/Microsoft/TypeScript/blob/v2.8.1/lib/lib.es6.d.ts#L4703)

packages/types/src/index.ts
```diff
@@ -0,0 +1,61 @@
+import { ethers } from "ethers";
+
+export enum AssetType {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: there are now 4 (!) `enum AssetType`s in this codebase...

packages/machine/src/models/state-channel.ts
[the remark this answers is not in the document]
snario: Ah I see. I kind of prefer it tbh, it's clear that the only thing changing are the balances

packages/machine/src/models/app-instance.ts
```diff
@@ -0,0 +1,231 @@
...
+  latestState: object;
+  latestNonce: number;
+  latestTimeout: number;
+  hasBeenUninstalled: boolean;
+};
+
+/**
+ * Representation of an AppInstance.
+ *
+ * @property owner The address of the multisignature wallet on-chain for the
```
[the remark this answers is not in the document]
snario: https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format

packages/machine/src/ethereum/utils/free-balance.ts
```diff
@@ -0,0 +1,106 @@
...
+      0,
+      // TODO: See line 56...
+      latestState,
+      latestNonce,
+      latestTimeout
+    );
+  }
+
+  public get formattedState(): ETHBucketAppState {
+    return {
```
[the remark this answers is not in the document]
snario: If I do this I get the following error: 

`Type 'object' must have a '[Symbol.iterator]()' method that returns an iterator.ts(2488)`

packages/apps/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@counterfactual/apps",
+  "version": "0.0.1",
+  "description": "Collection of various apps built on Counterfactual",
+  "repository": "github.com/counterfactual/monorepo",
+  "license": "MIT",
+  "private": true,
+  "files": [
```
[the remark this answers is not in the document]
snario: Yes -- the idea is that this package exports the build ABI + bytecode of the apps

packages/apps/test/nim.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+      };
+
+      const action = {
+        pileIdx: 0,
+        takeAmnt: 7
+      };
+
+      await expect(
+        nim.functions.applyAction(preState, action)
+        // @ts-ignore
```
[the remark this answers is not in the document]
snario: `revertedWith` is not typed by `ethereum-waffle`. Need to add a rule to `typescript-typings` manually.

packages/cf.js/package.json
```diff
@@ -27,10 +26,10 @@
     "typescript": "^3.1.2"
   },
   "dependencies": {
-    "@counterfactual/contracts": "0.0.2",
+    "@counterfactual/contracts": "0.0.3",
     "@counterfactual/common-types": "0.0.1",
     "cuid": "^2.1.4",
-    "ethers": "^4.0.17",
+    "ethers": "4.0.18",
```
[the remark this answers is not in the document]
snario: Yes -- I think its better to pin when possible. While writing this PR ethers upgraded and triggered a very confusing bug I spent hours trying to fix.

packages/cf.js/src/legacy/eth-balance-refund-app.ts
```diff
@@ -1,4 +1,5 @@
-import * as ETHBalanceRefundAppContract from "@counterfactual/contracts/build/contracts/ETHBalanceRefundApp.json";
+import BuildArtifact from "@counterfactual/contracts/build/contracts/ETHBalanceRefundApp.json";
```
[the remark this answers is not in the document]
snario: Why?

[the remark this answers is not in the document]
snario: I think 0x is in a different position where people frequently reference their ABIs in their code and it might even ship as part of 0x.js

packages/contracts/contracts/default-apps/ETHBucket.sol
```diff
@@ -0,0 +1,39 @@
+pragma solidity 0.4.25;
+pragma experimental "ABIEncoderV2";
+
+import "@counterfactual/contracts/contracts/libs/Transfer.sol";
```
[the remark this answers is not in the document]
snario: We should always import via modules in my opinion

packages/contracts/contracts/libs/LibStateChannelApp.sol
```diff
@@ -0,0 +1,54 @@
+pragma solidity ^0.4.25;
```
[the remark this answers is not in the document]
snario: ehhh do you mean `LibStateChannelAppInstance`? maybe just call it `LibAppInstance`?

packages/contracts/contracts/mixins/MAppCaller.sol
```diff
@@ -0,0 +1,86 @@
...
+import "../libs/LibSignature.sol";
+import "../libs/LibStateChannelApp.sol";
+import "../libs/LibStaticCall.sol";
+import "../libs/Transfer.sol";
+
+
+/// @title MAppCaller
+/// @author Liam Horne - <liam@l4v.io>
+/// @notice A mixin for the AppRegistry to make staticcalls to Apps
+contract MAppCaller is LibSignature, LibStateChannelApp {
```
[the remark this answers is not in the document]
snario: `M<name>` is an abstract contract and `Mixin<name>` implements the abstract contract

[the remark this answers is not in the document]
snario: I think its explicit name is helpful since it is _only_ useful in this context

packages/contracts/contracts/libs/LibStateChannelApp.sol
```diff
@@ -0,0 +1,54 @@
...
+  struct AppIdentity {
+    address owner;
+    address[] signingKeys;
+    bytes32 appInterfaceHash;
+    bytes32 termsHash;
+    uint256 defaultTimeout;
+  }
+
+  // A minimal structure representing the state of an App from POV of the blockchain
+  struct AppChallenge {
```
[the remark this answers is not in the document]
snario: I think "challenge" is more descriptive because it includes the state of the app, the timeout of the challenge, the current nonce, who submitted the challenge etc. Perhaps the variable should be renamed to `appChallenges` in the mapping

packages/contracts/contracts/mixins/MAppRegistryCore.sol
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  /// @notice Compute a unique hash for a state of this state channel and application
+  /// @param _id The unique hash of an `AppIdentity`
+  /// @param appStateHash The hash of a state to be signed
+  /// @param nonce The nonce corresponding to the version of the state
+  /// @param timeout A dynamic timeout value representing the timeout for this state
+  /// @return A bytes32 hash of the arguments encoded with the signing keys for the channel
+  function computeStateHash(
+    bytes32 _id,
```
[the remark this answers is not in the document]
snario: I've always just been hesitant to rename variables to `id` because it feels like `id` should be a reserved keyword ... but it actually isn't.

packages/contracts/contracts/mixins/MixinSetStateWithAction.sol
snario: Not a modifier but a require statement

https://github.com/counterfactual/monorepo/blob/9dad2a7f4b39f678f7b579734220e1bf31c2928d/packages/contracts/contracts/mixins/MixinSetStateWithAction.sol#L57-L60

packages/contracts/test/utils/index.ts
snario: This class actually should probably re-use the model in the machine package. I wrote it before I wrote the other one.

packages/apps/test/nim.spec.ts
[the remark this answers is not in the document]
snario: done 491cea7e

snario:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll @cf19drofxots the tests in the `node` package are still failing for unrelated reasons

package.json
cf19drofxots: @snario that's a pretty useful test (as @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll mentioned) - let's either not remove it or make a tangible PR to add it back that's referenced here, so as to not lose track of it.

packages/apps/test/tictactoe.spec.ts
```diff
@@ -1,30 +1,43 @@
...
+chai.use(waffle.solidity);
 
-contract("TicTacToe", (accounts: string[]) => {
-  let game: ethers.Contract;
+const { expect } = chai;
 
-  const stateEncoding =
-    "tuple(address[2] players, uint256 turnNum, uint256 winner, uint256[3][3] board)";
+type TicTacToeAppState = {
+  players: string[];
```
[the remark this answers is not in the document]
cf19drofxots: I'm all for more consistent types, which is why I made the earlier comment. Even if we do punt this to later, at that point it'd be easier to track what needs to be updated as there'll be more relevant use cases of the types being updated.

This isn't a blocker but quite preferable.

packages/cf.js/src/legacy/eth-balance-refund-app.ts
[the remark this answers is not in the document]
cf19drofxots: Because it's not ideal having imports being generated at runtime. What other instance of an import is dynamically generated?

packages/contracts/contracts/libs/LibStateChannelApp.sol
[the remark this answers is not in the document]
cf19drofxots: Sure, that seems fine

packages/contracts/contracts/mixins/MAppCaller.sol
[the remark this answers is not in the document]
cf19drofxots: You mean `M<name>` is a base class?

The functions seem to have implementations, and for eg, `MixinAppCaller` doesn't exist. It's used in `MixinSetResolution` but that contract doesn't implement or override any of the functions here.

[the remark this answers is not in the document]
cf19drofxots: Still weird naming convention IMO, though not a blocker.

packages/cf.js/src/legacy/eth-balance-refund-app.ts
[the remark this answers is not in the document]
snario: What do you mean runtime? Personally I really like the clear seperation of source code, the build step to transform source into runnable executsles, and then the run/test steps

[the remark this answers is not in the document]
snario: We could just start publishing to npm now so the artifacts are available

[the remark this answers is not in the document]
cf19drofxots: Yeah it's slightly different use cases. I suppose from a consumer POV the artifacts are like any other thing being imported (i.e. if you don't run `yarn` you don't get the needed dependencies, if you don't run `<compile contracts>` you don't get the artifacts)

On one hand it _is_ nice generating them with each build to ensure bottom-up correctness. On the other hand, the contracts don't change that often which was the motivation for my earlier comment "build once, commit, and reuse artifacts"

packages/machine/test/unit/models/state-channel/uninstall-app.spec.ts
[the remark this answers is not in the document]
joelalejandro: Yeah, but as @cf19drofxots pointed out on his review, this is inconsistent with other tests. We should probably hardcode.

packages/machine/src/models/app-instance.ts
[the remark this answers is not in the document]
joelalejandro: @snario No, it would not be compliant. But why are we using Natspec on TypeScript files?

cf19drofxots:
@snario the build & test scripts compile the contracts twice btw

cf19drofxots:
Running this locally errors out:
```
@counterfactual/contracts: TypeError: Cannot read property 'gasUsed' of null
@counterfactual/contracts:     at names.forEach.name (/Users/nima/Counterfactual/monorepo/node_modules/eth-gas-reporter/gasStats.js:392:52)
@counterfactual/contracts:     at Array.forEach (<anonymous>)
@counterfactual/contracts:     at files.forEach.file (/Users/nima/Counterfactual/monorepo/node_modules/eth-gas-reporter/gasStats.js:371:11)
@counterfactual/contracts:     at Array.forEach (<anonymous>)
@counterfactual/contracts:     at Object.mapMethodsToContracts (/Users/nima/Counterfactual/monorepo/node_modules/eth-gas-reporter/gasStats.js:367:9)
@counterfactual/contracts:     at Runner.Gas.runner.on (/Users/nima/Counterfactual/monorepo/node_modules/eth-gas-reporter/index.js:119:65)
@counterfactual/contracts:     at Runner.emit (events.js:187:15)
@counterfactual/contracts:     at start (/Users/nima/Counterfactual/monorepo/node_modules/mocha/lib/runner.js:826:10)
@counterfactual/contracts:     at Runner.run (/Users/nima/Counterfactual/monorepo/node_modules/mocha/lib/runner.js:854:5)
@counterfactual/contracts:     at Mocha.run (/Users/nima/Counterfactual/monorepo/node_modules/mocha/lib/mocha.js:572:17)
@counterfactual/contracts:     at /Users/nima/Counterfactual/monorepo/node_modules/truffle/build/webpack:/packages/truffle-core/lib/test.js:120:1
@counterfactual/contracts:     at process._tickCallback (internal/process/next_tick.js:68:7)
@counterfactual/contracts: error Command failed with exit code 1.
@counterfactual/contracts: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

cf19drofxots:
CI config needs to be updated as well to resolve the changes introduced in this PR

packages/cf.js/src/index.ts
```diff
@@ -3,7 +3,17 @@ import * as types from "./types";
...
+  AppFactory,
+  Provider,
+  legacy,
+  types,
+  utils
+};
+
+export { Provider, legacy, types, utils, cf };
+
+export default cf;
```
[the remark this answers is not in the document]
snario: I was just mirroring `ethers` ... not really any reason in particular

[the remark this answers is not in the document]
ebryn: This is a bad practice IMO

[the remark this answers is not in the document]
snario: I think I agree -- I'll undo this

packages/machine/src/ethereum/install.ts
```diff
@@ -0,0 +1,75 @@
...
+      value: 0,
+      data: iface.functions.executeAppConditionalTransaction.encode([
+        this.networkContext.AppRegistry,
+        this.networkContext.NonceRegistry,
+        uninstallKey,
+        appInstanceId,
+        this.terms
+      ]),
+      operation: MultisigOperation.DelegateCall
+    };
```
ebryn: Why don't we construct a `MultisigTransaction` object here instead of returning a POJO?

packages/machine/src/ethereum/set-state.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  public transaction(sigs: Signature[]): Transaction {
+    return {
+      to: this.networkContext.AppRegistry,
+      value: 0,
+      data: iface.functions.setState.encode([
+        this.appIdentity,
+        this.getSignedStateHashUpdate(sigs)
+      ])
+    };
```
ebryn: Why don't we construct a `Transaction` object here instead of returning a POJO?

packages/machine/src/ethereum/set-state.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  private getSignedStateHashUpdate(
+    signatures: Signature[]
+  ): SignedStateHashUpdate {
+    return {
+      stateHash: keccak256(this.appState),
+      nonce: this.appLocalNonce,
+      timeout: this.timeout,
+      signatures: signaturesToSortedBytes(this.hashToSign(), ...signatures)
+    };
```
ebryn: Ditto above. If we have types, I feel like we should actually `new SignedStateHashUpdate(...)` instead of using POJOs.

packages/machine/src/ethereum/utils/free-balance.ts
```diff
@@ -0,0 +1,69 @@
+import ETHBucket from "@counterfactual/contracts/build/contracts/ETHBucket.json";
+import { AppInterface, ETHBucketAppState } from "@counterfactual/types";
+import { AddressZero, MaxUint256 } from "ethers/constants";
+import {
+  defaultAbiCoder,
+  // formatParamType,
```
ebryn: remove?

packages/machine/src/ethereum/utils/multisend-encoder.ts
```diff
@@ -0,0 +1,12 @@
+import { defaultAbiCoder } from "ethers/utils";
+
+import { MultisigTransaction } from "./types";
+
+const ENCODING = ["uint256", "address", "uint256", "bytes"];
+
+export const encodeTransactions = (txs: MultisigTransaction[]) =>
```
ebryn: I don't like arrow function abuse. IMO this should be:

`export function encodeTransactions(txs: MultisigTransaction[]) {`

packages/machine/src/ethereum/install.ts
snario: There is no `MultisigTransaction` object, it is only an `interface`

packages/machine/src/ethereum/set-state.ts
snario: There is no `Transaction` object, it is only an `interface`

packages/machine/src/ethereum/set-state.ts
snario: What is the benefit of wrapping `type`s/`interface`s as `classes` besides being able to syntactically instantiate them with `new`?

packages/machine/src/ethereum/utils/free-balance.ts
snario: I want to keep it here just because there is a big `TODO` further down in the code that will dramatically benefit from the usage of this and I think it's useful to keep in mind. Maybe unnecessary though and that should just be an issue instead..

packages/machine/src/ethereum/utils/multisend-encoder.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why is the use of arrow functions "abuse"?

packages/machine/test/integration/setup-then-set-state.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+      pass: received.eq(argument),
+      message: () => `expected ${received} not to be equal to ${argument}`
+    };
+  }
+});
+
+// TODO: This will be re-used for all integration tests, so
+//       move it somewhere re-usable when we add a new test
+beforeAll(async () => {
+  require("dotenv-safe").config();
```
cf19drofxots: The `require` is inconsistent with how other dependencies are used

package.json
```diff
@@ -14,35 +14,19 @@
   },
   "scripts": {
     "build": "sh build.sh",
-    "clean": "lerna run clean --parallel --no-bail",
+    "clean": "git clean -Xdf --exclude=\"!.env\"",
```
cf19drofxots: Why not add `.env` to the gitignore?

packages/machine/test/integration/setup-then-set-state.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+
+// The AppRegistry.setState call _could_ be estimated but we haven't
+// written this test to do that yet
+const SETSTATE_COMMITMENT_GAS = 6e9;
+
+let networkId: number;
+let provider: JsonRpcProvider;
+let wallet: Wallet;
+
+// TODO: When we add a second use case of this custom mathcer,
```
cf19drofxots: So why is this custom matcher needed? It's not explained in the comments

packages/machine/test/integration/setup-then-set-state.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+    const users = signingKeys.map(x => x.address);
+
+    const proxyFactory = new Contract(
+      ProxyFactory.networks[networkId].address,
+      ProxyFactory.abi,
+      wallet
+    );
+
+    proxyFactory.on("ProxyCreation", async proxy => {
+      let stateChannel = new StateChannel(proxy, users).setupChannel(network);
```
cf19drofxots: Why is `setupChannel` not a static method that'd return a new StateChannel instance?

Giving `users` to such a function more strongly parallels how the multisig contract gets setup as well.

packages/machine/README.md
```diff
@@ -1,6 +1,6 @@
 # [Machine](https://counterfactual.com) <img align="right" src="https://static1.squarespace.com/static/59ee6243268b96cc1fb2b14a/t/5af73bca1ae6cf80fc1cc250/1529369816810/?format=1500w" height="80px" />
 
-This is the typescript implementation of the [Counterfactual protocol](https://github.com/counterfactual/specs/blob/master/v0/protocols.md). It's responsible for executing these protocols, _producing state commitments_, and thereby effectively facilitating user interaction for off-chain channelized applications.
+This is the core functional TypeScript implementation of the [Counterfactual protocol](https://github.com/counterfactual/specs/blob/master/v0/protocols.md). It is responsible for executing the protocols described by Counterfactual and producing correctly constructed signed commitments that correspond to state transitions of the users' state channels.
```
cf19drofxots: Are the words `core` and `functional` necessary here?

`This is the TypeScript implementation...` seems sufficient.

It's also weird phrasing to say "the protocols described by Counterfactual". Anything wrong with "executing the Counterfactual protocols"?

package.json
snario: It is in the `.gitignore` -- this is saying, do not delete the `.env` when doing a clean. Usually it has secrets you would prefer to keep

packages/machine/test/integration/setup-then-set-state.spec.ts
snario: To be able to use `ethers.utils.BigNumber.prototype.eq` which I think is self-explanatory

packages/machine/test/integration/setup-then-set-state.spec.ts
snario: Because it requires a protocol execution to go from "not set up" to "set up" and I think these are two distinct stages in the lifecycle of a `StateChannel`

packages/machine/README.md
snario: 7b82b340d3456acf807ff35d2aa4c511708604ff

packages/machine/test/integration/setup-then-set-state.spec.ts
snario: I've done it this way to mimic the way it is most often used in example code, docs, other codebases etc. We could make it conform to the `import` style but somehow I feel this is more expressive and easy to read? What do you think

cf19drofxots:
```suggestion
import("dotenv-safe").config()
```

Does the same thing and retains consistent style. This is not just a code aesthetics thing, the entire codebase uses ES6 import/exports so doing a CommonJS require feels out of place.

packages/machine/src/models/state-channel.ts
```diff
@@ -0,0 +1,239 @@
...
+    );
+
+    appInstances.delete(appToBeUninstalled.id);
+
+    appInstances.set(
+      fb.id,
+      fb.setState({ ...currentState, aliceBalance, bobBalance })
+    );
+
+    return new StateChannel(
```
ebryn: It doesn't feel right to me that we return new instances of `StateChannel` objects from each of these methods. It seems like the `StateChannel` object reference should be stable across all these operations.

snario: Do you mean that the class should not be immutable or that the implementation should hide the fact that it is immutable?

ebryn: @snario It depends. It could just be a naming issue. Are these objects purely internal to the machine or are they vended out to the Node / Playground Server as well?

snario: At the moment, purely internal. However, the intention is to use them in the `Node`. It is unclear how they will be distributed. My recommendation is to make the `machine` a subfolder of `Node` and hoist the `models` folder up into the `node` source code

ebryn: @snario Then I think they should probably be like how CF.js's `AppInstance` objects work in that state channels are singletons. API consumers only see one consistent state channel object.

PR #167, opened 2018-11-05
Migrate CfOpSetup to not use Multisend but instead Multisig.execTransaction

mvanderh:
### Description

Migrate CfOpSetup to not use Multisend but instead Multisig.execTransaction.

- Add abstract superclass CfMultisigTxOp: a CfOp that calls execTransaction() on a multisig 
- Migrate CfMultisendOp and CfOpSetup to use CfMultisigTxOp
- Add test to check that CfOpSetup transaction is multisig execTransaction 

### Related issues

Resolves https://github.com/counterfactual/specs/issues/26


mvanderh:
Yess! That CF prefix has annoyed me forever

packages/machine/src/middleware/cf-operation/cf-op-setup.ts
```diff
@@ -18,11 +17,14 @@ import {
...
-export class CfOpSetup extends CfMultiSendOp {
+export class CfOpSetup extends MultisigTxOp {
   /**
    * Helper method to get hash of an input calldata
+   * @param networkContext
    * @param multisig
-   * @param multisigInput
+   * @param freeBalanceStateChannel
+   * @param freeBalance
+   * @param dependencyNonce
```
snario: 1. Why is there a docstring here but nowhere else in this source code
2. If there is a docstring it should contain descriptions next to the parameters

PR #169, opened 2018-11-05
[WIP] [contracts] add root nonce support to contracts 

snario:
Description
This PR adds contract support for root nonces (counterfactual/specs#11)

TODO
=====
- [ ]  Ensure tests pass on contracts
- [ ]  Implement changes onto the machine's cfoperations
- [ ]  Ensure tests pass after implementing changes to cfoperations


snario:
I've begun work on integrating this into the repo. Some notes:

The contracts package needs to have some more tests that test this functionality. @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll can you please add these?

The machine's cf-operations layer had to be updated, specifically the setup and install protocols according to what @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll added in the specs.

The addition of the extra argument is not a problem but it changes what the state channel objects need to keep track of. Now, in addition to the free balances, multisig address, etc; we need to track the root nonce. There is some work to be done here namely in the implementation of the wallet PoC. @cf19drofxots is looking into this on top of the commit i just made: counterfactual/monorepo@27c1d99

I believe this introduced the need for this cleanup protocol which also needs to be implemented. @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll can you please comment further.

cf19drofxots:
Is this a WIP? If so, it should be labelled so, as the check boxes in the description aren't checked off.

cf19drofxots:
@snario Status update?

snario:
> @snario Status update?

@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll is working on it.

Issue #137, opened 2018-10-21
[contracts] Set up Typechain to set up typings for contracts in tests


snario:
It would be nice to have typings for the contracts when we test them in the `contracts` package.

I spent some time trying this out, but there were a few limitations that will have to be considered before we fully integrate it into the code base.

1. At the moment the set of typings in the related package [`truffle-typings`](https://github.com/ethereum-ts/truffle-typings#readme) are fairly limited. For example, there is no `address` property on a `TruffleContract` object. We still use `TruffleContract` objects in the repo, for example when we use [`artifacts.require`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/test/integration/twoPartyPayments.spec.ts#L109). We'll need to extend that package or just extend our own package( [`packages/typescript-typings`](http://github.com/counterfactual/monorepo/blob/master/packages/typescript-typings)).

2. In many places, we actually don't even use the `TruffleContract` object, we actually use `ethers.Contract`. Thus, we need to wait until [this pull request](https://github.com/ethereum-ts/TypeChain/pull/112) is merged into Typechain adding support for `ethers.js` as a "target".

snario:
Since Typechain is getting rid of ethers support I am closing this.

Issue #144, opened 2018-10-24
[machine] ActionName shouldn't mix operations and client actions


joelalejandro:
https://github.com/counterfactual/monorepo/blob/1a20e08f2f07f5765e718dd5ab5c1ef2d017a0a0/packages/machine/src/types.ts#L425-L438

PR #170, opened 2018-11-05
[WIP][contracts] Add typechain to contracts tests

snario:
This is to implement [counterfactual/monorepo#144](https://github.com/counterfactual/monorepo/issues/144).

There are a handful of problems before we can merge this though:

- [ ]  TypeChain doesn't have support for return types that are structs, which forces us to [define our own return type](https://github.com/counterfactual/monorepo/compare/liam/typechain#diff-418f3593ded3b2dba0a06acebdbd3afeR47) which kind of defeats the point. Added an issue on the TypeChain repo ([ethereum-ts/TypeChain#119](https://github.com/ethereum-ts/TypeChain/issues/119)). Let's wait until they add support for this.

-  [ ] The ethers target on TypeChain also doesn't support extra options passed into function calls like `gasLimit` for example. Documented here: [ethereum-ts/TypeChain#118](https://github.com/ethereum-ts/TypeChain/issues/118). It means [lines like these fail type checking](https://github.com/counterfactual/monorepo/blob/liam/typechain/packages/contracts/test/unit/transfer.spec.ts#L72).

Minor issues:

-  [ ]  The same problem described [here](https://github.com/ethereum-ts/TypeChain/issues/103) regarding things like `BigNumber` not being included affects us because `ethers` function calls usually expect these kinds of types. Doesn't matter for small numbers ([as you can see](https://github.com/counterfactual/monorepo/compare/liam/typechain?expand=1#diff-4cde56b88bcdc4ea77487ee2c4279624R52) I can just replace them with `number` types) but for large numbers (e.g., when we use [`UNIT_ETH`](https://github.com/counterfactual/monorepo/blob/liam/typechain/packages/contracts/test/integration/twoPartyPayments.spec.ts#L51)) it does matter.

- [ ]  Solidity `bytes` are types as `(string | number[])[]` right now. There's [an issue](https://github.com/ethereum-ts/TypeChain/issues/98) to change this. It affects us because the `signMessage` library returns a `string` for signatures so that also fails type checking. Either need to get typechain to allow `string` or change how `signMessage` works to produce an array of bytes.


snario:
Fixes #137

snario:
It seems like TypeChain is removing ethers support ... https://github.com/ethereum-ts/TypeChain/issues/123#issuecomment-443548517

Sad!

PR #171, opened 2018-11-05
[machine] More unit tests for protocol commitments

snario:
We should add several more unit tests for every cf-operation in accordance to the specs. Good task for anyone that wants to understand in detail the commitments we are signing, too.

Tests
===
- [x]  setState
- [ ]  install
- [ ]  uninstall
- [x]  setup

Goal
===
For this pull-request we should test every `hashToSign` method and probably have ~50 total unique tests.


packages/machine/test/unit/cf-operations/set-state.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+describe("CfOpSetState", () => {
+  const TEST_NETWORK_CONTEXT = new NetworkContext(
+    ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+    ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+    ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+    ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+    ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+    ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+    ethers.utils.hexlify(ethers.utils.randomBytes(20)),
+    ethers.utils.hexlify(ethers.utils.randomBytes(20))
```
mvanderh: It's better not to use random input for your tests unless you're explicitly testing what your code does under random inputs. I would recommend using "0x11111.." etc instead

packages/machine/test/unit/cf-operations/set-state.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      ]).functions.proxyCall.encode([
+        TEST_NETWORK_CONTEXT.registryAddr,
+        app.cfAddress(),
+        new ethers.utils.Interface([
+          "setState(bytes32,uint256,uint256,bytes)"
+        ]).functions.setState.encode([
+          TEST_APP_STATE_HASH,
+          TEST_LOCAL_NONCE,
+          TEST_TIMEOUT,
+          `0x${sig1.toString().substr(2)}${sig2.toString().substr(2)}`
```
mvanderh: I reckon this will sporadically break because you're not considering the order of these (random) signatures. This is a good reason not to use random inputs.

mvanderh: Actually I take that back, the private keys aren't random. Still nonrandom input is better

mvanderh:
I'm commandeering this PR

packages/machine/test/unit/cf-operations/set-state.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        ])
+      ])
+    );
+  });
+
+  // https://github.com/counterfactual/specs/blob/master/v0/protocols.md#digest
```
snario: I think this URL is out of date now

packages/machine/test/integration/cf-operations.spec.ts
```diff
@@ -190,7 +180,7 @@ describe("Setup Protocol", async () => {
       values
     );
 
-    const appInstance = new CfAppInstance(
+    const cfStateChannel = new cf.app.CfAppInstance(
```
snario: Unnecessary change of variable name?

packages/machine/test/unit/cf-operations/set-state.spec.ts
```diff
@@ -79,12 +37,20 @@ describe("CfOpSetState", () => {
     const digest = op.hashToSign();
     const vra1 = TEST_SIGNING_KEYS[0].signDigest(digest);
```
snario: This variable is typo'd, should be `vrs1` not `vra1`

packages/machine/test/unit/cf-operations/set-state.spec.ts
```diff
@@ -79,12 +37,20 @@ describe("CfOpSetState", () => {
...
     const vra2 = TEST_SIGNING_KEYS[1].signDigest(digest);
-    const sig1 = new Signature(vra1.recoveryParam as number, vra1.r, vra1.s);
-    const sig2 = new Signature(vra2.recoveryParam as number, vra2.r, vra2.s);
+    const sig1 = new cf.utils.Signature(
```
snario: It should probably be possible to pass in `vra1` directly into the constructor rather than decomposing it. Perhaps using spread operator like `new Signature(...vra1)`

packages/cf.js/src/utils/signature.ts
```diff
@@ -49,6 +49,10 @@ export class Signature {
     return signaturesList;
   }
 
+  public static fromEthersSignature(sig: ethers.utils.Signature): Signature {
+    return new Signature(sig.recoveryParam as number, sig.r, sig.s);
+  }
```
snario: Wait, there's an `ethers.utils.Signature`? Then I would recommend we strictly use this class versus re-inventing our own.

packages/machine/test/unit/cf-operations/set-state.spec.ts
snario: https://specs.counterfactual.com/04-setup-protocol#commitments

packages/cf.js/src/utils/signature.ts
mvanderh: Yeah, that's what `signDigest()` returns. The only annoying thing about it is it doesn't have a method to convert to bytestring.

packages/cf.js/src/utils/peer-balance.ts
```diff
@@ -1,4 +1,4 @@
-import { ethers } from "ethers";
+import * as ethers from "ethers";
```
snario: I think the preferred is `import ethers from "ethers"`

PR #183, opened 2018-11-06
[machine] Polishes instantiation of VM and removes if-wrapping around sendResponse() 

joelalejandro:
### Description

Adds some minor improvements to the VM code. See commits for more info.

### DIFF
--- a/packages/machine/src/vm.ts
+++ b/packages/machine/src/vm.ts
@@ -122,9 +122,13 @@ export class CounterfactualVM implements Observable {
   }
 
   public receive(msg: cf.node.ClientActionMessage): cf.node.WalletResponse {
-    this.validateMessage(msg);
+    if (!this.validateMessage(msg)) {
+      throw new Error("Cannot receive invalid message");
+    }
+
     const action = new Action(msg.requestId, msg.action, msg);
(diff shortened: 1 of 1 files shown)


packages/machine/src/vm.ts
```diff
@@ -173,10 +176,12 @@ export class CounterfactualVM implements Observable {
...
     if (!execution.action.isAckSide) {
-      this.responseHandler.sendResponse(
-        new cf.node.Response(execution.action.requestId, status)
-      );
+      return;
     }
+
+    this.responseHandler.sendResponse(
+      new cf.node.Response(execution.action.requestId, status)
+    );
```
snario: Why the change?

joelalejandro: It's usually not good practice to have a single if-condition wrapping the entirety of a function. If for any reason, a function must abort its execution, it's preferred to do so as quickly as possible.

snario: Shouldn't the if clause be reversed though?

joelalejandro: Yeah, you're right on that, forgot to remove the `!`.

PR #188, opened 2018-11-07
[WIP] [contracts] Upgrade to Solidity 0.5

snario:
This is a proof of concept pull request that shouldn't be merged, but shows how `abi.decode` is useful in our context. Some of the core benefits:

- Applications can have dynamic types in their defined `AppState`
- Actions can be defined to have different data encoded within them per action type (which means you don't need to encode _all possible_ action data in a single struct anymore)
- The `AppInstance` doesn't need to rely on ABIEncoderV2 struct encoding concatenation to hack around `abi.decode` anymore and we can make regular Solidity cross-contract function calls
- `getTurnTaker` can now accept `signingKeys` since it can accept dynamic types
- We won't need to store the four `bytes4` sighashes in `AppDefinition` anymore

# TODO
- [x] Need to use a newer version of `solc` I think that supports `abi.decode`
- [ ] Replace the actual functions in `AppInstance.sol` entirely
- [ ] Re-write apps in appDefinitions folder to use new style
- [ ] Remove sighashes from being necessary in AppInterface


cf19drofxots:
@snario still on hold?

snario:
> @snario still on hold?

Yeah.

snario:
I'm gonna redo this after #163 is merged

PR #189, opened 2018-11-07
[machine] Improves serializer to a more readable, scalable solution

joelalejandro:
The new serializer code works with pattern matching in a declarative style, allowing to inject new use cases as needed with no need of altering the deserialize() function.

It also allows to clearly see what happens for each data type.


joelalejandro:
Sure thing @cf19drofxots!

joelalejandro:
Closing this in favor of a cleaner branch. Got lots of merge conflicts here.

PR #191, opened 2018-11-08
[playground] Adds initial scaffolding and core component structure for the Playground

joelalejandro:
### Description

This PR creates the `playground` package in the monorepo. It uses Web Components and TypeScript, bundled via Webpack + some custom scripting.

#### Scaffolding

Every Web component has its own folder:

```
src
|- components
|  |- app
|  |  |- app.css         // Component stylesheet
|  |  |- app.ts          // Component main class (extends from "Component")
|  |  |- index.ts        // Component exporter
| ...
```

#### Component structure

Any component can be created using this pattern:

```typescript
// my-component.ts
export default class MyComponent extends Component {
  getComponentName() {
    return "my-component";
  }
}
```

```html
<template id="my-component">
  <span class="my-component__label"></span>
</template>
```

The `Component` abstract class takes care of binding automatically the template and rendering it into its shadow DOM automatically.

In order for a component to be exposed, it must be included in the `./src/components/index.ts` file.

#### Building

The build script `build.js` concatenates all of the HTML templates into a single file, which is then injected via `HTMLWebpackPlugin`. We're not using `link rel="import"` to load them due to Firefox not supporting it. It will also copy all styles to the `dist/` folder.

#### Outstanding issues

- [x] **Testing**: JSDom does not support the CustomElements API, so we need to use Puppeteer in Jest. I'm still researching how to do that.
- [x] **Build chain**: The build process isn't great; I'm not too familiar with Webpack and HMR, so at the moment every script change will trigger a browser refresh, and template/style changes will restart `webpack-dev-server`. Looking for better ways to do this.
  - Update: **Will continue this in a different PR.**


joelalejandro:
At the moment, the only test I managed to succeed in writing is a Puppeteer-driven test that checks if the `playground-app` component is instantiated in the URL exposed by `webpack-dev-server`.

joelalejandro:
Closing this in favor of a more opinionated approach with Stencil.js.

PR #192, opened 2018-11-08
Remove Signatures utility class, instead use ethers.utils.Signature and related functions

mvanderh:
### Description

Remove Signatures utility class, instead use ethers.utils.Signature and related functions. This decreases the complexity of CF.js utils at low cost. Using ethers.utils.Signature also makes it easier to understand for outsiders.

- Remove Signatures utility class
- Add two utility functions for Signatures: signaturesToBytes and signaturesToSortedBytes
- Replace usages of Signature with these functions

### Related issues

Inspired by #171


packages/cf.js/src/utils/serializer.ts
```diff
@@ -28,7 +26,7 @@ const deserializeArray = data => data.map(value => deserialize(value));
 
 const deserializeBigNumber = data => ethers.utils.bigNumberify(data._hex);
 
-const deserializeSignature = data => new Signature(data.v, data.r, data.s);
+const deserializeSignature = data => data;
```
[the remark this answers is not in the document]
cf19drofxots: Wouldn't we want to instantiate an ethers Signature using the data being passed in? Otherwise this deserialization seems to just pass through the data that _would be_ in the signature but not an ethers Signature instance itself.

packages/iframe-wallet-poc/src/iframe/user.ts
```diff
@@ -249,14 +249,14 @@ async function signMyUpdate(
...
+): Promise<ethers.utils.Signature> {
   const operation: machine.cfTypes.CfOperation = machine.middleware.getFirstResult(
     machine.instructions.Instruction.OP_GENERATE,
     context.results
   ).value;
   const digest = operation.hashToSign();
-  const sig = user.signingKey.signDigest(digest);
-  return new cf.utils.Signature(sig.recoveryParam! + 27, sig.r, sig.s);
+  const { recoveryParam, r, s } = user.signingKey.signDigest(digest);
+  return { r, s, v: recoveryParam! + 27 };
```
cf19drofxots: Is returning this object equivalent to returning an instance of an `ethers.utils.Signature`?

Even if so, wouldn't be easier to parse if a signature instance is constructed using this return data and have the instance be returned?

packages/cf.js/src/node/index.ts
```diff
@@ -1,5 +1,7 @@
+import { ethers } from "ethers";
```
snario: We should be consistently using `import ethers from "ethers"` @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll

packages/cf.js/src/utils/signature.ts
```diff
@@ -1,69 +1,26 @@
...
-  // https://github.com/counterfactual/monorepo/issues/124
-  constructor(readonly v: number, readonly r: string, readonly s: string) {}
-
-  public recoverAddress(digest: H256): Address {
-    return ethers.utils.recoverAddress(digest, this.toString());
-  }
+import { H256 } from "./index";
+
+export function signaturesToBytes(
+  ...signatures: ethers.utils.Signature[]
```
snario: Why use spread operator? Seems weird

packages/cf.js/src/utils/signature.ts
```diff
@@ -1,69 +1,26 @@
...
+    .join("");
+  return `0x${signaturesHexString}`;
+}
 
-  public toString(): string {
-    return `0x${this.r.substr(2)}${this.s.substr(2)}${ethers.utils
-      .hexlify(this.v)
-      .substr(2)}`;
-  }
+export function signaturesToSortedBytes(
```
snario: The existence of this function makes me think we aren't doing a good job of keeping local data in context through operations; we shouldn't need to ecrecover signatures we should just know which order is correct

packages/machine/test/integration/test-commitment-store.ts
```diff
@@ -193,13 +195,7 @@ export class TestCommitmentStore {
...
     }
 
-    const sigs = [signature, counterpartySignature].map(sig => {
-      if (!(sig instanceof cf.utils.Signature)) {
-        const { v, r, s } = sig as any;
-        return new cf.utils.Signature(v, r, s);
-      }
-      return sig;
-    });
+    const sigs = [signature, counterpartySignature];
```
snario: Probably unnecessary to declare a variable here and can just pass it into the function below inline

snario: ala https://techbeacon.com/why-unnecessary-variables-are-bad-your-code

packages/machine/test/integration/test-response-sink.ts
```diff
@@ -157,16 +168,14 @@ export class TestResponseSink implements cf.node.ResponseSink {
...
       Instruction.OP_GENERATE,
       context.results
     ).value;
     const digest = operation.hashToSign();
-    const sig = this.signingKey.signDigest(digest);
-    return Promise.resolve(
-      new cf.utils.Signature(sig.recoveryParam! + 27, sig.r, sig.s)
-    );
+    const { recoveryParam, r, s } = this.signingKey.signDigest(digest);
+    return Promise.resolve({ r, s, v: recoveryParam! + 27 });
```
snario: Does this need to be wrapped in a Promis.resolve? Also the return type is not an `ethers.utils.Signature` anymore

packages/cf.js/src/utils/serializer.ts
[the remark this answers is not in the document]
mvanderh: ethers.utils.Signature is just a Typescript interface, not a class. Anything with a v, r and s property is considered an ethers.utils.Signature

packages/cf.js/src/utils/signature.ts
mvanderh: Makes passing a single signature look nicer, which is a pretty common use case.

packages/machine/test/integration/test-response-sink.ts
mvanderh: Yeah, `signMyUpdate` is expected to return a Promise.

Also, if that were the case (return type =/= `ethers.utils.Signature`) then the Typescript build would fail, which it doesn't. ethers.utils.Signature is just any object with an `r`,`s` and `v`

packages/cf.js/src/utils/signature.ts
mvanderh: Yup, agreed. It's pretty innocent right now though, it's only used when generating multisig transactions.

packages/iframe-wallet-poc/src/iframe/user.ts
mvanderh: > Is returning this object equivalent to returning an instance of an ethers.utils.Signature?
Yup

There is no such thing as a "Signature instance", it's just an interface now.

packages/cf.js/src/utils/serializer.ts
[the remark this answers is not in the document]
cf19drofxots: Got it - in which case @joelalejandro's comment is more useful here

PR #194, opened 2018-11-09
[contracts] fixed misleading require error message

jtakalai:


packages/contracts/contracts/appDefinitions/Nim.sol
```diff
@@ -42,7 +42,6 @@ contract Nim {
     pure
     returns (bytes)
   {
-    require(0 <= action.pileIdx, "pileIdx was invalid");
```
jtakalai: Actually scratch that, it's actually checking it's not zero. But that's... curious, since state.pileHeights[0] is used later on, inside `isWin`. Specifically `isWin` suggests that pileIdx should be 0, 1, or 2. Why would `action.pileIdx == 0` be invalid?

jtakalai: It's even tested: https://github.com/counterfactual/monorepo/blob/master/packages/contracts/test/integration/nim.spec.ts#L31 creates an action with pileIdx == 0, and applies it in the next statement. Really interesting that the test still passes. Hmm. I admit, my changes were simply a result of reading the code and editing in GitHub, not actually running the tests. But of course you guys have CI set up as it should be, and it will run the tests for sure  (;

snario: cc-ing @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll who wrote this

snario:
@jtakalai can you make another commit? Turned on CircleCi for external forks.

PR #195, opened 2018-11-09
[playground] Adds initial Stencil-based scaffolding

joelalejandro:
### Description

This PR is an alternate version of #191, built using [Stencil.js](https://stenciljs.com). It is a more opinionated, yet still thin layer on top of Web Components.

The current setup includes a testing suite with Jest + Puppeteer and some basic root components.


ebryn:
LGTM

packages/playground/README.md
```diff
@@ -0,0 +1,29 @@
+# [Playground](https://counterfactual.com) <img align="right" src="https://static1.squarespace.com/static/59ee6243268b96cc1fb2b14a/t/5af73bca1ae6cf80fc1cc250/1529369816810/?format=1500w" height="80px" />
+
+This is an environment to run dApps using [CF.js](../cf.js). It allows to showcase different demo apps, presenting a variety of use cases where state channels are applicable.
```
snario: dApps -> state channel based decentralized applications

packages/playground/src/components/app-home/app-home.tsx
```diff
@@ -0,0 +1,17 @@
+import { Component } from "@stencil/core";
+
+@Component({
+  tag: "app-home",
+  styleUrl: "app-home.css",
+  shadow: true
+})
+export class AppHome {
+  render() {
+    // tslint:disable-next-line:prettier
```
snario: Why?

packages/playground/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "declaration": false,
+    "experimentalDecorators": true,
+    "lib": [
+      "dom",
+      "es2015"
+    ],
+    "moduleResolution": "node",
+    "module": "esnext",
+    "target": "es2017",
+    "noUnusedLocals": true,
```
snario: I think this one is a duplication from root config

packages/playground/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "noUnusedParameters": true,
```
snario: This one too? if not, it should be in root

packages/playground/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "jsx": "react",
+    "jsxFactory": "h"
+  },
+  "include": [
+    "src"
+  ],
+  "exclude": [
```
snario: Should be in root

packages/playground/src/components.d.ts
```diff
@@ -0,0 +1,65 @@
+/* tslint:disable */
+/**
+ * This is an autogenerated file created by the Stencil compiler.
```
snario: If it is auto generated do we need to commit it? Shouldn't it be part of some kind of build process

packages/playground/src/components/app-home/app-home.tsx
joelalejandro: At first, because it forced me to use a self-closing tag for an empty DIV, which isn't correct AFAIK.

packages/playground/tsconfig.json
joelalejandro: Probably, will clean up and see what happens.

packages/playground/tsconfig.json
joelalejandro: Will apply it in a different PR, there are lots of unused variables in other packages and will make this PR too big.

packages/playground/src/components/app-root/app-root.css
```diff
@@ -0,0 +1,15 @@
+header {
```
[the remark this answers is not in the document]
joelalejandro: @patience-tema-baron Yes, the CSS is scoped to the component. In regards to its children, you can play around with the [`contain`](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) property.

PR #198, opened 2018-11-09
[contracts] Contract existence check

cf19drofxots:
### Description

This checks whether the contract being made a call to via the registry actually exists or not.

It also adds better error messaging when a signature being used is invalid.

### DIFF
--- a/packages/contracts/contracts/Registry.sol
+++ b/packages/contracts/contracts/Registry.sol
@@ -1,11 +1,14 @@
 pragma solidity 0.4.25;
 
+import "openzeppelin-eth/contracts/utils/Address.sol";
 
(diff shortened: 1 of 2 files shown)


packages/contracts/contracts/Registry.sol
```diff
@@ -1,12 +1,15 @@
 pragma solidity 0.4.25;
 
+import "openzeppelin-solidity/contracts/AddressUtils.sol";
```
snario: Should be `openzeppelin-eth`

cf19drofxots:
I remember debugging some contracts code a while ago, and came across an instance when the code was calling a contract that didn't exist (don't remember the exact details of the context), so I made this change but didn't include it earlier.

PR #199, opened 2018-11-09
Add CFjs 0.0.2 API reference

mvanderh:
### Description

Rough outline of API for CFjs "0.0.2" AKA the next iteration. Please give feedback.

### Related issues

Relocated from https://github.com/counterfactual/specs/pull/34


packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+        - `async getTerms(): [assetType, limit, tokenAddress]`
+    - App lifecycle
+        - `on(eventType, callback)`
+            - EventType: stateUpdate, uninstall
+- `types`
+    - `NodeProvider` (external)
+        - (Injected into webpage)
+        - Instance methods
+            - `postMessage(message)`
+            - `onMessage(callback)`
```
cf19drofxots: Should `onMessage` expect to register a callback for a specific message?

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+        - `appActionEncoding`: ABI encoding for App Action
+    - `AppManifest`
+        - `name`: human-readable name of app e.g. "TicTacToe"
+        - `version`: semantic version of app definition contract
+        - `definition: AppDefinition`
+    - `AppInfo`
+        - `definition: AppDefinition`
+        - `appId: AppId`
+        - `terms: [assetType, limit, tokenAddress]`
+        - `addresses: Address[]`
```
cf19drofxots: What are `addresses` for an AppInfo?

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+    - `NodeProvider` (external)
+        - (Injected into webpage)
+        - Instance methods
+            - `postMessage(message)`
+            - `onMessage(callback)`
+    - `AppDefinition`
+        - `address`: on-chain address for the app definition contract
+        - `appStateEncoding`: ABI encoding for App State
+        - `appActionEncoding`: ABI encoding for App Action
+    - `AppManifest`
```
cf19drofxots: Can we add clarification on how AppManifest is effectively different than AppInfo?

Seems like AppInfo has the same information, and more.

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
+# API V0.0.2
+## `cf.js`
+
+- `Client`
+    - Properties
+        - `nodeProvider`
+    - Instance methods
+        - `getAppInfos(): AppInfo[]`
```
cf19drofxots: I think the term `AppInfo` arose for some historic reason(s) that I'm not entirely sure about.

Any thoughts on renaming this to just `App`? Seems cleaner (i.e. `getApps()`)

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+        - `createAppFactory(appDefinition: AppDefinition): AppFactory`
```
cf19drofxots: Will the `AppFactory` that creates an instance of its app implicitly give the app a handle on the `nodeProvider` from the client so that the app can talk to the node?

packages/cf.js/API_REFERENCE.md
cf19drofxots: Oh are these the app-specific signing keys?

packages/cf.js/API_REFERENCE.md
snario: Seems odd to have this method versus using the 'new' operstor. @ebryn

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+    - Client lifecycle
+        - `on(eventType, callback)`
+            - EventType: proposeInstall, install, rejectInstall
```
snario: Are these all event types?

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+        - `async install(appId: AppID): Promise<AppInstance>`
+- `AppInstance`
+    - Properties
+        - `id: AppID` — Identifier for this specific app instance
+        - `definition: AppDefinition`
+    - Instance methods
+        - `async applyAction(action): AppState`
+            - `action`: JSON-encoded object congruent with AppAction struct
+            - Returns ABI decoded representation of the latest signed state of the app.
+        - `async proposeState(state)`: Propose a state to countersign
```
snario: How is state typed? Ideally it can be a POJO that is auto encoded using the app's app state encoding

packages/cf.js/API_REFERENCE.md
mvanderh: Yeah. First the idea was to use usernames, which we could still do if we think it's worth it in the short term. Alternatively addresses are an easy solution

packages/cf.js/API_REFERENCE.md
mvanderh: I think they should be the same.

packages/cf.js/API_REFERENCE.md
mvanderh: Argh but then the difference between App and AppInstance becomes confusing.

packages/cf.js/API_REFERENCE.md
mvanderh: Yes to what @cf19drofxots said. It's somewhat odd but less odd than passing the injected `nodeProvider` to each instance of AppFactory and AppInstance to communicate with the node.

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+        - `async uninstall()`
+            - Uninstall the app
+        - `async getTerms(): [assetType, limit, tokenAddress]`
```
patience-tema-baron: As an alternative to getting the terms asynchronously after the app's instantiation, could we just provide the terms (along with other app instance properties like `AppId`) in the `onInstall` notification? @cf19drofxots

packages/cf.js/API_REFERENCE.md
mvanderh: Yeah

packages/cf.js/API_REFERENCE.md
mvanderh: I imagined it as a POJO myself. 

Side note, the existence of this method means there have to be two extra methods: rejectState and acceptState.

packages/cf.js/API_REFERENCE.md
mvanderh: I don't know. Doesn't matter from CF.js' perspective, we just consume whatever the node injects. Who owns the node provider and what do they think?

packages/cf.js/API_REFERENCE.md
snario: I'm confused why this is in the app manifest

packages/cf.js/API_REFERENCE.md
patience-tema-baron: Should we also have an `error` event type? For instance, if the user submits an invalid `proposeInstall`.

packages/cf.js/API_REFERENCE.md
patience-tema-baron: It's not the manifest, but in `AppInfo`

packages/cf.js/API_REFERENCE.md
cf19drofxots: Yeah, I remember there being some friction here. Moving on with `AppInfo`.

As for the terms, what's the rationale for an `AppInstance` not having it as a property, and for it needing to be fetched? IIUC an `AppInstance` is instantiated when the app actually gets installed, by which time the terms should be known and agreed to, and hence the state of the app instance should include this info. Is there some nuance I'm missing here?

packages/cf.js/API_REFERENCE.md
mvanderh: I would say so @patience-tema-baron

cf19drofxots: Who would be a recipient of such an event?

Ideally clients (by which I mean the node) should do verification to confirm the validity of a proposal on sending/receiving.

Let's say the receiver of a proposal is notified via the `proposeInstall` event, and upon checking the proposal realizes the proposal is invalid. That seems to just be a `rejectInstall` response, with specific details about the rejection in the payload being passed back.

On the other hand, let's say the sender wants to send a proposal (which is invalid). Ideally the sender is immediately notified (i.e. awaits the "send-proposal" call) via a verification process by the node that the proposal is invalid.

packages/cf.js/API_REFERENCE.md
cf19drofxots: Right. Seems like it should, but cc'ing @joelalejandro

patience-tema-baron: @cf19drofxots just to clarify, are you imagining something like `onMessage(type, cb)`, so that we'd `onMessage('install', () => {...})` and `onMessage('stateUpdate', () => {...})`?

packages/cf.js/API_REFERENCE.md
patience-tema-baron: Exactly. I'd imagine the `error` notification to be this latter case, useful for disambiguating a system response from a peer response. That being said, perhaps they could both fall under the umbrella of `rejectInstall` and (as you suggest) we would just differentiate between them with the payload.

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,54 @@
...
+    - App lifecycle
+        - `on(eventType, callback)`
+            - EventType: stateUpdate, uninstall
```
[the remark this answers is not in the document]
patience-tema-baron: Agreed. What are your thoughts on adding a `getState` method so that returning users (after a page refresh, for instance) can correctly initialize state?

packages/cf.js/API_REFERENCE.md
patience-tema-baron: Earlier, when I said the only difference between an `AppInstance` and `AppInfo` is the terms, I was wrong. `AppInfo` also has the app's `name`, which is necessary for the playground to display in its 'active apps' list. The playground may want other manifest.json data too, such as an icon or description. If we make `AppInstances` and `AppInfos` the same thing, we'll probably want to include some sort of reference to the app's manifest in its properties.

packages/cf.js/API_REFERENCE.md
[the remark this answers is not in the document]
snario: Makes sense to me

packages/cf.js/API_REFERENCE.md
ebryn: At Devcon, I mentioned it was odd to have an object that you can conditionally call one or the other methods on
(`applyAction` vs `proposeState`) based on how the contract was written. It seems like we need to do something about that here. I had proposed perhaps we have two different `AppInstance` classes to handle the API differences between them. /cc @snario

packages/cf.js/API_REFERENCE.md
cf19drofxots: @patience-tema-baron yeah

mvanderh:
@patience-tema-baron NOTE: Add a `whoAmI()` or `getName()` method on Client that gets the user's name from the node.

patience-tema-baron:
> @patience-tema-baron NOTE: Add a `whoAmI()` or `getName()` method on Client that gets the user's name from the node.

Posted this on slack, but it might be better to have the conversation here:

Here's another thing to think about re: matchmaking. As I understand it, matchingmaking happens through the playground, not the node. If we add a method like `getRandomUser()` to cf.js, then cf.js becomes an interface for both the node and the playground, which I assume we don't want.

The same concern is true for usernames. These are managed by the playground as opposed to the node, and probably we don't want to pollute cf.js's api to support a concerns beyond the node.

I wonder if we should have a separate playground interface to manage these interactions from within the app?

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,50 @@
+# API V0.0.2
+## `cf.js`
+
+- `Client`
+    - Properties
+        - `nodeProvider`
+    - Instance methods
+        - `getApps(): AppInstance[]`
```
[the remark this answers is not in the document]
patience-tema-baron: I'm thinking of ways to support this. The loosest approach would be to group `AppInstances` by their app name or `manifestAddress`. This has a significant security flaw, though: a dapp could query a different dapp's `manifestAddress`, thereby getting the other dapp's `AppInstances`.

Another approach would be to assign an id to an instantiate `AppFactory`, so that when the `AppFactory` installs an `AppInstance`, the `AppInstance` can be assigned the `AppFactory`'s id. The `AppFactory` could then have its own `getApps` method which collects all the `AppInstances` with that `AppFactory` id.

This has implications for the way channels are set up, and I'm curious what @cf19drofxots thinks about that. I'm also not a fan of assigning an id to a factory. It feels like there should be another layer between the factory and the instances.

packages/cf.js/API_REFERENCE.md
mvanderh: @ebryn Do you have any input here?

packages/cf.js/API_REFERENCE.md
mvanderh: Ping @joelalejandro

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,50 @@
...
+    - Properties
+        - `appDefinition: AppDefinition`
+    - Instance methods
+        - `async proposeInstall(peerAddress, terms): Promise<AppID>`
+        - `async install(appId: AppID): Promise<AppInstance>`
+- `AppInstance`
+    - Properties
+        - `id: AppID` — Identifier for this specific app instance
+        - `definition: AppDefinition`
+        - `state: object` - the decoded app state, updated automatically during `stateUpdate` notifications
```
mvanderh: NOTE: Move this to an async getState() method

packages/cf.js/API_REFERENCE.md
```diff
@@ -0,0 +1,50 @@
+# API V0.0.2
+## `cf.js`
+
+- `Client`
+    - Properties
+        - `nodeProvider`
+    - Instance methods
```
mvanderh: NOTE: Add async getUsername()

packages/cf.js/API_REFERENCE.md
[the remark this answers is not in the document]
mvanderh: For the sake of demo, we can get away with giving out all the apps for now. Should add a note in the spec about future security measures for authenticating apps and the playground.

Issue #208, opened 2018-11-12
[monorepo] Cache files based on package.json not on yarn.lock 


snario:
The CI breaks when yarn.lock is computed using a different version of `yarn` from `1.10.1`. I think a fix to this should be to change `circle.yaml` to cache based on the various `package.json` files instead of caching based on the `yarn.lock` file.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
why would an out-of-sync node_modules cause the build to fail?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
FYI: the recommended thing to cache is the yarn cache directory, not the `node_modules` folders: https://circleci.com/docs/2.0/yarn/#caching

snario:
#209 fixed this specific issue. Probably should look into @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll's comment above though.

cf19drofxots:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll Do you know what the effective difference is?

The article doesn't mention how saving the yarn cache is different than saving the `node_modules` folders. I even verified that the content of the packages being saved is identical between the two (not really surprising).

So curious as to why following their recommendation is necessarily better.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
the deeper issue is that the `yarn` command should not go from a failing state to a passing state just because `node_modules` is changed...

cf19drofxots:
It didn't go from a failing to passing state because `node_modules` was changed.

It was because the local yarn version was different from the one used in CI, producing a `yarn.lock` version of a different format. Since saving/restoring the dependencies (regardless of whether that's `~/.cache/yarn` or `node_modules`) relied on the checksum of the `yarn.lock` file, using different `yarn.lock` files resulted in the fails, not the actual cache.

cf19drofxots:
But we're digressing, this is far from pressing. Can revisit when/if this becomes an actual issue.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
> It didn't go from a failing to passing state because node_modules was changed.

I must be understanding it wrong, I thought the only purpose of the `restore-cache` step is to populate the `node_modules` directory. The only effect of changing the cache key on the test job, is that a different `node_modules` will be used, right?

PR #209, opened 2018-11-12
[monorepo] Update caching key

cf19drofxots:
### Description

This updates the caching key used to not rely on a specific version of Yarn.

This gets rid of this problem: https://github.com/counterfactual/monorepo/pull/201#issuecomment-437900936

### Related issues

https://github.com/counterfactual/monorepo/issues/208


mvanderh:
Hmm, kind of a double edged sword though. Now there's a possibility of yarn version-specific bugs.

cf19drofxots:
@mvanderh hmm yeah - though seems like that'd have a smaller surface area affecting our use cases than the current setup.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
per my comment in https://github.com/counterfactual/monorepo/issues/208 - using package.json as the cache key is very weird

PR #212, opened 2018-11-13
use STATE_TRANSITION_PROPOSE

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
replace the magic number `0` with `STATE_TRANSITION_PROPOSE`


cf19drofxots:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll what's the status of this? WIP or needs review?

packages/iframe-wallet-poc/src/iframe/user.ts
```diff
@@ -152,7 +152,10 @@ export class User implements machine.mixins.Observable, cf.node.ResponseSink {
   }
 
   public generateObserverNotification(notification: cf.node.Notification): any {
-    return notification.data.results.find(result => result.opCode === 0).value;
+    const PROPOSE = machine.instructions.Opcode.STATE_TRANSITION_PROPOSE;
+    return notification.data.results.find(result => {
```
joelalejandro: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll There's no need to expand the arrow function, we can still do:

```ts
return notification.data.results.find(
  result => result.opCode === PROPOSE
).value;
```

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: yeah an old linter rejected that...

PR #214, opened 2018-11-13
[machine] Unit test for Install protocol operation

mvanderh:
### Description

More test coverage and protection against regression.


packages/machine/package.json
```diff
@@ -14,6 +14,7 @@
     "build": "tsc -p tsconfig.test.json && rollup -c",
     "test": "tsc -b && jest --runInBand --detectOpenHandles --bail",
     "test-debug": "node --inspect-brk jest --runInBand",
+    "test-debug-ide": "node $NODE_DEBUG_OPTION ./node_modules/.bin/jest --runInBand",
```
mvanderh: This is really useful for those of us using JetBrains IDEs as it allows you to use the visual debugger. Huge productivity boost

snario: Does this work with VS Code too?

mvanderh: Not sure.

packages/machine/test/unit/cf-operations/fixture.ts
```diff
@@ -68,3 +58,70 @@ export const TEST_FREE_BALANCE = new cf.utils.FreeBalance(
...
+  TEST_NETWORK_CONTEXT,
+  TEST_MULTISIG_ADDRESS,
+  TEST_PARTICIPANTS,
+  TEST_APP_INTERFACE,
+  TEST_TERMS,
+  TEST_TIMEOUT,
+  TEST_APP_UNIQUE_ID
+);
+
+export function constructContractCall(funcSig: string, ...args: any): string {
```
cf19drofxots: Is the string `funcSig` referring to some abi-style signature?

Might be useful to add a couple doc lines to this function to clarify what the args/return are.

packages/machine/test/unit/cf-operations/fixture.ts
```diff
@@ -68,3 +58,70 @@ export const TEST_FREE_BALANCE = new cf.utils.FreeBalance(
...
+  return constructContractCall(
+    "multiSend(bytes)",
+    `0x${subcalls.map(call => call.substr(2)).join("")}`
+  );
+}
+
+export function constructMultiSendSubCall(
+  operation: "delegatecall" | "call",
+  to: string,
+  value: string | number,
```
cf19drofxots: bigNumber here instead?

packages/machine/test/unit/cf-operations/fixture.ts
```diff
@@ -68,3 +58,70 @@ export const TEST_FREE_BALANCE = new cf.utils.FreeBalance(
...
+  funcSig: string,
+  args: any[]
```
cf19drofxots: Can we tighten the type here? What needs to be supported here besides string and perhaps bigNumber?

packages/machine/test/unit/cf-operations/fixture.ts
```diff
@@ -68,3 +58,70 @@ export const TEST_FREE_BALANCE = new cf.utils.FreeBalance(
...
+  return ethers.utils.defaultAbiCoder.encode(
+    ["uint8", "address", "uint256", "bytes"],
+    [operation === "delegatecall" ? 1 : 0, to, value, data]
+  );
+}
+
+export function constructMultisigExecTransaction(
+  operation: "delegatecall" | "call",
+  to: string,
+  value: string | number,
```
cf19drofxots: `value` isn't being used in this function.

packages/machine/test/unit/cf-operations/install.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  const salt = keccak256(
+    cf.utils.abi.encodePacked(["uint256"], [nonceUniqueId])
+  );
+  const uninstallKey = keccak256(
+    cf.utils.abi.encodePacked(
+      ["address", "uint256", "uint256"],
+      [TEST_MULTISIG_ADDRESS, 0, salt]
+    )
+  );
+  const { assetType, limit, token } = TEST_TERMS;
```
cf19drofxots: Can we pass `TERMS_TERMS` directly into the `constructMultiSendSubCall` below instead of extracting out its fields?

packages/machine/test/unit/cf-operations/set-state.spec.ts
```diff
@@ -4,6 +4,7 @@ import { ethers } from "ethers";
 import { OpSetState } from "../../../src/middleware/protocol-operation";
 
 import {
+  constructContractCall, TEST_APP_INSTANCE,
```
cf19drofxots: This should get caught by the linter - need to be consistent in terms of importing, so all imports should be on their own line.

packages/machine/test/unit/cf-operations/setup.spec.ts
```diff
@@ -4,15 +4,38 @@ import { ethers } from "ethers";
 import { OpSetup } from "../../../src/middleware/protocol-operation";
 
 import {
+  constructContractCall, constructMultisigExecTransaction,
```
cf19drofxots: Same comment as above re: imports.

packages/machine/test/unit/cf-operations/fixture.ts
```diff
@@ -68,3 +58,70 @@ export const TEST_FREE_BALANCE = new cf.utils.FreeBalance(
   100,
   new cf.utils.Nonce(true, 0, 5)
 );
+
+export const TEST_FREE_BALANCE_APP_INTERFACE = cf.utils.FreeBalance.contractInterface(
+  TEST_NETWORK_CONTEXT
```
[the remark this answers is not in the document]
mvanderh: Easier debuggability! Now if i see "1111111.." somewhere in the tx data i know it's referencing the Registry

packages/machine/test/unit/cf-operations/install.spec.ts
mvanderh: No, it expects an array. I prefer it this way

packages/machine/test/unit/cf-operations/set-state.spec.ts
mvanderh: Yeah weird, I ran a lint:fix on it too

packages/machine/test/unit/cf-operations/fixture.ts
mvanderh: I suppose it's technically more correct, but that would make it more annoying to use in tests. Honestly I could probably drop the value parameter everywhere and just hardcode 0

packages/machine/test/unit/cf-operations/fixture.ts
mvanderh: I use what ethers uses for contract call arguments.

packages/machine/test/unit/cf-operations/fixture.ts
mvanderh: Oooops

PR #215, opened 2018-11-13
[machine][node] Move state out of the machine

cf19drofxots:
### Description

This is an initial PR to move the state handling out of the machine (as the machine is a stateless instruction executor), while introducing the node package which takes over the responsibility of handling said state.

- The machine now does not depend on the node package.
- This also introduces slightly better type safety around operation generation.

Remainder:
- [ ] update the iframe wallet to reflect these changes

Some of this will probably have to be refactored mainly due to:
- https://github.com/counterfactual/monorepo/pull/226
- https://github.com/counterfactual/monorepo/pull/163


packages/machine/src/instruction-executor.ts
```diff
@@ -31,15 +31,15 @@ export class InstructionExecutor implements Observable {
...
+  public node: Node;
 
   // Observable
   public observers: Map<NotificationType, Function[]> = new Map();
 
   constructor(config: InstructionExecutorConfig) {
     this.responseHandler = config.responseHandler;
-    this.nodeState = new NodeState(config.state || {}, config.network);
-    this.middleware = new Middleware(this.nodeState, config.opGenerator);
+    this.node = new Node(config.state || {}, config.network);
```
snario: This seems like a bizarre decision, why would this class have a Node?

cf19drofxots: Handle to the context it's executing in.

snario: Can't we pass in only the minimum amount of information it needs? Like what does this class actually require to know?

cf19drofxots: We can, hence the WIP :)

.circleci/config.yml
```diff
@@ -6,25 +6,26 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: dependency-cache-{{ checksum "packages/cf.js/package.json" }}{{ checksum "packages/contracts/package.json"}}{{ checksum "packages/iframe-wallet-poc/package.json"}}{{ checksum "packages/machine/package.json" }}{{  checksum "packages/playground/package.json" }}{{ checksum "packages/typescript-typings/package.json" }}
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why change the key?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
can you do a separate PR to rename `NodeState => Node` and `nodeState => node`. most of this PR is this change and it's hard to review.

cf19drofxots:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll ack - will do

.circleci/config.yml
cf19drofxots: The PR introduces a new package so the key needs to be updated.

.circleci/config.yml
```diff
@@ -6,25 +6,26 @@ jobs:
...
+          key: dependency-cache-{{ checksum "packages/cf.js/package.json" }}{{ checksum "packages/contracts/package.json"}}{{ checksum "packages/iframe-wallet-poc/package.json"}}{{ checksum "packages/machine/package.json" }}{{  checksum "packages/playground/package.json" }}{{ checksum "packages/typescript-typings/package.json" }}{{ checksum "packages/node/package.json" }}
       - run: yarn
       - save_cache:
-          key: dependency-cache-{{ checksum "packages/cf.js/package.json" }}{{ checksum "packages/contracts/package.json"}}{{ checksum "packages/iframe-wallet-poc/package.json"}}{{ checksum "packages/machine/package.json" }}{{  checksum "packages/playground/package.json" }}{{ checksum "packages/typescript-typings/package.json" }}
+          key: dependency-cache-{{ checksum "packages/cf.js/package.json" }}{{ checksum "packages/contracts/package.json"}}{{ checksum "packages/iframe-wallet-poc/package.json"}}{{ checksum "packages/machine/package.json" }}{{  checksum "packages/playground/package.json" }}{{ checksum "packages/typescript-typings/package.json" }}{{ checksum "packages/node/package.json" }}
```
snario: Does this thing support globs?

.circleci/config.yml
```diff
@@ -6,25 +6,26 @@ jobs:
...
           paths:
             - node_modules
             - packages/contracts/node_modules
             - packages/iframe-wallet-poc/node_modules
             - packages/machine/node_modules
             - packages/cf.js/node_modules
             - packages/playground/node_modules
+            - packages/node/node_modules
```
snario: Also globs?

packages/cf.js/src/legacy/utils/index.ts
```diff
@@ -2,7 +2,7 @@ import ethers from "ethers";
 import * as _ from "lodash";
 
 import * as abi from "./abi";
-import { FreeBalance } from "./free-balance";
+import { EMPTY_FREE_BALANCE, FreeBalance } from "./free-balance";
```
snario: Seems like a weird choice to export this like this. Shouldn't it be in a `constants` subdirectory like `ethers.constants`?

packages/machine/src/instruction-executor.ts
```diff
@@ -1,11 +1,11 @@
 import * as cf from "@counterfactual/cf.js";
+import { Node } from "@counterfactual/node";
```
snario: This still seems like a massive antipattern to me. The `Node` should not be consumed by the `machine`. We should mark it as TO-FIX or something or just find a way to not do this in the first place.

packages/machine/src/middleware/state-transition/install-proposer.ts
```diff
@@ -1,9 +1,9 @@
 import * as cf from "@counterfactual/cf.js";
+import { Node, StateChannelInfoImpl } from "@counterfactual/node";
```
snario: Also really dislike this dependency.

packages/node/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@counterfactual/node",
+  "version": "0.0.1",
+  "main": "dist/index.js",
+  "types": "dist/src/index.d.ts",
+  "module": "dist/index.es.js",
+  "iife": "dist/index-iife.js",
```
snario: Are all these various export methods actually necessary? We don't actually import this into the Playground environment yet so why don't we hold off on prematurely packaging it up until we need to?

packages/node/test/unit/channels.spec.ts
```diff
@@ -0,0 +1,48 @@
+import * as cf from "@counterfactual/cf.js";
+import { ethers } from "ethers";
+import * as _ from "lodash";
+
+import { Node, StateChannelInfoImpl } from "../../src/node";
+
+describe("Node can contain channels", () => {
+  it("Node has no channels", () => {
```
snario: Should write unit tests for basic instantiation e.g., https://github.com/gnosis/pm-js/blob/master/test/test_gnosis.js#L17-L35

packages/node/test/unit/channels.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+        cf.legacy.utils.EMPTY_FREE_BALANCE
+      )
+    };
+    const node = new Node(channels, cf.legacy.network.EMPTY_NETWORK_CONTEXT);
+    expect(_.size(node.channelStates)).toEqual(1);
+  });
+
+  it("Node has multiple channels", () => {
+    const channels = {
+      [ethers.Wallet.createRandom().address]: new StateChannelInfoImpl(
```
snario: `ethers.Wallet.createRandom().address` => `ethers.utils.hexlify(ethers.utils.randomBytes(20))`

However as @mvanderh mentioned the other day, it's better to not literally compute random values each test.

packages/node/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "sourceRoot": ".",
+    // TODO: Document why we need declarationDir here
+    "declarationDir": "dist/",
```
snario: Do we need `declarationDir`?

packages/node/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "outDir": "dist/"
+  },
+  // TODO: Figure out how to add "test" here so they get linted by tslint when using
+  // the -p flag but *also* such that rollup doesn't include tests in the build.
```
snario: If you remove `rollup` then we can get rid of this TODO.

packages/node/tsconfig.test.json
```diff
@@ -0,0 +1,12 @@
+{
```
snario: Do we need to have two of these?

packages/cf.js/src/legacy/utils/index.ts
cf19drofxots: Yeah that's better.

packages/machine/src/instruction-executor.ts
cf19drofxots: Not gonna mark as `FIXME` - rather not do it this way in the first place and introduce it to the codebase.

packages/machine/src/middleware/state-transition/install-proposer.ts
cf19drofxots: Will also remove/refactor.

packages/node/test/unit/channels.spec.ts
cf19drofxots: I recall that as well. Generally 👍 to not compute random values each time.

packages/node/tsconfig.json
cf19drofxots: I do not think we do. No package consumes its typings.

packages/node/tsconfig.test.json
cf19drofxots: The only difference here is `"include": ["src", "test"],` which allows for compiling the tests, but not rolling up tests (since the `tsconfig.json` is used for rollup)

snario: Seems like the rolls should be reversed then, e.g., `tsconfig.rollup.json` should be the edge case and `tsconfig.json` be the one that includes `src` and `test`

cf19drofxots: That works too. Could either be `tsconfig.test.json` or `tsconfig.rollup.json`

I prefer your suggestion actually.

.circleci/config.yml
cf19drofxots: Removed this formatting completely as CircleCI yaml parsing doesn't support interpreting globs.

https://github.com/counterfactual/monorepo/pull/235

cf19drofxots:
The commits to setup the Node package have been cherry-picked into another PR: https://github.com/counterfactual/monorepo/pull/236/files

cf19drofxots:
Closing this PR as 1) it's still blocked and 2) a lot of changes have occurred.

Will open another later when more appropriate.

PR #217, opened 2018-11-14
pass artifactName

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR makes `AbstractContract`s, if initialized from artifact names, remember the artifact name and print it out in error messages


packages/contracts/utils/contract.ts
```diff
@@ -42,13 +42,15 @@ export class AbstractContract {
    */
   public static async fromBuildArtifact(
     buildArtifact: BuildArtifact,
-    links?: { [name: string]: Promise<AbstractContract> }
+    links?: { [name: string]: Promise<AbstractContract> },
+    artifactName?: string
```
joelalejandro: If the argument will have a default value, it should be declared here:

```ts
artifactName: string = "UntitledContract"
```

This way you don't need to deal with the nuances of `T | undefined`.

PR #218, opened 2018-11-14
cleanup

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- rename `startAck` and `receive` methods of `InstructionExecutor`
- privatize the `next` method
- delete `validateMessage` (empty code)
- do not return from `receive`


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
The unit tests continue to test these private methods

joelalejandro:
Yes, but it is bad practice to access private methods, even in test contexts.

PR #220, opened 2018-11-14
[contracts] add ETHVirtualAppAgreement.sol

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- adds `ETHVirtualAppAgreement.sol` as the object to which state is committed in a virtual app protocol
- adds `ResolveToPay5ETHApp.sol` as a test fixture
- adds tests


packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
```diff
@@ -2,8 +2,8 @@ pragma solidity 0.4.25;
 pragma experimental "ABIEncoderV2";
 
 import "../lib/Transfer.sol";
-import "../NonceRegistry.sol";
 import "../Registry.sol";
+import "../AppInstance.sol";
 
 
 contract VirtualAppAgreement {
```
snario: Since this is specific to ETH, I recommend prefixing the contract with ETH

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
```diff
@@ -12,47 +12,49 @@ contract VirtualAppAgreement {
   using Transfer for Transfer.Terms;
 
   struct Agreement {
-    Registry registry;
-    NonceRegistry nonceRegistry;
+    address registryAddr;
```
snario: I think we might be able to store this as a constant value on the contract since it will be known at migration time

snario: I also think you can store `Registry` instead of `address` as the type

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
```diff
@@ -12,47 +12,49 @@ contract VirtualAppAgreement {
...
 
   function delegateTarget(
-    bytes32 salt,
-    uint256 nonce,
     Agreement agreement
   )
     public
   {
+    Registry registry = Registry(agreement.registryAddr);
+    require(agreement.expiry <= block.number, "agreement has not expired yet");
```
snario: I think the string `"agreement has not expired yet"` does not encapsulate the error state here. This is the case where someone tries to cash out their proxy agreement but the agreed amount of time for which the funds are locked up has not been reached yet; this is not really an expiry so much as a "release of borrowed / escrowed funds"

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
```diff
@@ -12,47 +12,49 @@ contract VirtualAppAgreement {
...
-    );
-
-    address[] memory to = new address[](2);
-    to[0] = agreement.registry.resolver(agreement.virtualStateDeposit);
-    to[1] = agreement.intermediary;
+    require(resolution.value.length == 1, "returned invalid resolution");
+    require(resolution.to.length == 1, "returned invalid resolution");
+    require(agreement.terms.assetType == resolution.assetType, "returned incompatible resolution");
+    require(agreement.terms.token == resolution.token, "returned incompatible resolution");
+    require(agreement.capitalProvided > resolution.value[0], "returned incompatible resolution");
```
snario: The logic for all of this is pretty critical. I'd like to recommend we document this in the header of the file, or perhaps on a docstring for the function. Ideally we reference to it on specs.counterfactual.com.

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
```diff
@@ -12,47 +12,49 @@ contract VirtualAppAgreement {
...
 
     uint256[] memory amount = new uint256[](2);
-    amount[0] = agreement.capitalProvided;
-    amount[1] = agreement.capitalProvided;
+    amount[0] = resolution.value[0];
+    amount[1] = agreement.capitalProvided - amount[0];
+
     bytes[] memory data = new bytes[](2);
```
snario: Generally speaking, it's annoying that we have to construct all these arguments every time. Is there a way we can create a `LibTransaction` that has convenience methods for constructing 2-person payouts of ETH for example to clean up this code?

packages/contracts/contracts/fixtures/FixedResolution.sol
```diff
@@ -0,0 +1,30 @@
+pragma solidity 0.4.25;
+pragma experimental "ABIEncoderV2";
+
+import "../lib/Transfer.sol";
+
+
+contract FixedResolutionApp {
```
snario: To be absurdly explicit here, I would rename this to `ResolveToPay5ETHApp` or something like that. Alternatively, you could just reuse `PaymentApp` for your test case and store `[ addrA, 5, addr B, 0 ]` as the app state.

packages/contracts/test/integration/virtualAppAgreement.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    virtualAppAgreement = await (await buildArtifacts.VirtualAppAgreement).deploy(
+      unlockedAccount
+    );
+    registry = await (await buildArtifacts.Registry).deploy(unlockedAccount);
+    delegateProxy = await (await AbstractContract.fromArtifactName(
+      "DelegateProxy"
+    )).deploy(unlockedAccount);
+  });
+
+  describe("test", () => {
```
snario: better naming

packages/contracts/test/integration/virtualAppAgreement.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    it("test", async () => {
```
snario: better naming

packages/contracts/test/integration/virtualAppAgreement.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+        tx,
+        Utils.HIGH_GAS_LIMIT
+      );
+
+      expect(
+        (await unlockedAccount.provider.getBalance(beneficiaries[0])).toString()
+      ).to.eq(ethers.utils.parseEther("5").toString());
+      expect(
+        (await unlockedAccount.provider.getBalance(beneficiaries[1])).toString()
+      ).to.eq(ethers.utils.parseEther("5").toString());
```
snario: This test case handles fixed values of 10 capitalProvided to a specific target app -- can you abstract it to run for other cases? Examples:

- target app being 0x0
- capital provided being 0 and resolution being nonzero
- capital provided being 0.99999 and resolution being 1 (or generally speaking, just not being `==`)
- expiry time not being reached yet
- terms limit not being met precisely
- other ones you can think of

packages/contracts/utils/buildArtifacts.ts
```diff
@@ -45,6 +45,15 @@ export const MinimumViableMultisig = AbstractContract.fromArtifactName(
   }
 );
 
+// tslint:disable-next-line
```
snario: y?

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: how is this done?

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Yes, I had a (really) old PR on the legacy contracts repo for this. I can reopen that PR separately from this one

packages/contracts/contracts/fixtures/FixedResolution.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I prefer not to use `PaymentApp` since there's some (unnecessary for a fixture IMO) indirection between the app instance and the app definition.

packages/contracts/utils/buildArtifacts.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: all of the exports have this - it's because of the non camelCased variable name

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I'm guessing at the bytecode level there would be `__registryAddr__` at 2-3 places in the code that gets replaced at link time

snario: I'm not 100% sure in the case of delegatecalls because I don't know how `constant`s work in Solidity precisely but it seems like you should be able to do this

packages/contracts/contracts/fixtures/FixedResolution.sol
snario: Okay, I still recommend the extremely explicit name

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this is passed in via calldata; are you saying to rely on some implicit conversion from `address` to `Registry`?

packages/contracts/contracts/fixtures/ResolveToPay5ETHApp.sol
```diff
@@ -4,7 +4,12 @@ pragma experimental "ABIEncoderV2";
 import "../lib/Transfer.sol";
 
 
-contract FixedResolutionApp {
+/*
+This contract is a test fixture meant to emulate an AppInstance contract. An
+AppInstance has a getResolution() function that returns a `Transfer.Transaction`
+object when the channel is closed.
+*/
```
snario: not to harp but can you use the standard docstring style? example

```solidity
/// @title MinimumViableMultisig - An example multisig exemplifying the minimal
/// viable multisignature wallet requirements for state channels support
/// @author Liam Horne - <liam@l4v.io>
/// @notice Supports the ability for each of:
/// (a) Executing artbitrary transactions using `CALL` or `DELEGATECALL`
/// (b) Requiring n-of-n unanimous consent
/// (c) Deterministic signature verification without an on-chain address
/// (d) Non-nonce based replay protection
contract MinimumViableMultisig { ... }
```

packages/contracts/migrations/2_deploy_libraries.js
```diff
@@ -4,7 +4,7 @@ const Conditional = artifacts.require("Conditional");
 const ConditionalTransaction = artifacts.require("ConditionalTransaction");
 const StaticCall = artifacts.require("StaticCall");
 const Transfer = artifacts.require("Transfer");
-const VirtualAppAgreement = artifacts.require("VirtualAppAgreement");
+const VirtualAppAgreement = artifacts.require("EthVirtualAppAgreement");
```
snario: Preference: `ETH` over `Eth`

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
snario: yes

packages/contracts/contracts/fixtures/FixedResolution.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: fixed

packages/contracts/test/integration/virtualAppAgreement.spec.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `limit` is not checked, btw - the `agreement.capitalProvided > resolution.value[0]` line does the same check

packages/contracts/contracts/delegateTargets/VirtualAppAgreement.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: well just doing `  address constant REGISTRY_ADDR = __registryAddr__` failed with a syntax error; will mark as TODO since I have no idea how to do this

packages/contracts/contracts/delegateTargets/ETHVirtualAppAgreement.sol
```diff
@@ -0,0 +1,69 @@
...
+    Transfer.Terms terms;
+    uint256 expiry;
+    bytes32 target;
+    uint256 capitalProvided;
+    address[2] beneficiaries;
+  }
+
+  function delegateTarget(
+    Agreement agreement
+  )
```
snario: why 3 lines

packages/contracts/test/integration/ethVirtualAppAgreement.spec.ts
```diff
@@ -0,0 +1,81 @@
+import * as ethers from "ethers";
```
snario: filename -- trying to keep files using lower-and-hyphen-case

packages/contracts/test/integration/ethVirtualAppAgreement.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    virtualAppAgreement = await (await buildArtifacts.VirtualAppAgreement).deploy(
+      unlockedAccount
+    );
+    registry = await (await buildArtifacts.Registry).deploy(unlockedAccount);
+    delegateProxy = await (await AbstractContract.fromArtifactName(
+      "DelegateProxy"
+    )).deploy(unlockedAccount);
+  });
+
+  describe("ETHVirtualAppAgreement", () => {
```
snario: seems like it needs more test cases

packages/contracts/test/integration/eth-virtual-app-agreement.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  capitalProvided: ethers.utils.BigNumber,
+  assetType: number
+): Promise<string[]> => {
+  const delegateProxy = await (await AbstractContract.fromArtifactName(
+    "DelegateProxy"
+  )).deploy(unlockedAccount);
+
+  await unlockedAccount.sendTransaction({
+    to: delegateProxy.address,
+    value: ethers.utils.parseEther("100")
```
snario: style: We're moving towards importing things like `parseEther` at module level `import { parseEther } from "ethers/utils";` and do on for the rest of this file

packages/contracts/test/integration/eth-virtual-app-agreement.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+        limit: 0,
+        token: ethers.constants.AddressZero
+      },
+      target: resolutionAddr
+    }
+  ]);
+
+  expect(
+    (await unlockedAccount.provider.getBalance(beneficiaries[0])).toString()
+  ).to.eq("0");
```
snario: Might be cleaner to do 

```typescript
  expect(await unlockedAccount.provider.getBalance(beneficiaries[0]))).to.eq(Zero);
```

packages/contracts/test/integration/eth-virtual-app-agreement.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  expect(
+    (await unlockedAccount.provider.getBalance(beneficiaries[1])).toString()
+  ).to.eq("0");
+
+  await delegateProxy.functions.delegate(
+    virtualAppAgreement.address,
+    tx,
+    Utils.HIGH_GAS_LIMIT
```
snario: You should check the output from the tests to get a rough estimate of the gas used in this transaction and make a constant `ESTIMATED_GAS_FOR_VIRTUAL_APP_DELEGATE = 400000` or whatever the number ends up being.

packages/contracts/test/integration/eth-virtual-app-agreement.spec.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Can we make that change elsewhere then?

packages/contracts/test/integration/eth-virtual-app-agreement.spec.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: doesn't work - the error is that `"0x0" != "0x00"`

PR #222, opened 2018-11-14
[contracts] Update migrate documentation step

cf19drofxots:
### Description

This updates the documentation to refer to the correct yarn script to locally migrate the contracts for testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
           name: "Run ganache"
           command: yarn ganache:ci
           background: true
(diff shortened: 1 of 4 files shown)


snario:
Couldn't we mimic the `truffle migrate` command here and just support its arguments? e.g., allow `yarn migrate --network rinkeby`

cf19drofxots:
We technically do, since the `yarn migrate` command is an alias for `truffle migrate --reset`, so you could point it to either `ganache` or `rinkeby`.

So I guess the better approach would be to document saying `yarn migrate --network ganache` and remove `yarn migrate:local`

PR #223, opened 2018-11-14
Move current CFjs code to legacy/ folder

mvanderh:
### Description

Make room for the new CFjs library according to https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md. Will move code over from legacy as we build the new CFjs.


packages/cf.js/src/legacy/network.ts
```diff
@@ -1,7 +1,7 @@
 import ethers from "ethers";
 import * as _ from "lodash";
 
-import { Address } from "./utils";
+import { Address } from "./utils/index";
```
snario: Isn't `index` auto imported as default if you use `"./utils"`

snario:
There are things in here that you're putting into `legacy` that seem to be critical pieces we will want to keep in the core package for now. Ideally `legacy` is only needed for the pieces that keep Ethmo working and the rest can be integrated into 0.0.2 spec

mvanderh:
I think you misunderstand the purpose of this PR. This PR keeps all current functionality intact, it just moves all the old code to a legacy sub-package. This makes it much easier to build the new CFjs and slowly migrate old code over if we need it.

I dug into the current CFjs and how it's used in our other packages and concluded this was the easiest way to build CFjs 0.0.2 without breaking everything. 

@patience-tema-baron is still working on fixing & freezing Ethmo by publishing NPM packages, in parallel.

packages/cf.js/src/legacy/network.ts
mvanderh: Yeah, blame my IDE. Fix pending

cf19drofxots:
Reminder to follow the template for PR titles as well

PR #230, opened 2018-11-15
[playground] Adds support for Netlify

joelalejandro:
### Description

This PR injects the necessary configuration for Netlify to deploy a static version of the Playground.


snario:
I think you need to specify `packages/playground/dist`

packages/playground/.nvmrc
```diff
@@ -0,0 +1 @@
+10.13.0
```
snario: I'm on 10.5.0. Why did you choose 10.13.0?

joelalejandro: @snario 10.13.0 is the standard LTS version. We should be working on top of that.

packages/playground/netlify.toml
```diff
@@ -1,4 +1,4 @@
 [build]
-  base    = "packages/playground"
+  base    = "."
   publish = "dist"
   command = "npm run build"
```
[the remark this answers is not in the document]
joelalejandro: @snario Does Netlify has built-in "yarn" support? If so, I'll change it. I didn't see anywhere that it's preinstalled on the container or something.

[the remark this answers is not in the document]
joelalejandro: @snario Actually, scratch that - I just saw in the logs it does.

joelalejandro:
@snario About the "base" value, since it's now on `packages/playground` it shouldn't need the entire path.

packages/playground/.nvmrc
snario: Okay

PR #233, opened 2018-11-15
[CFjs] Scaffolding + Client: getAppInstances()

mvanderh:
### Description

Scaffolding of CFjs 0.0.2 package layout. 
Test + implementation for Client and getAppInstances()


packages/cf.js/test/client.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    expect(queryMessage.data.queryType).toBe("allAppInstances"); // TODO
+
+    nodeProvider.simulateMessageFromNode({
+      requestId: queryMessage.requestId,
+      messageType: NodeMessageType.QUERY,
+      data: {
+        appInstances: [testInstance]
+      }
+    });
+  });
```
patience-tema-baron: Should we add a test for `createAppFactory`?

mvanderh: In the future when we build AppFactory, but for now there's really nothing to test.

packages/cf.js/src/client.ts
```diff
@@ -0,0 +1,87 @@
...
+  NodeMessageType,
+  NodeProvider,
+  QueryType
+} from "./structs";
+
+export enum ClientEventType {
+  INSTALL = "install",
+  PROPOSE_INSTALL = "proposeInstall",
+  REJECT_INSTALL = "rejectInstall"
+}
```
snario: Are these the message types the NodeProvider expects? Can we prefix them with `cf_`?

packages/cf.js/src/client.ts
```diff
@@ -0,0 +1,87 @@
...
+import {
+  AppDefinition,
+  MessageDataQuery,
+  NodeMessage,
+  NodeMessageType,
+  NodeProvider,
+  QueryType
+} from "./structs";
+
+export enum ClientEventType {
```
snario: These are moreso "Request"s not "ClientEvent"s?

packages/cf.js/src/client.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export interface ClientEvent {
+  readonly eventType: ClientEventType;
+  readonly data: any; // TODO
+}
+
+const NODE_REQUEST_TIMEOUT = 1500;
+
+export class Client {
```
snario: I think this needs a better name... this is the equivalent of an `ethers` Provider right?

packages/cf.js/src/types.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./simple-types";
```
snario: I don't think we should have a `types.ts` file. Is this meant to mimic `ethers.types`?

packages/cf.js/test/client.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  NodeMessageType,
+  NodeProvider,
+  QueryType
+} from "../src/structs";
+
+class TestNodeProvider implements NodeProvider {
+  public postedMessages: NodeMessage[] = [];
+  readonly callbacks: ((message: NodeMessage) => void)[] = [];
+
+  public sendMessageToClient(message: NodeMessage) {
```
snario: Can it just be `send` instead of `sendMessageToClient`? Also it's not sending it to a client its sending it to a Node right?

packages/cf.js/src/client.ts
joelalejandro: @snario Curious to know if we're prefixing them for any reason. Is name collision possible in this context?

packages/cf.js/src/structs.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export interface MessageDataQuery {
+  queryType: QueryType;
+  appInstances?: any[];
+}
+
+export interface MessageDataError {
+  message: string;
+  extra?: any;
```
joelalejandro: @mvanderh Should this be at least `object`?

packages/cf.js/src/structs.ts
```diff
@@ -0,0 +1,38 @@
...
+  ERROR = "error"
+}
+
+export enum QueryType {
+  GET_APP_INSTANCES = "getAppInstances"
+}
+
+export interface MessageDataQuery {
+  queryType: QueryType;
+  appInstances?: any[];
```
joelalejandro: @mvanderh Why do we use `any` if we have `AppInstance` defined?

packages/cf.js/src/structs.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export interface MessageDataError {
+  message: string;
+  extra?: any;
+}
+
+export interface NodeMessage {
+  requestId: string;
+  messageType: NodeMessageType;
+  data: MessageDataQuery | MessageDataError | null;
```
joelalejandro: @mvanderh Instead of accepting `null` as a type, wouldn't it better to tag the property as `data?`?

packages/cf.js/src/client.ts
snario: Mirroring web3.

packages/cf.js/src/structs.ts
mvanderh: I prefer forcing the node provider to explicitly specify null if there's no data.

packages/cf.js/src/structs.ts
mvanderh: Maybe. `object` is kind of annoying though because you have to cast it before you can access itt like a JS object.

packages/cf.js/src/client.ts
mvanderh: Yep, it might indeed be a good idea to rename it to Provider

packages/cf.js/src/types.ts
mvanderh: Why not? Where do we put `Address`, `H256`, `ABIEncoding` typedefs?

packages/cf.js/src/client.ts
mvanderh: No. There are events that aren't request-response e.g. your counterparty proposes to install an app.

packages/cf.js/src/client.ts
mvanderh: OTOH, we already have something in our Counterfactual namespace with "Provider" i.e. NodeProvider. Might be a good idea to stick with Client and prevent confusion

packages/cf.js/src/structs.ts
joelalejandro: @mvanderh Is there anything to structure it somehow? Maybe an interface or a type? I tend to think that using `any` on a strongly-typed environment is kinda weird.

packages/cf.js/src/structs.ts
joelalejandro: You could do something like `{ [index: string]: any }`, maybe at least to suggest it has keys?

mvanderh: Sure!

packages/cf.js/src/structs.ts
mvanderh: Good idea.

packages/cf.js/test/client.spec.ts
```diff
@@ -42,7 +42,7 @@ describe("CF.js Client", async () => {
     const queryMessage = nodeProvider.postedMessages[0];
 
     expect(queryMessage.messageType).toBe(NodeMessageType.QUERY);
-    const queryData = queryMessage.data! as MessageDataQuery;
+    const queryData = queryMessage.data! as NodeQueryData;
```
joelalejandro: Is the `!` necessary, despite casting `null` as NodeQueryData?

mvanderh:
@joelalejandro A re-review would be helful since I moved a bunch of interfaces into node-provider, a package you own.

PR #234, opened 2018-11-15
missing features for cfjs@0.0.2

patience-tema-baron:
For discussion purposes only.

### DIFF
--- a/packages/cf.js/considerations-for-0-0-2.md
+++ b/packages/cf.js/considerations-for-0-0-2.md
@@ -1,0 +1,13 @@
+# missing features of v0.0.2
+
+- need a way to provide a `timeout` during the install flow
+    - argument to `proposeInstall`?
+    - attribute of the manifest.json
+- currently, the user's address is never being set
+    - provide as an argument: `new Client(nodeProvider, address)`
+    - fetch it from the wallet through an `async init` method?
+- do dapp's specify the deposit? if so, `proposeInstall` and `install` need a deposit arg
+- should the `terms` be set during `install` instead of `proposeInstall`, since we won't know the full limit of the terms until both parties have deposited?
+- should the dapp be able to provide an initial state during `install`? or do we expect dapps to immediately perform an `updateState`?
+- in the legacy api, we provide uninstall with `{ peerABalance, peerBBalance }` to determine how much eth to deposit to their respective accounts. should we support that with `0.0.2`? or is this handled in another way now?
+- how do we support `proposeInstall` before it's implemented node-side? there's a communication issue here, where we need to get the proposal to the counterparty. how do we handle peer-to-peer communications outside of node notifications?


patience-tema-baron:
@snario @mvanderh

joelalejandro:
@patience-tema-baron One thing we discussed with @ebryn was the need for `AppManifest` to contain include a URL for opening the dApp in the Playground. Otherwise, after `install` is completed, we wouldn't know how to open the dApp.

packages/cf.js/considerations-for-0-0-2.md
```diff
@@ -0,0 +1,13 @@
+# missing features of v0.0.2
+
+- need a way to provide a `timeout` during the install flow
+    - argument to `proposeInstall`?
+    - attribute of the manifest.json
+- currently, the user's address is never being set
```
mvanderh: Why do we need this? Per-app user addresses should be generated deterministically by the node.

packages/cf.js/considerations-for-0-0-2.md
```diff
@@ -0,0 +1,13 @@
...
+    - provide as an argument: `new Client(nodeProvider, address)`
+    - fetch it from the wallet through an `async init` method?
+- do dapp's specify the deposit? if so, `proposeInstall` and `install` need a deposit arg
```
mvanderh: I'm still not clear on this. Doesn't the proposeInstall step need to specify _both_ deposits? Isn't the app expected to handle terms negotiation?

packages/cf.js/considerations-for-0-0-2.md
```diff
@@ -0,0 +1,13 @@
...
+- should the `terms` be set during `install` instead of `proposeInstall`, since we won't know the full limit of the terms until both parties have deposited?
+- should the dapp be able to provide an initial state during `install`? or do we expect dapps to immediately perform an `updateState`?
```
mvanderh: Definitely need to pass an initial state to install().  Good catch

packages/cf.js/considerations-for-0-0-2.md
```diff
@@ -0,0 +1,13 @@
...
+- in the legacy api, we provide uninstall with `{ peerABalance, peerBBalance }` to determine how much eth to deposit to their respective accounts. should we support that with `0.0.2`? or is this handled in another way now?
```
mvanderh: Wait, we did? That seems insecure. Isn't the `resolve` function of the app supposed to be the only authority on payouts?

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: Currently, we provide the `fromAddress` in all action messages to the node, and I suspect we'll still want to, right? Or will the next generation of the node not need this information because it's able to deterministically generate that address?

mvanderh: I would definitely think the latter. The node should be able to generate or recall `fromAddress` based on just the appId.

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: My understanding of `proposeInstall` is that it is the initial contact between parties, and so the initiator wouldn't yet have the counterparty's deposit amount. This would be an optimal time for them to provide their own deposit amount though, and then the counterparty would be able to provide both their own deposit amount and the initiator's deposit to `install`.

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: Yeah, now that you point it out, that does seem totally insecure.

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: That would be cool. It might be worth hashing out with @cf19drofxots what our node requests should look like.

packages/cf.js/considerations-for-0-0-2.md
mvanderh: Okay. The initiator would still get to Accept or Decline the `install` in their node, right?

patience-tema-baron: I believe so. Can you confirm, @snario?

packages/cf.js/considerations-for-0-0-2.md
cf19drofxots: What is the context here? Where is the user's address not being set?

And yes, in the next version of the Node, addresses are deterministically generated per-app via ephemeral signing keys.

packages/cf.js/considerations-for-0-0-2.md
```diff
@@ -0,0 +1,13 @@
+# missing features of v0.0.2
+
+- need a way to provide a `timeout` during the install flow
+    - argument to `proposeInstall`?
+    - attribute of the manifest.json
+- currently, the user's address is never being set
+    - provide as an argument: `new Client(nodeProvider, address)`
```
cf19drofxots: I'm failing to see how a user's address is relevant to the initialization of a _Client_.

What's the intent of providing an address here?

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: The user's address is never being set in the client interface, though it seems like that will no longer be necessary. We were using it previously to provide the `fromAddress` in our action requests, but it sounds like the node will won't need that information any longer.

packages/cf.js/considerations-for-0-0-2.md
cf19drofxots: Yes.

That's part of the validation logic.

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: We were using it previously to provide the fromAddress in our action requests, but it sounds like the node will won't need that information any longer.

packages/cf.js/considerations-for-0-0-2.md
```diff
@@ -0,0 +1,13 @@
...
+    - fetch it from the wallet through an `async init` method?
+- do dapp's specify the deposit? if so, `proposeInstall` and `install` need a deposit arg
+- should the `terms` be set during `install` instead of `proposeInstall`, since we won't know the full limit of the terms until both parties have deposited?
```
cf19drofxots: This would then cause the `terms` to possibly be out of sync between the parties.

In further thinking about the overall flow here, I'm leaning towards more simplicity (at least for the initial version) and making the negotiation the responsibility of the app to handle. 

Simple scenario:
A `installProposal(terms, deposit, ...)` -->  B `ackInstallProposal(terms, deposit, ...)` --> A `install(terms, deposit, ...)`

B doesn't agree:
A `installProposal(terms, deposit, ...)` -->  B `rejectInstallProposal(terms, deposit, ...)`

Upon receiving the rejection, A can either send a new proposal or B can become the initiator.

This makes it such that the `terms` can be mutually agreed on and _then_ having the install proceed, providing better security guarantees.

packages/cf.js/considerations-for-0-0-2.md
cf19drofxots: What's the rationale behind providing an initial state?

Is this motivated by recovering from a crash? Otherwise apps should be instantiated with a default state, no?

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: Would this necessitate that A specify the deposit amount for B? As I understand it, `terms.limit` is the sum of their two deposits.

cf19drofxots: That's correct.

In which case, B can effectively send a counter proposal with a deposit amount B is happy with.

patience-tema-baron: That makes sense. I wonder if the api should look like this then:

```
proposeInstall(peerAddress, deposit, peerDeposit, terms): Promise<appId>
install(appId)
```

packages/cf.js/considerations-for-0-0-2.md
cf19drofxots: Yeah I can see why that'd be required for Ethmo. Ethmo's default state is effectively the input to the app, and I can see other apps having the same dynamic.

Though I think some others apps wouldn't need it.

For example, if I'm instantiating a game of Chess, it'd be annoying for me to construct an initial state for it.

packages/cf.js/considerations-for-0-0-2.md
cf19drofxots: Yeah, essentially something like that.

cf19drofxots: Just to confirm, the promise resolves to the app ID and rejects with a `rejectInstall` event type?

patience-tema-baron: I suspect that the promise shouldn't resolve with an event type, but that we should instead rely the event observers to handle that. (But, I may be biased. I still don't think we should be return a promise here at all.) What are your thoughts, @mvanderh? Both about returning a `rejectInstall` event and adjusting the api as I described above?

packages/cf.js/considerations-for-0-0-2.md
patience-tema-baron: I wonder if it might be worth having a `default_initial_state` attribute in the manifest? So that dapps could simply fetch/use that as a fallback. Regardless, in the case of the chess game, you'd have to define it somewhere.

mvanderh: I think you always need it @cf19drofxots, even for something like Chess. You need to _at least_ provide the payout addresses in the initial state, otherwise money you deposit into an app is lost forever.

cf19drofxots:
Is this PR still active? I see https://github.com/counterfactual/monorepo/pull/237 as what seems to be the takeaway

mvanderh:
I think this PR is done. Similar questions might come up, but we'll address them then.

PR #235, opened 2018-11-15
[monorepo] Cleanup cache directories

cf19drofxots:
### Description

This generalizes the circle CI directories being used for caching.


.dependencies_checksum
```diff
@@ -0,0 +1 @@
+e9230c2f0c6db0ae559ee7be5d4045cd6c2af6557e3259d7c8bb67649a9be760  -
```
snario: Is there meant to be a `  -` at the end?

package.json
```diff
@@ -31,12 +31,19 @@
...
     "tslint": "^5.11.0",
     "tslint-config-airbnb": "^5.11.0",
     "tslint-config-prettier": "^1.15.0",
     "tslint-plugin-prettier": "^2.0.0",
     "typescript": "^3.1.3"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "cat packages/*/package.json | ./node_modules/.bin/sha256sum > .dependencies_checksum"
+    }
```
snario: Couldn't we just take this command and run it inside `circleci` before the cache is loaded and then we wouldn't need to add this to the repo?

cf19drofxots: We could

.dependencies_checksum
cf19drofxots: Yeah, that's the entire output of `sha256sum` - doesn't seem like a big deal

cf19drofxots: Oh the `-` is actually the name of the file being checksummed.

Since the input is from the shell (redirected from `cat`), it displays `-`.

cf19drofxots:
@snario any further comments?

PR #242, opened 2018-11-18
[dapp-high-roller] Create initial boilerplate High Roller dApp

Alonski:
### Description

Initialized Stencil app.
One thing that I found weird is that I had to have a component other than the `app-home` one for the contents of `app-home` to render.
Couldn't figure out why.

I noticed this issue with the `Playground` package as well.


joelalejandro:
@Alonski The `app-root` component is the top-level element in the Stencil app hierarchy, AFAIK. I believe it has to do with the fact that this component includes `stencil-router` and the primary layout for the app.

Alonski:
@joelalejandro Yea I understand. Right now if you run the app it will render the content inside of `app-home`. However, if you remove the `app-profile` folder and then rebuild the app all of a sudden the contents of `app-home` won't render. 

Maybe it was only happening to me?

Alonski:
@joelalejandro Actually while running the Playground I also didn't see the home page content.

Alonski:
@joelalejandro Found an issue for what I was describing:
https://github.com/ionic-team/stencil-router/issues/62

PR #247, opened 2018-11-20
machine test harness cleanup

### LINKED ISSUE #96: [machine] Code in TestIoProvider.findMessage() isn't type-safe
https://github.com/counterfactual/monorepo/blob/95462c8172cfb6c7a8bc06b0b78942319d1cf6fb/packages/machine/test/integration/test-io-provider.ts#L71-L80

### LINKED ISSUE #105: [machine] TestResponseSink.sendIoMessageToClient() can be confused with TestResponseSink.sendMessageToClient()
https://github.com/counterfactual/monorepo/blob/95462c8172cfb6c7a8bc06b0b78942319d1cf6fb/packages/machine/test/integration/test-response-sink.ts#L143-L149

### LINKED ISSUE #106: [machine] TestResponseSink.sendIoMessageToClient() should which client gets the message
https://github.com/counterfactual/monorepo/blob/95462c8172cfb6c7a8bc06b0b78942319d1cf6fb/packages/machine/test/integration/test-response-sink.ts#L143-L149

### LINKED ISSUE #147: [machine] TestResponseSink.messageListener should support multiple callbacks
https://github.com/counterfactual/monorepo/blob/95462c8172cfb6c7a8bc06b0b78942319d1cf6fb/packages/machine/test/integration/test-response-sink.ts#L151-L154

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- replace `TestIOProvider::listeners` with a single listener now called `waitForIoContinuation`. name explanation: this is the continuation captured by `await TestIOProvider::waitForIo`
- do not do any checks on incoming messages in `TestIOProvider`
- delete unused functions in `TestResponseSink`
- replace `TestResponseSink::requests` with a single request now called `runProtocolContinuation`. name explanation: this is the continuation captured by `await TestResponseSink::runProtocol`
- set the return type of `runProtocol` to be `Response` instead of `WalletResponse`
- fixes #96 
- fixes #105 
- fixes #106 
- fixes #147 -


packages/machine/test/integration/test-io-provider.ts
```diff
@@ -34,70 +31,30 @@ export class TestIOProvider {
...
-      }
-    });
-
-    // Now remove all listeners we just invoked
-    executedListeners.forEach(index => this.listeners.splice(index, 1));
-
-    // initiate ack side if needed
-    if (message.seq === 1) {
+    if (this.waitForIoContinuation !== undefined) {
+      const k = this.waitForIoContinuation;
```
cf19drofxots: Can we give this variable a better name?

packages/machine/test/integration/test-io-provider.ts
```diff
@@ -34,70 +31,30 @@ export class TestIOProvider {
...
+      this.waitForIoContinuation = undefined;
```
cf19drofxots: It's not immediately clear what the intent here is.

Why are we reassigning `undefined` to this?

packages/machine/test/integration/test-io-provider.ts
```diff
@@ -34,70 +31,30 @@ export class TestIOProvider {
...
+      k(message);
+    } else if (message.seq === 1) {
```
cf19drofxots: Can we document why this equality is being done?

It's kind of opaque without context.

packages/machine/test/integration/test-io-provider.ts
```diff
@@ -34,70 +31,30 @@ export class TestIOProvider {
...
+    const message = this.messages.pop();
+
+    if (message !== undefined) {
+      continuation(message);
     } else {
-      this.messages.splice(this.messages.indexOf(message), 1);
-      method(message);
+      this.waitForIoContinuation = continuation;
     }
+
```
cf19drofxots: Lint

packages/machine/test/integration/test-response-sink.ts
```diff
@@ -24,15 +23,13 @@ export class TestResponseSink implements cf.legacy.node.ResponseSink {
...
-  private messageListener?: Function;
+  private runProtocolContinuation?: Function;
 
   constructor(
     readonly privateKey: string,
     networkContext?: cf.legacy.network.NetworkContext
   ) {
-    // A mapping of requsts that are coming into the response sink.
-    this.requests = new Map<string, Function>();
+    this.runProtocolContinuation = undefined;
```
cf19drofxots: This isn't needed. Its default value is `undefined` upon construction.

packages/machine/test/integration/test-response-sink.ts
```diff
@@ -24,15 +23,13 @@ export class TestResponseSink implements cf.legacy.node.ResponseSink {
   public store: TestCommitmentStore;
   public signingKey: ethers.utils.SigningKey;
 
-  private requests: Map<string, Function>;
-  private messageListener?: Function;
+  private runProtocolContinuation?: Function;
```
cf19drofxots: Can the typing here be tightened at all for input/output accounting instead of a generic `Function`?

packages/machine/test/integration/test-response-sink.ts
```diff
@@ -123,14 +120,13 @@ export class TestResponseSink implements cf.legacy.node.ResponseSink {
    * Resolves the registered promise so the test can continue.
    */
   public sendResponse(res: cf.legacy.node.WalletResponse) {
-    if ("requestId" in res && this.requests[res.requestId] !== undefined) {
-      const promise = this.requests[res.requestId];
-      delete this.requests[res.requestId];
-      promise(res);
+    if (this.runProtocolContinuation) {
+      const k = this.runProtocolContinuation;
```
cf19drofxots: Naming here as well.

packages/machine/test/integration/test-response-sink.ts
cf19drofxots: Looks like it only ever accepts `cf.legacy.node.WalletResponse`

packages/machine/test/integration/test-io-provider.ts
```diff
@@ -117,42 +74,20 @@ export class TestIOProvider {
...
-
     this.listenOnce(
       msg => {
         context.intermediateResults.inbox = msg;
         resolve(msg);
       },
-      multisig,
-      appId
     );
+
```
cf19drofxots: Lint

packages/machine/test/integration/test-io-provider.ts
```diff
@@ -6,25 +6,22 @@ import { InternalMessage } from "../../src/types";
...
-    method: Function;
-  }[];
+  private waitForIoContinuation?: ClientActionMessageConsumer;
 
   public ackMethod: Function = Object.create(null);
 
   constructor() {
     this.messages = [];
-    this.listeners = [];
+    this.waitForIoContinuation = undefined;
```
cf19drofxots: Not needed. Default value is `undefined`.

packages/machine/test/integration/test-io-provider.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The name is explained in the PR. When not undefined this variable is the continuation (https://en.m.wikipedia.org/wiki/Continuation) of the call to waitForIo

cf19drofxots: Right, but it's probably better if it's explained here as the code is the source of the truth.

Cross-referencing with the PR that merged the code adds unrealistic overhead to reach the explanation.

packages/machine/test/integration/test-io-provider.ts
snario: I think the idea is that you had a continuation ready to be used on some data when the time was right, and this code snippet is only executed "when the time is right" and so it executes it and gets rid of the continuation

snario: tbh I think it's a cool use of continuations

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: @cf19drofxots is right, the reassignment is not actually needed

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: ah actually it is needed, because of the check here https://github.com/counterfactual/monorepo/pull/247/files#diff-dce421262ecb0310b0ecc493a5bd9fbeR33

packages/machine/test/integration/test-io-provider.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: added a comment

PR #252, opened 2018-11-20
[WIP] [CFjs] AppFactory.proposeInstall

mvanderh:
### Description

AppFactory is now able to propose the installation of an app.

### DIFF
--- a/packages/cf.js/src/app-factory.ts
+++ b/packages/cf.js/src/app-factory.ts
@@ -1,6 +1,6 @@
 import {
-  NodeInstallProposalData,
-  NodeMessageType
+  NodeMessageType,
+  NodeProposeInstallData
 } from "@counterfactual/node-provider";
 import { BigNumber, BigNumberish } from "ethers/utils";
 
@@ -37,6 +37,6 @@ export class AppFactory {
         appDefinition: this.appDefinition
       }
     );
-    return (response.data as NodeInstallProposalData).appInstanceId!;
+    return (response.data as NodeProposeInstallData).appInstanceId!;
   }
 }
--- a/packages/cf.js/src/app-factory.ts
+++ b/packages/cf.js/src/app-factory.ts
@@ -1,5 +1,42 @@
-import { AppDefinition } from "./structs";
+import {
+  NodeInstallProposalData,
+  NodeMessageType
+} from "@counterfactual/node-provider";
+import { BigNumber, BigNumberish } from "ethers/utils";
+
+import { AppDefinition, BlockchainAsset } from "./app-types";
+import { Provider } from "./provider";
+import { Address, AppInstanceID } from "./simple-types";
+
+export interface InstallProposalRequest {
+  peerAddress: Address;
+  asset: BlockchainAsset;
+  myDeposit: BigNumberish;
+  peerDeposit: BigNumberish;
+  initialState: any;
+}
 
 export class AppFactory {
-  constructor(readonly appDefinition: AppDefinition) {}
+  constructor(
+    readonly provider: Provider,
+    readonly appDefinition: AppDefinition
+  ) {}
+
+  async proposeInstall(
+    proposal: InstallProposalRequest
+  ): Promise<AppInstanceID> {
+    const response = await this.provider.sendRawNodeRequest(
+      NodeMessageType.PROPOSE_INSTALL,
+      {
+        assetType: proposal.asset.assetType,
+        token: proposal.asset.token,
+        peerAddress: proposal.peerAddress,
+        myDeposit: new BigNumber(proposal.myDeposit).toString(),
+        peerDeposit: new BigNumber(proposal.peerDeposit).toString(),
+        initialState: proposal.initialState,
+        appDefinition: this.appDefinition
+      }
+    );
+    return (response.data as NodeInstallProposalData).appInstanceId!;
+  }
 }
--- a/packages/cf.js/src/app-types.ts
+++ b/packages/cf.js/src/app-types.ts
@@ -1,0 +1,28 @@
+import { BigNumberish } from "ethers/utils";
+
+import { ABIEncoding, Address } from "./simple-types";
+
+export enum BlockchainAssetType {
+  ETH = 0,
+  ERC20 = 1,
+  OTHER = 2
+}
+
+export interface BlockchainAsset {
+  assetType: BlockchainAssetType;
+  token?: Address;
+}
+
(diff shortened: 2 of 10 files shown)


packages/cf.js/src/app-factory.ts
```diff
@@ -1,5 +1,42 @@
...
+  NodeMessageType,
+  NodeProposeInstallData
+} from "@counterfactual/node-provider";
+import { BigNumber, BigNumberish } from "ethers/utils";
+
+import { AppDefinition, BlockchainAsset } from "./app-types";
+import { Provider } from "./provider";
+import { Address, AppInstanceID } from "./simple-types";
+
+export interface InstallProposalRequest {
```
snario: Shouldn't this include the app you're installing?

packages/cf.js/src/app-factory.ts
```diff
@@ -1,5 +1,42 @@
...
+  async proposeInstall(
+    proposal: InstallProposalRequest
+  ): Promise<AppInstanceID> {
+    const response = await this.provider.sendRawNodeRequest(
+      NodeMessageType.PROPOSE_INSTALL,
+      {
+        assetType: proposal.asset.assetType,
+        token: proposal.asset.token,
+        peerAddress: proposal.peerAddress,
+        myDeposit: new BigNumber(proposal.myDeposit).toString(),
```
snario: Why are these types strings?

packages/cf.js/src/app-types.ts
```diff
@@ -0,0 +1,28 @@
...
+  token?: Address;
+}
+
+export interface AppDefinition {
+  address: Address;
+  appStateEncoding: ABIEncoding;
+  appActionEncoding: ABIEncoding;
+}
+
+export interface AppInstallProposal {
```
snario: Why is this different from `InstallProposalRequest`

packages/cf.js/src/provider.ts
```diff
@@ -10,17 +11,25 @@ import * as uuid from "uuid";
...
 export enum CounterfactualEventType {
   INSTALL = "cf_install",
   PROPOSE_INSTALL = "cf_proposeInstall",
   REJECT_INSTALL = "cf_rejectInstall"
 }
 
+export interface InstallProposalData {
+  proposal: AppInstallProposal;
+
+  reject();
```
snario: ?

packages/cf.js/src/app-factory.ts
```diff
@@ -1,5 +1,42 @@
...
 
 export class AppFactory {
-  constructor(readonly appDefinition: AppDefinition) {}
+  constructor(
+    readonly provider: Provider,
+    readonly appDefinition: AppDefinition
+  ) {}
+
+  async proposeInstall(
+    proposal: InstallProposalRequest
```
snario: Is this pattern of making an interface for function arguments really necessary?

packages/cf.js/src/simple-types.ts
```diff
@@ -1,3 +1,4 @@
 export type ABIEncoding = string;
 export type AppInstanceID = string;
 export type Address = string;
+export type H256 = string;
```
snario: Why do we use "H256"? Seems super ambiguous compared to "bytes32" or something

packages/cf.js/test/app-factory.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+describe("App Factory", async () => {
+  const TEST_APP_DEFINITION = {
+    address: ethers.constants.AddressZero,
+    appActionEncoding: "tuple(uint256 increase)",
+    appStateEncoding: "tuple(uint256 val)"
+  };
+  let nodeProvider: TestNodeProvider;
+  let provider: Provider;
+
+  beforeEach(() => {
```
snario: does this need to be a `beforeEach` or can it just be `beforeAll` / `before`

packages/node-provider/src/types.ts
```diff
@@ -4,6 +4,7 @@ export interface INodeProvider {
 }
 
 export enum NodeMessageType {
```
snario: Is there a specification for this?

packages/cf.js/src/app-factory.ts
mvanderh: Oh yeah totally, this is still a rough WIP

packages/cf.js/src/app-factory.ts
mvanderh: Not normally, but in this case i think it's better to use an interface as opposed to 5+ positional arguments.

packages/node-provider/src/types.ts
mvanderh: I don't think so. @joelalejandro

packages/cf.js/test/app-factory.spec.ts
mvanderh: beforeEach is better, makes individual tests idempotent.

packages/cf.js/src/simple-types.ts
mvanderh: Amen.

packages/cf.js/src/app-factory.ts
```diff
@@ -1,5 +1,42 @@
...
+  ): Promise<AppInstanceID> {
+    const response = await this.provider.sendRawNodeRequest(
+      NodeMessageType.PROPOSE_INSTALL,
+      {
+        assetType: proposal.asset.assetType,
+        token: proposal.asset.token,
+        peerAddress: proposal.peerAddress,
+        myDeposit: new BigNumber(proposal.myDeposit).toString(),
+        peerDeposit: new BigNumber(proposal.peerDeposit).toString(),
```
patience-tema-baron: In `AppInstallProposal`, we define the deposits as `deposits: { [addr: string]: BigNumberish };`. Why not here, too?

packages/cf.js/src/provider.ts
mvanderh: Yeah, we were thinking of making rejectProposal a callback on the event. Probably not a good idea. Likely this is going to be changed anyway after sync with the Playground folks

packages/cf.js/src/app-factory.ts
patience-tema-baron: Maybe not? Since the AppFactory will already have the app definition? https://github.com/counterfactual/monorepo/pull/252/files#diff-0ecb04453a0af52b368b562894258f61R37

PR #253, opened 2018-11-21
[machine] metachannel test harness and messaging

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
# Features

- partial metachannel support in machine: added a metachannel install application protocol stub (https://github.com/counterfactual/monorepo/pull/253/files#diff-7cf9fe7ef1079b1e6d94fd6a768d1ed0R273) that does more complex message passing and message waiting among three agents
- add test harness for the above (https://github.com/counterfactual/monorepo/pull/253/files#diff-0b69721704c0e927bf7ceeccefae87dcR32)
- refactor machine API
  + `InstructionExecutor::dispatchReceivedMessage`
  + `InstructionExecutor::run{Update,Uninstall,Install,Setup,InstallMetachannelApp}Protocol`

# Supporting features

- machine
  + make `instructions.ts` indexed by protocol name and message sequence number (overload `0` to mean "for initiating")
  + support inline functions in instruction groups
- machine integration test harness:
  + `TestResponseSink` statefully keep track of whether it is `active` (i.e. whether a protocol run that is not the result of a received message is happening)
  + `TestResponseSink` uses `active` flag to determine whether to dispatch received messages to the machine, instead of testing `message.seq === 1` 
  + remove deserialization/hydration of messages
  + make `inbox` and `outbox` arrays
- fix bug where `to/from` in many messages were incorrectly set
- inline the following opcodes:
  + `OP_GENERATE`
  + `IO_PREPARE_SEND`
  + `KEY_GENERATE`
- delete the `ALL` pseudo-opcode
- disallow middlewares from calling `next`

# Cleanup

- delete legacy classes:
  + `cf.legacy.AppInstanceClient`
  + `cf.legacy.Channel`
  + `cf.legacy.Client`
- delete write-ahead log
- delete unused field `appId` of `cf.legacy.AppInstanceInfo`
- delete fields `requestId` and `error` of `cf.legacy.node.Response`
- delete interfaces `WalletMessaging`
- merge `ClientActionMessage` with `ClientMessage`
- delete interfaces `ClientQuery`, `ClientResponse`, `UserDataClientResponse`, `StateChannelDataClientResponse`, `FreeBalanceClientResponse`, `InstallClientResponse`, `WalletMessage`, `WalletResponse`
- do not apply `Observable` mixin on `InstructionExecutor`


packages/machine/test/integration/cf-operations.spec.ts
```diff
@@ -459,14 +444,16 @@ async function uninstallBalanceRefund(
...
-  const msg = startUninstallBalanceRefundMsg(
-    multisigAddr,
-    cfAddr,
+  const response = await walletA.runUninstallProtocol(
     walletA.signingKey.address,
     walletB.signingKey.address,
-    amountA
+    multisigAddr,
+     [
+      new cf.legacy.utils.PeerBalance(walletA.signingKey.address, amountA),
```
snario: I think we should remove the use of this `PeerBalance` class if we can; it doesn't really do anything useful

packages/machine/test/integration/lifecycle/depositor.ts
```diff
@@ -65,10 +65,16 @@ export class Depositor {
...
+      peerA.signingKey.address!,
+      peerB.signingKey.address!,
+      UNUSED_FUNDED_ACCOUNT,
+      msg
+    )
     expect(response.status).toEqual(cf.legacy.node.ResponseStatus.COMPLETED);
     // since the machine is async, we need to wait for peerB to finish up its
     // side of the protocol before inspecting it's state
+    // todo(ldct): wtf?
     await new Promise(resolve => setTimeout(resolve, 50));
```
snario: Marking this as something to fix

packages/machine/test/integration/test-response-sink.ts
```diff
@@ -104,25 +91,86 @@ export class TestResponseSink implements cf.legacy.node.ResponseSink {
...
+    peerAmounts: cf.legacy.utils.PeerBalance[],
+    appId: string
+  ): Promise<cf.legacy.node.Response> {
+    this.active = true;
+    const promise = new Promise<cf.legacy.node.Response>((resolve, reject) => {
+      this.runProtocolContinuation = resolve;
+    });
+    this.instructionExecutor.runUninstallProtocol(
+      fromAddress, toAddress, multisigAddress, peerAmounts, appId);
+    return promise;
```
snario: You can return the promise on this line without defining it above

packages/machine/test/integration/lifecycle/depositor.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: ack, I thought I'd removed this!

packages/machine/src/instructions.ts
```diff
@@ -41,109 +41,282 @@ export enum Opcode {
...
-    Opcode.OP_SIGN_VALIDATE,
-    Opcode.OP_SIGN,
-    Opcode.IO_PREPARE_SEND,
-    Opcode.IO_SEND,
-    Opcode.STATE_TRANSITION_COMMIT
-  ]
+export const FLOWS = {
+  [cf.legacy.node.ActionName.UPDATE]: {
+    0: [
+      (message, context, node) => {
```
ebryn: IMO we shouldn't be using fat arrow syntax if we're not actually using `this`. Just use regular `function() {`

packages/machine/src/instructions.ts
```diff
@@ -41,109 +41,282 @@ export enum Opcode {
...
+          clientMessage.data.responding;
+        context.intermediateResults.outbox[0].toAddress =
+          clientMessage.data.intermediary;
+      },
+      Opcode.IO_SEND
+
+      // // wait for self-remove
+      // Opcode.IO_WAIT
+    ]
+  }
```
ebryn: This file is kind of a monstrosity. Any ideas for how we could more clearly break this up? Perhaps into a file per "flow"?

packages/machine/src/middleware/middleware.ts
```diff
@@ -61,153 +33,39 @@ export class Middleware {
...
-      }
-      // This is hacky, prevents next from being called more than once
-      counter += 1;
-      const middleware = middlewares[opCode][counter];
-      if (opCode === Opcode.ALL || middleware.scope === opCode) {
-        return middleware.method(msg, callback, context);
-      }
-      return callback();
+    for (const middleware of middlewares[opCode]) {
+      await middleware.method(msg, () => {}, context);
```
ebryn: The callback argument is not correctly implemented here

packages/machine/src/instructions.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: eh, the two are completely equivalent for this function - why not use arrow syntax for everything?

packages/machine/test/integration/cf-operations.spec.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: This is used in a few places and I don't think there's any replacement for it; we can remove it in a separate PR

packages/machine/src/instruction-executor.ts
```diff
@@ -34,121 +31,169 @@ export class InstructionExecutor implements Observable {
...
-  /**
-   * Restarts all protocols that were stopped mid execution, and returns when
-   * they all finish.
-   */
-  public async resume(log: Log) {
-    const executions = this.buildExecutionsFromLog(log);
-    return executions.reduce(
-      (promise, exec) => promise.then(_ => this.run(exec)),
-      Promise.resolve()
+  public runUpdateProtocol(
```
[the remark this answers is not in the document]
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: hmm, actually I think it's not worth the downside of making the instruction executor's API harder to use

packages/machine/src/middleware/middleware.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I am choosing not to support this argument anymore

ebryn: I think it should be, as we previously discussed privately.

packages/machine/src/instructions.ts
```diff
@@ -1,149 +1,16 @@
...
-  /**
-   * Represents all instructions. Registering for this instruction will yield
-   * all messages.
-   */
-  ALL
-}
+import { INSTALL_FLOW } from "./install";
+import { METACHANNEL_INSTALL_APP_FLOW } from "./metachannel-install-app";
+import { SETUP_FLOW } from "./setup";
+import { UPDATE_FLOW } from "./update";
```
ebryn: Should these modules maybe be in a `flow` subdir?

packages/machine/src/middleware/middleware.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: can you summarize your argument here?

I would summarize mine as two points

1. I do not see any use-cases for supporting custom hooks apart from the existing ones implemented - messaging and signing
2. a promise-based interface between the existing custom hooks and the machine is sufficient

snario: (will discuss on wednesday, keeping `next` for now)

PR #255, opened 2018-11-21
[CFjs] Update spec: move events to Provider, remove proposeInstall event

mvanderh:
### Description

Update the CFjs spec after a productive call with @ebryn, @joelalejandro, @patience-tema-baron.

1) Add all event listeners to top-level Provider. Keep "updateState", "uninstall" and "proposeState" event listeners on AppInstance. dApps will likely be most interested in events for ALL app instances, and only in specific cases be interested in events relating to specific app instances.

2) dApps don't need to know when an installation is proposed. Install proposals go through the Playground (later the Wallet). dApps only need to know whether an app instance was succesfully installed, or rejected. 

+ some renames

### DIFF
--- a/packages/cf.js/API_REFERENCE.md
+++ b/packages/cf.js/API_REFERENCE.md
@@ -10,22 +10,24 @@
     - Lifecycle
         - `on(eventType, callback: Function)`
             - eventTypes
-                - `proposeInstall(proposal: {appInstanceId, appDefinition, terms}, function reject())`
                 - `install(appInstance)`
-                - `rejectInstall(proposal: {appInstanceId, appDefinition, terms})`
+                - `rejectInstall(appInstance)`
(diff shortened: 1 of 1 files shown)


packages/cf.js/API_REFERENCE.md
```diff
@@ -10,22 +10,24 @@
     - Lifecycle
         - `on(eventType, callback: Function)`
             - eventTypes
-                - `proposeInstall(proposal: {appInstanceId, appDefinition, terms}, function reject())`
                 - `install(appInstance)`
-                - `rejectInstall(proposal: {appInstanceId, appDefinition, terms})`
+                - `rejectInstall(appInstance)`
```
joelalejandro: @mvanderh @patience-tema-baron Is there are reason we offer an `appInstance` on the `rejectInstall` event? Isn't the `appInstance` created *after* `install`?

PR #260, opened 2018-11-22
[WIP] [Node] Add rough spec for Node JSON-RPC API

mvanderh:
### Description

Solidify 
- the methods and events that a Node is expected to implement
- error cases

Such that CF.js and the Node can be implemented in parallel while being sure that their interfaces will be compatible.

### DIFF
--- a/packages/node-provider/PROTOCOL_SPEC.md
+++ b/packages/node-provider/PROTOCOL_SPEC.md
@@ -1,10 +1,18 @@
 # Protocol Spec
 
-## Public JSON-RPC Methods
+## JSON-RPC calls
+
+Message fields:
+- `json-rpc`: set to `"2.0"`
+- `id`: set to a generated unique ID for this request
+- `method`: name of the method to call
+- `params`: array of parameters passed to the method call  
+
+## Methods
 
 | method         | params                                                       | result                                                                  |
 |----------------|--------------------------------------------------------------|-------------------------------------------------------------------------|
-| cf_getApps     | []                                                           | {id, definition, state, asset, myDeposit, peerDeposit, timeout}[]       |
+| cf_getApps     | []                                                           | list of {id, definition, state, asset, myDeposit, peerDeposit, timeout} |
 | cf_install     | [definition, state, asset, myDeposit, peerDeposit, timeout]  | {id}                                                                    |
 | cf_getState    | [id]                                                         | {state}                                                                 |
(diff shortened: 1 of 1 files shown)


packages/node-provider/PROTOCOL_SPEC.md
```diff
@@ -0,0 +1,69 @@
+# Protocol Spec
+
+## Public JSON-RPC Methods
+
+| method         | params                                                       | result                                                                  |
+|----------------|--------------------------------------------------------------|-------------------------------------------------------------------------|
+| cf_getApps     | []                                                           | {id, definition, state, asset, myDeposit, peerDeposit, timeout}[]       |
+| cf_install     | [definition, state, asset, myDeposit, peerDeposit, timeout]  | {id}                                                                    |
```
cf19drofxots: What about proposeInstall?

cf19drofxots: Based on https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md

Seems like `proposeInstall` would return the ID of the AppInstance, which would _then_ be used to make the actual install upon acceptance, correct?

joelalejandro:
@mvanderh Should there be any methods related to propose/accept/reject state? I'm not sure if `stateUpdate` represents those messages.

packages/node-provider/PROTOCOL_SPEC.md
```diff
@@ -0,0 +1,77 @@
...
+
+TODO
+
+
+
+## Events
+
+| event          | params                                                          |
+|----------------|-----------------------------------------------------------------|
+| install        | {id, definition, state, asset, myDeposit, peerDeposit, timeout} |
```
snario: Need type information

packages/node-provider/PROTOCOL_SPEC.md
```diff
@@ -0,0 +1,77 @@
...
+- `id`: set to a generated unique ID for this request
+- `method`: name of the method to call
+- `params`: array of parameters passed to the method call  
+
+## Methods
+
+| method         | params                                                       | result                                                                  |
+|----------------|--------------------------------------------------------------|-------------------------------------------------------------------------|
+| cf_getApps     | []                                                           | list of {id, definition, state, asset, myDeposit, peerDeposit, timeout} |
+| cf_install     | [definition, state, asset, myDeposit, peerDeposit, timeout]  | {id}                                                                    |
```
snario: need type information

PR #263, opened 2018-11-23
[poc] bridges cf.js into playground-poc for testing

snario:
@mvanderh 

- used `cuid` over `uuid`
- kept dependencies external
- removed lodash entirely

### DIFF
--- a/packages/cf.js/src/legacy/network.ts
+++ b/packages/cf.js/src/legacy/network.ts
@@ -1,5 +1,4 @@
-import { ethers } from "ethers";
-import * as _ from "lodash";
+import ethers from "ethers";
 
 import { Address } from "./utils";
 
@@ -43,7 +42,7 @@ export class NetworkContext {
 
   public linkedBytecode(unlinkedBytecode: string): string {
     let bytecode = unlinkedBytecode;
-    for (const contractName of _.keys(this.contractToVar)) {
+    for (const contractName of Object.keys(this.contractToVar)) {
       const regex = new RegExp(`__${contractName}_+`, "g");
       const address = this[this.contractToVar[contractName]].substr(2);
       bytecode = bytecode.replace(regex, address);
--- a/packages/cf.js/src/legacy/network.ts
+++ b/packages/cf.js/src/legacy/network.ts
@@ -42,7 +42,7 @@ export class NetworkContext {
 
   public linkedBytecode(unlinkedBytecode: string): string {
     let bytecode = unlinkedBytecode;
-    for (const contractName of Object.keys(this.contractToVar)) {
+    for (const contractName in this.contractToVar) {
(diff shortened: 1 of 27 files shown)


snario:
I had to use `import * as ethers from "ethers"` b/c ethers does not support `import ethers from "ethers"` apparently (the tests treat it as undefined). Then, `import { ethers } from ethers` also doesn't work because the rollup can't find the subproperty because the browser version doesn't expose `ethers.ethers`.

packages/cf.js/src/legacy/network.ts
```diff
@@ -43,7 +42,7 @@ export class NetworkContext {
 
   public linkedBytecode(unlinkedBytecode: string): string {
     let bytecode = unlinkedBytecode;
-    for (const contractName of _.keys(this.contractToVar)) {
+    for (const contractName of Object.keys(this.contractToVar)) {
```
mvanderh: I think you can even just use `for (const contractName in this.contractToVar)`

packages/cf.js/src/legacy/utils/index.ts
```diff
@@ -31,5 +30,5 @@ export async function mineBlocks(
   n: number,
   provider: ethers.providers.JsonRpcProvider
 ) {
-  _.times(n, async () => await mineOneBlock(provider));
+  [...Array(n)].forEach(async () => await mineOneBlock(provider));
```
mvanderh: `Array(n).forEach` doesn't work?

packages/playground-poc/README.md
```diff
@@ -0,0 +1,12 @@
...
+```
+python -m SimpleHTTPServer
+```
+Here
+```
+node server
+```
+open
+```
+localhost:8080
```
mvanderh: I think you mean `localhost:8000`

packages/playground-poc/public/index.html
```diff
@@ -20,9 +20,13 @@
     <script src="https://unpkg.com/eventemitter3@latest/umd/eventemitter3.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/vex-js/4.1.0/js/vex.combined.min.js"></script>
     <script src="/socket.io/socket.io.js"></script>
-    <script src="/scripts/mini-node.js"></script>
-    <script src="/scripts/dapp.js"></script>
-    <script src="/scripts/playground.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/cuid/1.3.8/browser-cuid.min.js"></script>
+    <script src="http://localhost:8000/node_modules/ethers/dist/ethers.js"></script>
```
mvanderh: Ah so packaging ethers into the IIFE didn't work?

mvanderh: 👍

PR #264, opened 2018-11-23
[poc] bridges node and playground-poc

cf19drofxots:
Main file of interest: packages/node/src/node.ts


packages/node/src/channel.ts
```diff
@@ -0,0 +1,21 @@
+import * as cf from "@counterfactual/cf.js";
```
snario: should avoid *

packages/node/src/channel.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Not strictly needed, added for easier legibility
```
snario: probably bad variable naming if this comment exists

packages/node/src/channel.ts
```diff
@@ -0,0 +1,21 @@
...
+export type MultisigAddress = string;
+
+export class StateChannelInfo implements cf.legacy.channel.StateChannelInfo {
```
snario: do we really need this `cf.legacy.channel.StateChannelInfo`

packages/node/src/index.ts
```diff
@@ -1,3 +1,5 @@
-import Node from "./node";
+import { Node, NodeOptions } from "./node";
 
-export { Node };
+export function create(options: NodeOptions) {
```
snario: create wrappers around `new` keyword usually a bad idea

packages/node/src/node.ts
```diff
@@ -1 +1,113 @@
-export default class Node {}
+import * as cf from "@counterfactual/cf.js";
```
snario: *

packages/node/src/node.ts
```diff
@@ -1 +1,113 @@
...
+import { EventEmitter } from "events";
+import * as _ from "lodash";
```
snario: leaning towards getting rid of lodash, would recommend implementing this file without it

packages/node/src/node.ts
```diff
@@ -1 +1,113 @@
...
+
+import { MultisigAddress, StateChannelInfo } from "./channel";
+
+export interface NodeOptions {
```
snario: is this necessary? why not just put it in the constructor and only abstract it as an interface if like 2 or more other classes end up using the same thing. seems like we're adding too many unnecessary things like this in the code

packages/node/src/node.ts
```diff
@@ -1 +1,113 @@
...
+
+/**
+ * The Node class is the entry point to the @counterfactual/node package.
+ * It's responsible for:
+ * - encapsulating channels and their apps
+ * - connecting with a provided database to persistently store commitments
+ * - registering a signing service
+ */
+export class Node {
+  private channels: Map<MultisigAddress, StateChannelInfo> = new Map();
```
snario: should abstract this inside a database connection; can start with redis

packages/node/src/node.ts
```diff
@@ -1 +1,113 @@
...
+  private eventEmitters: Map<string, EventEmitter> = new Map();
+
+  constructor(options: NodeOptions) {}
+
+  /**
+   * Given a channel, this adds said channel to the state of the Node if a
+   * channel with the same multisig address does not exist.
+   * @param channel The channel to keep track of.
+   */
+  openChannel(channel: StateChannelInfo): [boolean, string] {
```
snario: returning an array of two things is weird

snario: also is this just a mock? all it seems to do is set a variable atm

packages/node/src/node.ts
```diff
@@ -1 +1,113 @@
...
+   * TODO: add auth token to retrieve only authorized channel(s)
+   */
+  getChannels(): Map<MultisigAddress, StateChannelInfo> {
+    return this.channels;
+  }
+
+  /**
+   * Returns all of the apps installed across all of the channels in the Node.
+   */
+  getApps(): cf.legacy.app.AppInstanceInfo[] {
```
snario: if we can avoid re-using these legacy types I think  it will help us

packages/node/src/node.ts
```diff
@@ -1 +1,113 @@
...
+    const apps: cf.legacy.app.AppInstanceInfo[] = [];
+    this.channels.forEach(
+      (channel: StateChannelInfo, multisigAddress: string) => {
+        _.values(channel.appInstances).forEach(appInstance => {
```
snario:
```suggestion
        Object.values(channel.appInstances).forEach(appInstance => {
```

packages/node/test/unit/apps.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { Node } from "../../src/node";
+
+describe("Apps", () => {
```
snario: Test cases should be more descriptive

packages/node/test/unit/channels.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    cf.legacy.constants.B_ADDRESS,
+    cf.legacy.constants.A_ADDRESS,
+    cf.legacy.constants.C_ADDRESS,
+    {},
+    cf.legacy.constants.EMPTY_FREE_BALANCE
+  );
+
+  beforeEach(() => {
+    node = new Node({});
+  });
```
snario: with this small of a beforeEach you might as well just write the one line in each it clause

packages/playground-poc/public/index.html
```diff
@@ -1,9 +1,25 @@
...
+      body {
+        font-family: sans-serif;
+      }
+      fieldset {
+        display: inline-block;
+      }
+      iframe {
+        display: block;
+      }
+    </style>
```
snario: can we undo this changeset to keep them single lines

packages/playground-poc/public/scripts/node.js
```diff
@@ -0,0 +1 @@
+../../../node/dist/index.iife.js
```
snario: is this a symlink? does it work with `express.static`? I tried it before and it didn't but if this works gr8

packages/playground-poc/public/scripts/playground.js
```diff
@@ -4,7 +4,7 @@ class Playground {
     this.socket = null;
     this.user = null;
     this.appManifests = appManifests;
-    this.node = new PlaygroundNode();
+    this.node = cfNode.create();
```
snario: again, I don't understand this obsession with `.create()`

packages/node/src/index.ts
cf19drofxots: Doing `new cfNode.Node()` is rather awkward tbh

packages/node/src/node.ts
cf19drofxots: When was this decision made / what's the rationale? My understanding was that from a while ago we decided to converge on using lodash

packages/node/src/index.ts
snario: Also why prefix with `cf`? We internally call this thing `Node` but thats just a super vanilla name, we can come up with something better

packages/node/src/node.ts
snario: Yeah we should be more explicit about it but I did some thinking and looking around and general consensus right now is that for 90% of the ways people (including us) use lodash, ES6 does that for you. e.g., https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

packages/playground-poc/public/scripts/node.js
snario: oh cool @mvanderh you can prob do this in the other PR #263

mvanderh: neat!

packages/node/src/channel.ts
mvanderh: Seconded. Legacy will be phased out soon, and CFjs.2 won't have a concept of State Channel (might change when metachannels are implemented). Just create your own StateChannelInfo without implementing the legacy interface

packages/node/src/channel.ts
cf19drofxots: This will probably not even be needed. Will remove.

packages/node/src/node.ts
cf19drofxots: WIP on another branch

packages/node/src/node.ts
cf19drofxots: @mvanderh if/when they're available, lmk - I'll update this asap

packages/node/test/unit/apps.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+  let node: Node;
+  beforeEach(() => {
+    node = new Node();
+  });
+
+  it("can open a connection with an app", () => {
+    const appInstanceEventEmitter = node.openApp("1");
+    expect(appInstanceEventEmitter).toBeDefined();
```
mvanderh: Is this really the best test of a connection? Can't you send a test message?

cf19drofxots: This test has a poor description. No listeners exist for this event emitter to be useful at the moment.

When they do, appropriate tests for them will be added. This should just be confirming that an EventEmitter is given back.

mvanderh: give it a better description?

cf19drofxots: ~can open a connection with an app~

packages/node/src/node.ts
mvanderh: You can now take https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#data-type-appinstanceinfo and turn it into a Typescript interface & use it.

cf19drofxots: sounds good

mvanderh: Also, how is this supposed to be called? I see nowhere outside of tests that this is called.

I'm assuming you're gonna listen for a "getAppInstances" event on app instance event emitters.

cf19drofxots: correct

mvanderh: @cf19drofxots 

1. Yeah, not sure where the Node protocol data types should live. My intuition says CFjs since that's where the spec is located. 

2. I'm confused. I think you're wrong and that we _do_ want to call getAppInstances through the Node/NodeProvider EventEmitter. 

Also I don't get why you call it an "AppInstance EventEmitter" because in my mind a Node shouldn't have to have a concept of a single AppInstance. It only needs a concept of "something that wants to communicate with the Node".

packages/node/src/node.ts
```diff
@@ -1 +1,64 @@
...
+   * Opens a connection specifically for this app with the consumer of this
+   * node.
+   * @returns An EventEmitter to emit events related to this app for consumers
+   * subscribing to app updates.
+   */
+  openApp(appInstanceID: string): EventEmitter {
+    const appInstanceEventEmitter = new EventEmitter();
+    this.setupListeners(appInstanceEventEmitter);
+    this.eventEmitters.set(appInstanceID, appInstanceEventEmitter);
+    return appInstanceEventEmitter;
```
mvanderh: Why does the appInstanceID matter at all? Why not simply return an event emitter that's correctly set up, without recording app instance ID?

cf19drofxots: This might change with the updated playground work.

mvanderh: Ok

cf19drofxots:
This is a bit stale now - closing in favor of starting with more fresh commits.

PR #269, opened 2018-11-26
add cf.js data types

patience-tema-baron:
Moves the cf.js types into `src/types` and adds the new data types specified [here](https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#data-types).

### DIFF
--- a/packages/cf.js/src/types/simple-types.ts
+++ b/packages/cf.js/src/types/simple-types.ts
@@ -3,5 +3,5 @@ export type AppInstanceID = string;
 export type Address = string;
 export type Bytes32 = string;
 
-export type AppState = object;
-export type AppAction = object;
+export type AppState = any;
+export type AppAction = any;
--- a/packages/cf.js/src/types/simple-types.ts
+++ b/packages/cf.js/src/types/simple-types.ts
@@ -2,3 +2,6 @@ export type ABIEncoding = string;
 export type AppInstanceID = string;
 export type Address = string;
 export type Bytes32 = string;
+
+export type AppState = object;
+export type AppAction = object;
--- a/packages/cf.js/src/types/structs.ts
+++ b/packages/cf.js/src/types/structs.ts
@@ -1,0 +1,28 @@
+import { ABIEncoding, Address } from "./simple-types";
+import { BigNumber } from "ethers/utils";
+
+export interface AppDefinition {
+  address: Address;
+  appStateEncoding: ABIEncoding;
+  appActionEncoding: ABIEncoding;
+}
+
+export interface AppInstanceInfo {
+  id: string;
+  appId: string;
+  abiEncodings: AppABIEncodings;
+  asset: BlockchainAsset;
+  myDeposit: BigNumber;
+  peerDeposit: BigNumber;
+  timeout: BigNumber;
+}
+
+export interface AppABIEncodings {
+  stateEncoding: string;
+  actionEncoding?: string;
+}
+
+export interface BlockchainAsset {
+  assetType: number;
+  token?: string;
+}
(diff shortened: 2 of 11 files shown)


packages/cf.js/src/types/structs.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface AppABIEncodings {
+  stateEncoding: string;
+  actionEncoding?: string;
+}
+
+export interface BlockchainAsset {
+  assetType: number;
+  token?: string;
```
mvanderh: Make this an Address

packages/cf.js/src/types/simple-types.ts
```diff
@@ -2,3 +2,6 @@ export type ABIEncoding = string;
 export type AppInstanceID = string;
 export type Address = string;
 export type Bytes32 = string;
+
+export type AppState = object;
+export type AppAction = object;
```
mvanderh: Technically these could also be simple `numbers` or `bytes` which wouldn't be translated into Typescript as `object`s. Better to make these `any`

packages/cf.js/src/types/structs.ts
```diff
@@ -0,0 +1,28 @@
...
+  abiEncodings: AppABIEncodings;
+  asset: BlockchainAsset;
+  myDeposit: BigNumber;
+  peerDeposit: BigNumber;
+  timeout: BigNumber;
+}
+
+export interface AppABIEncodings {
+  stateEncoding: string;
+  actionEncoding?: string;
```
mvanderh: These should be `ABIEncoding`s

packages/cf.js/src/types/structs.ts
```diff
@@ -0,0 +1,28 @@
+import { ABIEncoding, Address } from "./simple-types";
+import { BigNumber } from "ethers/utils";
+
+export interface AppDefinition {
```
mvanderh: Do we still need this?

packages/cf.js/src/types/structs.ts
```diff
@@ -0,0 +1,28 @@
...
+  timeout: BigNumber;
+}
+
+export interface AppABIEncodings {
+  stateEncoding: string;
+  actionEncoding?: string;
+}
+
+export interface BlockchainAsset {
+  assetType: number;
```
mvanderh: Create an enum with the possible values to make this easier to read.

packages/cf.js/src/types/structs.ts
```diff
@@ -0,0 +1,28 @@
+import { ABIEncoding, Address } from "./simple-types";
+import { BigNumber } from "ethers/utils";
+
```
mvanderh: Add a link to the source documentation here

packages/cf.js/src/types/structs.ts
cf19drofxots: Structs are typically more than just interfaces, so having interfaces in a structs file seems kinda weird.

Perhaps either just put them in the `src/types` file or in `src/types/index.ts`?

packages/cf.js/src/types/structs.ts
```diff
@@ -0,0 +1,28 @@
+import { ABIEncoding, Address } from "./simple-types";
```
mvanderh: Maybe rename this file to something more specific e.g. protocol-types?

packages/cf.js/src/types/structs.ts
mvanderh: I don't like putting real code in index.ts. Prefer index.ts just be an exports aggregator.

cf19drofxots: Yeah me neither actually. Was a poor suggestion.

patience-tema-baron: We're still using it throughout cf.js, including sometimes referencing `appDefinition.address`. We couldn't remove it without breaking that code, and there's no obvious replacement for it either.

PR #272, opened 2018-11-27
[machine] create `lifecycle/` folder

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR breaks up the `lifecycle.spec.ts` file and creates a `lifecycle` folder


packages/machine/test/integration/lifecycle/depositor.ts
```diff
@@ -0,0 +1,232 @@
...
+      peerB,
+      peerA,
+      ethers.utils.bigNumberify(5),
+      ethers.utils.bigNumberify(10)
+    );
+  }
+
+  /**
+   * @param amountA is the amount wallet A wants to deposit into the channel.
+   * @param amountBCumualtive is the amount wallet B already has in the channel,
```
snario: typo in a few places: "Cumulative"

packages/machine/test/integration/lifecycle/depositor.ts
```diff
@@ -0,0 +1,232 @@
...
+    peerB: TestResponseSink,
+    threshold: ethers.utils.BigNumber
+  ) {
+    const msg = Depositor.startInstallBalanceRefundMsg(
+      peerA.signingKey.address!,
+      peerB.signingKey.address!,
+      threshold
+    );
+    const response = await peerA.runProtocol(msg);
+    expect(response.status).toEqual(cf.legacy.node.ResponseStatus.COMPLETED);
```
snario: Seems odd to me that the `ResponseStatus` type is in `cf.legacy.node`.

packages/machine/test/integration/lifecycle/depositor.ts
```diff
@@ -0,0 +1,232 @@
...
+    // since the machine is async, we need to wait for peerB to finish up its
+    // side of the protocol before inspecting it's state
+    await cf.legacy.utils.sleep(50);
```
snario: Should take this out of `cf.legacy` and just put it in the file.

packages/machine/test/integration/lifecycle/depositor.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `ResponseStatus` is in a few places elsewhere as well - we should fix those in a separate PR

PR #273, opened 2018-11-27
[node-provider] Implements basic API

joelalejandro:
This PR implements the basic methods needed for NodeProvider to work:

- `sendMessage()`: used to pass messages *to* the dApp.
- `onMessage()`: used to subscribe callbacks when a message is received.
- `connect()`: implements the handshake between the NodeProvider and the dApp to share a MessageChannel binding two MessagePorts.


packages/node-provider/src/node-provider.ts
```diff
@@ -1,7 +1,35 @@
...
+  NodeProviderStatus
+} from "./types";
 
 export default class NodeProvider implements INodeProvider {
-  public onMessage(callback: (message: NodeMessage) => void) {}
+  private internalStatus: NodeProviderStatus = NodeProviderStatus.OFFLINE;
+
+  constructor() {}
+
+  public onMessage(callback: NodeMessageReceivedCallback) {}
```
joelalejandro: Should we support subscribing to any incoming message with `onMessage` and to specific message types with `on/once`?

packages/node-provider/test/unit/node-provider.spec.ts
```diff
@@ -1,7 +1,37 @@
...
+
+    expect(nodeProvider.status).toEqual(NodeProviderStatus.CONNECTED);
+  });
+  it("should post a message", () => {
+    const nodeProvider = new NodeProvider();
+
+    nodeProvider.postMessage({
+      messageType: NodeMessageType.INSTALL,
+      requestId: "0",
+      data: null
```
joelalejandro: I've noticed the `install` type has no expected data payload. Is this accurate?

packages/node-provider/src/types.ts
```diff
@@ -1,12 +1,25 @@
...
+export enum NodeProviderStatus {
+  OFFLINE = "offline",
+  CONNECTED = "connected"
 }
 
 export enum NodeMessageType {
   INSTALL = "install",
   QUERY = "query",
   ERROR = "error"
+  // Should we add a STATUS_CHANGED type to properly
```
joelalejandro: At the PoC work, we added `cf-node-provider:init` and `cf-node-provider:ready` internal messages to configure the MessagePort instances required to connect the Playground to the iframe. I think `NodeMessageType` refers only to Node-specific messages. How would we handle these _handshake_ messages?

packages/node-provider/src/node-provider.ts
```diff
@@ -1,7 +1,35 @@
...
+  INodeProvider,
+  NodeMessage,
+  NodeMessageReceivedCallback,
+  NodeMessageType,
+  NodeProviderStatus
+} from "./types";
 
 export default class NodeProvider implements INodeProvider {
-  public onMessage(callback: (message: NodeMessage) => void) {}
+  private internalStatus: NodeProviderStatus = NodeProviderStatus.OFFLINE;
```
joelalejandro: Since we require a `connect` method to bridge the Playground to the iframe, I thought this would be a good idea to prevent unwanted operations until the NodeProvider is fully connected. Thoughts?

packages/node-provider/src/node-provider.ts
ebryn: I think we can just go with `onMessage` for now. Reviewing our usage of the NodeProvider in CF.js makes me think it's unnecessary at this point for us to have to subscribe callbacks to specific message types rather than just have one we can dispatch from based on type.

packages/node-provider/src/types.ts
ebryn: Those are strictly internal messages for the playground and the NodeProvider to coordinate over postMessage. I don't think they need to be typed.

packages/node-provider/src/node-provider.ts
ebryn: Seems fine

ebryn: I think I like `isConnected` better

packages/node-provider/test/unit/node-provider.spec.ts
ebryn: I think we need to split `NodeMesage` into two separate request and response types

mvanderh: https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#method-install 

Install expects an `appInstanceId` param.

Also I think the split should be elsewhere @ebryn: between messages for "Methods" and for "Events". The former having a defined request-response format and the latter being spontaneous.

packages/node-provider/src/node-provider.ts
mvanderh: This `connect` method is only a temporary solution until the node runs in an extension/mobile app, right? The original idea was that a fully-initialized NodeProvider would be injected into the page.

packages/node-provider/src/node-provider.ts
mvanderh: `on/once` also requires extra maintenance to ensure CFjs subscribes to all the right message types. 

Instead, CFjs can (and will) expose `on/once` convenience methods.

packages/node-provider/src/types.ts
mvanderh: I think they should be typed but not exposed. It's almost always valuable to have type checks.

packages/node-provider/src/node-provider.ts
ebryn: @mvanderh I’m not sure it should be temporary. I am not particularly fond of the idea of injecting a global variable into every webpage, but that can be a debate for another day

packages/node-provider/src/types.ts
ebryn: they are implementation details of how we setup the communication between the iframe and parent frame. it was a misunderstanding to even bring up the idea of typing them as they are just strings passed to postMessage which is a DOM built in which takes any

packages/node-provider/src/node-provider.ts
mvanderh: For comparison, Metamask and other dApp wallets inject a global `ethereum` object into the browser context.  But yes, let's talk another day.

ebryn: Yeah I’m familiar with that and think it’s not great

packages/node-provider/src/types.ts
```diff
@@ -1,12 +1,25 @@
+export type NodeMessageReceivedCallback = (message: NodeMessage) => void;
+
 export interface INodeProvider {
-  onMessage(callback: (message: NodeMessage) => void);
+  status: NodeProviderStatus;
+  onMessage(callback: NodeMessageReceivedCallback);
   postMessage(message: NodeMessage);
```
mvanderh: When me and @snario whiteboarded CFjs we concluded that this method creates confusion with the browser PostMessage protocol used between NodeProvider and the Playground. 

"sendMessage" or something similar would be better.

ebryn: `sendMessage` is cool with me

packages/node-provider/src/node-provider.ts
```diff
@@ -1,7 +1,66 @@
...
+  public async connect(): Promise<NodeProvider> {
+    if (this.isConnected) {
+      console.warn("NodeProvider is already connected.");
+      return Promise.resolve(this);
+    }
+
+    return new Promise<NodeProvider>(this.getMessagePort.bind(this));
+  }
+
+  private getMessagePort(resolve, reject) {
```
mvanderh: I think you should inline this method. It's confusing that it's separate from the method in which the Promise is created.

packages/node-provider/src/node-provider.ts
```diff
@@ -1,7 +1,66 @@
-import { INodeProvider, NodeMessage } from "./types";
+import EventEmitter from "eventemitter3";
+
+import {
+  INodeProvider,
+  NodeMessage,
+  NodeMessageReceivedCallback
+} from "./types";
```
mvanderh: I think these types should be moved to cfjs/types. If not, then CFjs becomes dependent on the NodeProvider package, which is weird. Plus, these types correspond to the written spec inside CFjs.

packages/node-provider/src/types.ts
```diff
@@ -1,6 +1,9 @@
+export type NodeMessageReceivedCallback = (message: NodeMessage) => void;
+
 export interface INodeProvider {
```
mvanderh: I think these types should be moved to the CFjs package. Otherwise CFjs is dependent on NodeProvider for types, which is a weird relationship (CFjs should really not depend on any other Counterfactual packages). Plus, these types correspond to the written spec in CFjs (apart from INodeProvider).

packages/node-provider/test/unit/node-provider.spec.ts
```diff
@@ -1,7 +1,54 @@
...
+        },
+        initEvent() {},
+        preventDefault() {},
+        stopImmediatePropagation() {},
+        stopPropagation() {},
+        AT_TARGET: 0,
+        BUBBLING_PHASE: 1,
+        CAPTURING_PHASE: 2,
+        NONE: 3
+      });
```
snario: What's up with this huge object? Maybe needs some documentation

joelalejandro: It's a mock for a MessageEvent instance, it's basically everything an event has when it's triggered via `addEventListener`.

packages/node-provider/rollup.config.js
```diff
@@ -17,7 +18,11 @@ export default [
       }
     ],
     plugins: [
-      typescript()
+      typescript(),
+      resolve({
+        main: true
+      }),
+      commonjs(),
```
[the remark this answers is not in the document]
joelalejandro: @mvanderh You can compile to IIFE without it, but it won't bundle dependencies. NodeProvider depends on `eventemitter3`; if I don't add this, it'll regard `eventemitter3` as an external library exposed via a global variable. I don't think that's something we want.

packages/node-provider/src/types.ts
joelalejandro: @mvanderh I'll defer to your best judgement here. These types came to NodeProvider through you :P

packages/node-provider/rollup.config.js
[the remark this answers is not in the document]
cf19drofxots: We've been using the practice of not bundling up external dependencies and relying on the library being available in the context.

packages/node-provider/src/types.ts
mvanderh: Yeah I know, different ideas in different times :p

packages/node-provider/rollup.config.js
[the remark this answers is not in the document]
joelalejandro: That approach makes sense for peerDependencies, not necessarily explicit dependencies. I'm not sure I agree with loading dependencies like that on a browser. I wouldn't want to load 3-4 script files if I can have just one.

packages/node-provider/src/node-provider.ts
```diff
@@ -1,7 +1,59 @@
 import * as cf from "@counterfactual/cf.js";
+import EventEmitter from "eventemitter3";
 
 export default class NodeProvider implements cf.types.INodeProvider {
-  public onMessage(callback: (message: cf.types.Node.Message) => void) {}
+  private isConnected: boolean;
```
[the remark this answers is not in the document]
joelalejandro: 👍 Sure!

PR #275, opened 2018-11-27
[WIP] Node API spec

cf19drofxots:
### Description 

Solidifies the first version of the Node public API spec


cf19drofxots:
Re-opening when relevant later

PR #276, opened 2018-11-28
[node] Initial setup

cf19drofxots:
### Description

This implements the basic structure of the Node so cf.js and the playground can start using it.

More specifically: https://github.com/counterfactual/monorepo/blob/998630fe1ffc6f1eb51bea2a893dfb911128222f/packages/node/test/unit/get-app-instances.spec.ts#L10-L31

Follow up is to use the firestore to enable Node-to-Node communication.


packages/node/rollup.config.js
```diff
@@ -1,9 +1,11 @@
 import typescript from "rollup-plugin-typescript2";
 
 import pkg from "./package.json";
+
 const globals = {
-  ethers: "ethers",
-};
+  "@counterfactual/cf.js": "cfjs",
```
joelalejandro: @cf19drofxots The global here is actually `cf`.

packages/node/rollup.config.js
```diff
@@ -1,9 +1,11 @@
...
+  "eventemitter3": "EventEmitter3"
```
joelalejandro: @cf19drofxots I think this one is `EventEmitter`.

packages/node/src/node.ts
```diff
@@ -1 +1,77 @@
-export default class Node {}
+import * as cf from "@counterfactual/cf.js";
+import machine from "@counterfactual/machine";
+import * as ethers from "ethers";
+import { EventEmitter } from "eventemitter3";
+import firebase from "firebase";
+
+export class Node extends EventEmitter {
```
joelalejandro: @cf19drofxots If it's the only class in the file, should we use `export default`?

packages/node/src/node.ts
```diff
@@ -1 +1,77 @@
...
+  // @ts-ignore
```
joelalejandro: @cf19drofxots What's with these `@ts-ignore` here and in the constructor?

packages/node/src/node.ts
joelalejandro: @cf19drofxots Instead of extending from `EventEmitter`, maybe it's best to delegate to a private instance `this.eventEmitter`, so we can white-list events listened via `on`.

packages/node/rollup.config.js
joelalejandro: Weird. Never seen it injected like that :/

packages/node/rollup.config.js
joelalejandro: Yeah, that's because the UMD version does that. We'd arguably use a different version without that extra namespace.

cf19drofxots: Since there are no consumers yet to care about this, I'll change this to `EventEmitter`

packages/node/test/unit/node.spec.ts
```diff
@@ -1,12 +1,55 @@
...
+    const node = new Node(privateKey, provider, firestore, networkContext);
+    const requestId = "1";
+    const req: cf.types.Node.MethodRequest = {
+      requestId,
+      type: cf.types.Node.MethodName.GET_APP_INSTANCES,
+      params: {}
+    };
+
+    // Set up listener for the method response
+    node.on(requestId, (res: cf.types.Node.MethodResponse) => {
```
ebryn: This doesn't look right to me. The event we're listening for shouldn't be the `requestId`, it should probably be `req.type` and the response object would have a `requestId` property.

packages/node/src/node.ts
```diff
@@ -1 +1,99 @@
...
+    this.eventEmitter.on(
+      cf.types.Node.MethodName.GET_APP_INSTANCES,
+      (req: cf.types.Node.MethodRequest) => {
+        // TODO: empty for now
+        const res: cf.types.Node.MethodResponse = {
+          type: req.type,
+          requestId: req.requestId,
+          result: this.getAppInstances()
+        };
+        this.eventEmitter.emit(req.requestId, res);
```
mvanderh: Shouldn't this be `emit(res.type, res)`?

packages/node/src/node.ts
```diff
@@ -1 +1,99 @@
...
+    this.eventEmitter.on(cf.types.Node.MethodName.GET_STATE, () => {});
+    this.eventEmitter.on(
+      cf.types.Node.MethodName.GET_APP_INSTANCE_DETAILS,
+      () => {}
+    );
+    this.eventEmitter.on(cf.types.Node.MethodName.TAKE_ACTION, () => {});
+    this.eventEmitter.on(cf.types.Node.MethodName.UNINSTALL, () => {});
+    this.eventEmitter.on(cf.types.Node.MethodName.PROPOSE_INSTALL, () => {});
+    this.eventEmitter.on(cf.types.Node.MethodName.ACCEPT_STATE, () => {});
+    this.eventEmitter.on(cf.types.Node.MethodName.REJECT_INSTALL, () => {});
```
mvanderh: You should probably remove these instead of attaching empty listeners to events. You can add them as you implement them.

packages/node/rollup.config.js
```diff
@@ -14,6 +16,14 @@ export default [
...
         format: "cjs",
         globals: globals
+      },
+      {
+        file: pkg.iife,
+        sourcemap: true,
+        name: "window",
+        format: "iife",
+        extend: true,
+        globals: globals
```
mvanderh: Why do you need iife? Won't the Node only be used in modules with existing build processes?

packages/node/src/node.ts
```diff
@@ -1 +1,99 @@
...
+    // TODO:
+    return Promise.resolve(ethers.constants.AddressZero);
+  }
+
+  /**
+   * Delegates setting up a listener to the Node's EventEmitter instance.
+   * @param event
+   * @param callback
+   */
+  on(event: string, callback: (res: cf.types.Node.MethodResponse) => void) {
```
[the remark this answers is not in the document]
mvanderh: 1) Because that's where the protocol spec is. It'd be weird to have the spec be in a different package from the type definitions.
2) I see CFjs as the "common package" between other CF packages. In which case it would make to house Node protocol types as they're used by CFjs, Node and NodeProvider.

packages/node/test/unit/node.spec.ts
mvanderh: Seconded.

packages/node/src/node.ts
```diff
@@ -1 +1,99 @@
...
+  /**
+   * This sets up all the listeners for the methods the Node is expected to have
+   * as described at https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#node-protocol
+   *
+   * The responses to these calls are the events being listened on
+   * https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#events
+   */
+  private registerListeners() {
+    this.eventEmitter.on(
+      cf.types.Node.MethodName.GET_APP_INSTANCES,
```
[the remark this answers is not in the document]
snario: Also makes no sense why the type is in a different package

packages/node/src/node.ts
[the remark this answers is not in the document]
snario: This is a good case for the `typescript-typings` package I think. Feels like we should put types that multiple packages rely on inside there.

packages/node/rollup.config.js
cf19drofxots: If the playground builds it, then sure.

Until then, the poc directly assumed there was a build available and just used it (like this).

cc @joelalejandro

packages/node/src/node.ts
```diff
@@ -1 +1,99 @@
...
+export default class Node {
+  // @ts-ignore
+  private instructionExecutor: machine.InstructionExecutor = Object.create(
+    null
+  );
+  private eventEmitter: EventEmitter;
+
+  constructor(
+    // @ts-ignore
+    private privateKey: string,
```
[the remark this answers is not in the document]
cf19drofxots: How so?

If the consumer (for whatever reason) doesn't restrict access to the instantiated node, making these readonly can give unwanted access to whoever has access to the consumer as well.

What's the benefit for making them readonly as opposed to private?

packages/node/src/node.ts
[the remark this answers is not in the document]
cf19drofxots: I can extract the Node types to make the syntax better here, and this can be updated whenever/if the types move out.

packages/node/src/node.ts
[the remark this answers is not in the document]
cf19drofxots: Yeah will do

packages/node/src/node.ts
cf19drofxots: Sure.

packages/node/test/unit/node.spec.ts
cf19drofxots: That also makes sense.

This is why I brought this up earlier and I thought we agreed the event type would be the request ID, hence this.

Will update.

packages/node/src/node.ts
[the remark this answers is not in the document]
mvanderh: @cf19drofxots Do that, i'll handle the moving of the types in a later PR.

packages/node/src/node.ts
cf19drofxots: Will update this as well.

packages/node/src/node.ts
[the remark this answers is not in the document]
mvanderh: Err sure. I was just letting you know that they're not mutually exclusive like you seemed to believe.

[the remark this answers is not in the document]
cf19drofxots: Yup I misremembered.

packages/node/test/unit/node.spec.ts
```diff
@@ -1,12 +1,61 @@
...
     expect(Node).toBeDefined();
   });
 
   it("can be instantiated", () => {
-    const node = new Node();
+    const node = new Node(privateKey, provider, firestore, networkContext);
     expect(node).toBeDefined();
   });
+
+  it("can create a multisig", async () => {
```
snario: Can you flesh out this test -- should not be a mock.

packages/node/test/unit/node.spec.ts
```diff
@@ -1,12 +1,61 @@
...
+    const node = new Node(privateKey, provider, firestore, networkContext);
+    const multisig = await node.createMultisig(ethers.constants.AddressZero);
+    expect(multisig).toBeDefined();
+  });
+
+  it("can accept call to get app instances", done => {
```
snario: We should have a test file per method type in my opinion and make sure we test every kind of input and potential response.

cf19drofxots: Yeah on second thought that sounds cleaner. Will separate out.

packages/node/test/unit/node.spec.ts
cf19drofxots: I'll remove this method/test for the time being then - it shouldn't even be in this PR actually.

The task to implement that can be another PR.

packages/node/src/node-types.ts
```diff
@@ -0,0 +1,8 @@
+import { types } from "@counterfactual/cf.js";
+
+export import MethodRequest = types.Node.MethodRequest;
+export import MethodResponse = types.Node.MethodResponse;
+export import GetAppInstancesResult = types.Node.GetAppInstancesResult;
+export import Address = types.Address;
+export import AppInstanceInfo = types.AppInstanceInfo;
+export import MethodName = types.Node.MethodName;
```
snario: why not use typescript-typings

packages/node/src/node.ts
```diff
@@ -1 +1,96 @@
...
+   * sent to listeners expecting a request message.
+   **/
+  private incoming: EventEmitter;
+  private outgoing: EventEmitter;
+
+  constructor(
+    // @ts-ignore
+    private readonly privateKey: string,
+    // @ts-ignore
+    private readonly provider: ethers.providers.Web3Provider,
```
snario: Should be the more generic type of provider; we will want to pass in `localhost:8545` for example

packages/node/src/node.ts
```diff
@@ -1 +1,96 @@
...
+    // @ts-ignore
```
snario: we shouldn't be ts-ignoring ever

packages/node/src/node.ts
```diff
@@ -1 +1,96 @@
...
+  /**
+   * This sets up all the listeners for the methods the Node is expected to have
+   * as described at https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#node-protocol
+   *
+   * The responses to these calls are the events being listened on
+   * https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#events
+   */
+  private registerListeners() {
+    this.incoming.on(MethodName.GET_APP_INSTANCES, (req: MethodRequest) => {
+      // TODO: empty for now
```
snario: is this an outdated todo?

packages/node/test/unit/node.spec.ts
```diff
@@ -1,12 +1,61 @@
...
 import Node from "../../src/node";
+import {
+  Address,
+  AppInstanceInfo,
+  MethodName,
+  MethodRequest,
+  MethodResponse
+} from "../../src/node-types";
+
+describe("Node operations", () => {
```
snario: feels unnecessarily under-descriptive

packages/node/test/unit/node.spec.ts
```diff
@@ -1,12 +1,61 @@
...
+  const privateKey = process.env.npm_package_config_unlockedAccount0!;
```
snario: Should abstract all environment variables to a single file `env.ts`

snario: and export them as all-caps variables

packages/node/test/unit/node.spec.ts
```diff
@@ -1,12 +1,61 @@
...
+  const provider = ethers.getDefaultProvider() as ethers.providers.Web3Provider;
```
snario: I wouldn't use the default provider; that's the production etherscan geth instance I think. should assume localhost for tests

packages/node/test/unit/node.spec.ts
```diff
@@ -1,12 +1,61 @@
...
+  let firestore: firebase.firestore.Firestore;
+  let networkContext: Map<string, Address>;
```
snario: we have a type for this I thought?

packages/node/src/node-types.ts
mvanderh: That's my task to move all common types to typescript-typings. Will be my next goal after #288

cf19drofxots:
@snario addressed comments

package.json
```diff
@@ -34,7 +40,7 @@
     "off chain"
   ],
   "devDependencies": {
-    "ganache-cli": "^6.1.8",
+    "ganache-cli": "6.1.8",
```
cf19drofxots: The reason for this was noted in the commit message: https://stackoverflow.com/a/53413796/10047840

PR #285, opened 2018-11-29
[playground] Added support for opening a dApp

joelalejandro:
This PR expands on previous work done in #277 and #278. A dApp icon is now clickable, transitioning to a route `/dapp/<app-slug>`, where the `<dapp-container />` component is instantiated, opening the dApp's URL.

### DIFF
--- a/packages/playground/src/utils/app-list.ts
+++ b/packages/playground/src/utils/app-list.ts
@@ -1,0 +1,17 @@
+import { AppDefinition } from "../types";
+
+export default {
+  // TODO: How do we get a list of available apps?
+  "0x822c045f6F5e7E8090eA820E24A5f327C4E62c96": {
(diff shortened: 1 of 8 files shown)


packages/playground/src/utils/app-list.ts
```diff
@@ -0,0 +1,17 @@
+import { AppDefinition } from "../types";
+
+export default {
+  // TODO: How do we get a list of available apps?
+  "0x822c045f6F5e7E8090eA820E24A5f327C4E62c96": {
```
snario: How is the hash computed?

joelalejandro: @snario This is just mocked data, but it's the address where the dapp's smart contract is deployed.

PR #288, opened 2018-11-29
[CFjs] Support event listeners: rejectInstall, updateState

mvanderh:
### Description

- Extensive types for Provider-specific event types and data
- Support listening for events on Provider object
- Add `getOrCreateAppInstance` method that gets or creates a singleton AppInstance object given its unique ID.
- Support rejectInstall event as a first ditch effort

+ tests

**Depends on #284**


packages/cf.js/test/provider.spec.ts
```diff
@@ -173,4 +174,54 @@ describe("CF.js Provider", async () => {
...
+        oldState: "4",
+        action: "-1",
+        newState: "3"
+      }
+    });
+    expect(nodeProvider.postedMessages).toHaveLength(1);
+    const detailsRequest = nodeProvider.postedMessages[0] as Node.MethodRequest;
+    expect(detailsRequest.type).toBe(Node.MethodName.GET_APP_INSTANCE_DETAILS);
+    expect(
+      (detailsRequest.params as Node.GetAppInstanceDetailsParams).appInstanceId
```
snario: Is this typecast necessary?

packages/cf.js/test/provider.spec.ts
```diff
@@ -173,4 +174,54 @@ describe("CF.js Provider", async () => {
...
+    ).toBe(TEST_APP_INSTANCE_INFO.id);
+    nodeProvider.simulateMessageFromNode({
+      type: Node.MethodName.GET_APP_INSTANCE_DETAILS,
```
snario: maybe rename `type` to `method`?

packages/cf.js/test/provider.spec.ts
```diff
@@ -173,4 +174,54 @@ describe("CF.js Provider", async () => {
     nodeProvider.simulateMessageFromNode(msg);
     nodeProvider.simulateMessageFromNode(msg);
   });
+
+  it("should load app instance details on-demand", async () => {
```
snario: This test, overall, seems really long. Can you break it up into several smaller tests?

packages/cf.js/test/provider.spec.ts
mvanderh: It is

packages/cf.js/test/provider.spec.ts
mvanderh: I don't think so :( Testing on-demand loading of details requires all of these steps. 

What I could do is make the test more readable by extracting methods

packages/cf.js/src/provider.ts
```diff
@@ -30,8 +33,12 @@ export class Provider {
     return result.appInstances.map(info => new AppInstance(info));
   }
 
-  on(eventName: DappEventType, callback: (e: DappEvent) => void) {
-    // TODO: support notification observers
+  on(eventName: EventType, callback: (e: CounterfactualEvent) => void) {
+    this.eventEmitter.on(eventName, callback);
+  }
+
+  once(eventName: EventType, callback: (e: CounterfactualEvent) => void) {
```
cf19drofxots: What's the common use case for `once`?

I see it's being used in the tests for reject install, but wouldn't you want to constantly listen for reject installs and not just once?

mvanderh: *shrug* no idea, but can't hurt to expose a common convenience method.

packages/cf.js/src/provider.ts
```diff
@@ -40,43 +47,150 @@ export class Provider {
...
     return new Promise<Node.MethodResponse>((resolve, reject) => {
+      const request: Node.MethodRequest = {
+        requestId,
+        params,
+        type: methodName
+      };
       this.requestListeners[requestId] = response => {
         if (response.type === Node.ErrorType.ERROR) {
-          return reject(response.data);
+          return reject({
```
cf19drofxots: Is there a `MethodReject` interface that the `reject` param can be typed to?

mvanderh: No. There might be once I think more about the types of errors that methods can produce

packages/cf.js/src/provider.ts
```diff
@@ -40,43 +47,150 @@ export class Provider {
...
 
   private onNodeMessage(message: Node.Message) {
-    const requestId = (message as Node.MethodResponse).requestId;
-    if (requestId) {
-      if (this.requestListeners[requestId]) {
-        this.requestListeners[requestId](message);
-        delete this.requestListeners[requestId];
-      }
+    const type = message.type;
+    if (Object.values(Node.ErrorType).indexOf(type) !== -1) {
```
cf19drofxots: I wish there was a utility method to wrap this purely for syntactic sugar.

Though this doesn't seem like a common enough operation (yet?) to warrant that method probably.

mvanderh: I did that, but it turned out to be more confusing than it's worth.

packages/cf.js/src/provider.ts
```diff
@@ -40,43 +47,150 @@ export class Provider {
...
+      this.appInstances[id] = new AppInstance(newInfo);
+    }
+    return this.appInstances[id];
+  }
+
+  private async handleNodeEvent(nodeEvent: Node.Event) {
+    let event: CounterfactualEvent;
+    switch (nodeEvent.type) {
+      case Node.EventName.REJECT_INSTALL: {
+        const data = nodeEvent.data as Node.RejectInstallEventData;
```
cf19drofxots: Is type casting here equivalent to
```
const data: Node.RejectInstallEventData = nodeEvent.data
```
?

mvanderh: No actually, I tried that but Typescript complained.

packages/cf.js/test/provider.spec.ts
```diff
@@ -22,6 +30,16 @@ class TestNodeProvider implements INodeProvider {
   }
 }
 
+const TEST_APP_INSTANCE_INFO: AppInstanceInfo = {
```
cf19drofxots: This seems like a generic dummy that could be used for other tests. Might be worth putting it in a `common.ts` or somewhere.

PR #298, opened 2018-12-02
[node] Nodes can communicate

cf19drofxots:
### Description

Given a messaging service, Nodes can communicate with other Nodes.

https://github.com/counterfactual/monorepo/blob/6594bfc6142a65fed9900af45748cb415fa29de9/packages/node/test/integration/node-communication.spec.ts#L51-L68


packages/node/src/node.ts
```diff
@@ -60,6 +102,34 @@ export default class Node {
     });
   }
 
+  /**
+   * When a Node is first instantiated, it establishes a connection
+   * with firebase.
+   * @param firebase
+   */
+  private registerConnection(firebase: firebase.database.Database) {
+    if (!this.firebase.app) {
```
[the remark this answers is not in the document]
cf19drofxots: Seems like it, as the call to initialize it is `initializeApp`, so if there's no `firebase.app.App` member defined here then presumably it indicates that it wasn't initialized.

packages/node/package.json
```diff
@@ -13,19 +13,26 @@
...
+    "firebase-server": "^1.0.1",
     "rollup": "^0.67.0",
     "rollup-plugin-typescript2": "^0.18.0",
     "tslint": "^5.11.0",
     "typescript": "^3.1.2"
   },
   "dependencies": {
     "@counterfactual/cf.js": "0.0.1",
-    "eventemitter3": "^3.1.0"
+    "debug": "^4.1.0",
```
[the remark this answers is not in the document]
cf19drofxots: This package's dependency tree is very small (only has 1 dependency which has no dependencies).

But in spite of that, removing it is probably the better decision if it's not strictly needed.

packages/node/src/node.ts
```diff
@@ -1,35 +1,58 @@
...
    * Because the Node receives and sends out messages based on Event type
    * https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/src/types/node-protocol.ts#L21-L33
    * the same EventEmitter can't be used since response messages would get
    * sent to listeners expecting request messages.
    **/
-  private incoming: EventEmitter;
-  private outgoing: EventEmitter;
+  private readonly incoming: EventEmitter;
+  private readonly outgoing: EventEmitter;
+  private readonly signer: ethers.utils.SigningKey;
```
mvanderh: You're currently not signing anything with this key, only using its public address. Is that going to change in the near future?

packages/node/src/node.ts
```diff
@@ -42,6 +65,23 @@ export default class Node {
...
+   * Sends a message to another Node. It also auto-includes the from field
+   * in the message.
+   * @param peerAddress The peer to whom the message is being sent.
+   * @param msg The message that is being sent.
+   */
+  async send(peerAddress: Address, msg: object) {
+    const modifiedMsg = Object.assign({ from: this.address }, msg);
+    await this.messagingService
+      .ref(`${MESSAGES}/${peerAddress}`)
+      .set(modifiedMsg);
```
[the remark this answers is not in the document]
cf19drofxots: Probably will be required soon, this is naive for now.

packages/node/src/node.ts
cf19drofxots: Yes. Embedding the machine will effectively make more use of this.

packages/node/src/node.ts
```diff
@@ -42,6 +69,21 @@ export default class Node {
...
+  }
+
+  /**
+   * Sends a message to another Node. It also auto-includes the from field
+   * in the message.
+   * @param peerAddress The peer to whom the message is being sent.
+   * @param msg The message that is being sent.
+   */
+  async send(peerAddress: Address, msg: object) {
+    const modifiedMsg = Object.assign({ from: this.address }, msg);
```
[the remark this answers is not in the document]
snario: We don't care about the browser compatibility because we're compiling this code and using rollup which will probably just use `Object.assign` anyway. So it's purely a code cleanliness this from my point of view.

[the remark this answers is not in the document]
cf19drofxots: The syntactic difference is pretty negligible here IMO, but I'm mostly indifferent.

cf19drofxots:
@snario @mvanderh addressed comments & made some updates (primarily wrt to including a messaging service interface)

cf19drofxots:
@snario anything blocking this merge?

build.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin
-packages="contracts common-types machine cf.js node node-provider playground dapp-high-roller"
+packages="contracts common-types cf.js machine node node-provider playground dapp-high-roller"
+
+set -e
```
snario: usually `set -e` goes at top of file before any variables etc

joelalejandro: @snario Any particular reason?

cf19drofxots: The default behavior of shell scripts is to ignore any errors anywhere in the execution pipeline.

Putting `set -e` explicitly says to abort if any failure happens, so putting it at the top ensures anything that comes after it is covered.

joelalejandro: :+1: TIL

PR #300, opened 2018-12-03
[machine] Use return instead of next() for middleware progression

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR removes the usage of `next()` inside the `machine`. The idea is that since no explicit use case requires it yet, we shouldn't need to add this functionality. Counterpoints exist, such as maintaining similar conventions to other JS middleware-style implementations like Koa.


ebryn:
Why?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Because I do not know what machine feature this supports.

PR #311, opened 2018-12-04
[common-types] Move common types to their own package

mvanderh:
### Description

Some types are used by multiple Counterfactual packages:
- Node Protocol messages
- Data Types (AppInstanceInfo etc)
- Simple type aliases such as Address, Bytes32

These types are moved to a new package called common-types.

**Question**: Are there other common types that should be moved?


packages/node/src/node.ts
```diff
@@ -1,12 +1,8 @@
-import EventEmitter from "eventemitter3";
-
 import {
   AppInstanceInfo,
-  GetAppInstancesResult,
-  MethodName,
-  MethodRequest,
-  MethodResponse
-} from "./node-types";
+  Node as NodeTypes
```
mvanderh: @cf19drofxots Couldn't really come up with an elegent name for this. Can't call it "Node" because that would conflict with the Node class. Any suggestions?

packages/common-types/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@counterfactual/common-types",
```
cf19drofxots: Why not just `/types`?

packages/common-types/rollup.config.js
```diff
@@ -0,0 +1,18 @@
+import typescript from "rollup-plugin-typescript2";
```
cf19drofxots: Does this really need to be rolled up?

Only functionality we want from this package is effectively the `.d.ts` files, which are built with the tsc compiler, no?

packages/common-types/package.json
mvanderh: common-types is  more descriptive. "The types that everyone has in common"

packages/common-types/rollup.config.js
mvanderh: Yup, it breaks without it. Believe me i didn't want to.

mvanderh: Specifically the jest module resolver breaks.

Also we need it regardless because common-types has to expose enum values

packages/node/src/node.ts
mvanderh: Mkay

packages/common-types/package.json
cf19drofxots: Yeah I get that, it's that generally the commonality is implied given the namespace in which a generic package such as `types` would be placed in.

So adding `common` seems slightly redundant, but I digress.

packages/common-types/rollup.config.js
cf19drofxots: Hmm alright

cf19drofxots:
> Question: Are there other common types that should be moved?

More PRs can be made when more common types emerge

packages/cf.js/package.json
```diff
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@counterfactual/contracts": "0.0.2",
+    "@counterfactual/common-types": "0.0.1",
```
snario: Can we just call it `types`?

packages/common-types/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.0.1",
+  "description": "Types that other Counterfactual packages share in common",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "license": "MIT",
+  "private": true,
+  "scripts": {
+    "clean": "rm -rf dist",
+    "build": "tsc -b && rollup -c",
```
snario: I don't think we need rollup for this.

packages/common-types/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "sourceRoot": ".",
+    "declarationDir": "dist/",
```
snario: Why are you using `declarationDir`

packages/common-types/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "sourceRoot": ".",
```
snario: Isn't `sourceRoot` supposed to be `src`?

packages/common-types/package.json
mvanderh: See my earlier convo with @cf19drofxots

packages/cf.js/package.json
mvanderh: Why?

PR #314, opened 2018-12-05
[node] Nodes can use store service

cf19drofxots:
### Description

Given a store service, Nodes can store and retrieve data.

https://github.com/counterfactual/monorepo/blob/dc9e1e816bd70b67fa7d3588a3e725a237bf39d5/packages/node/test/integration/node-store.spec.ts#L31-L39


packages/node/test/integration/services/firebase-service.ts
```diff
@@ -0,0 +1,91 @@
...
+import firebase from "firebase";
+import FirebaseServer from "firebase-server";
+
+import {
+  IMessagingService,
+  IStoreService
+} from "../../../src/service-interfaces";
+
+const MESSAGING_SERVER_KEY = "messages";
+const STORE_SERVER_KEY = "store";
```
snario: Should pull these values from the environment and be different per deployment

packages/node/test/integration/services/firebase-service.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!this.firebase.app) {
+      console.error(
+        "Cannot register a connection with an uninitialized firebase handle"
+      );
+      return;
+    }
+
+    this.firebase
+      .ref(`${MESSAGING_SERVER_KEY}/${address}`)
+      // The snapshot being sent to this call _might_ be null
```
snario: why? this comment adds no information different from `| null` below

packages/node/test/integration/services/firebase-service.ts
```diff
@@ -0,0 +1,91 @@
...
+      .on("value", (snapshot: firebase.database.DataSnapshot | null) => {
+        if (snapshot === null) {
+          console.debug(
+            `Node with address ${address} received a "null" snapshot`
+          );
+          return;
+        }
+        const msg = snapshot.val();
+        callback(msg);
```
snario:
```suggestion
        callback(snapshot.val());
```

packages/node/test/integration/services/firebase-service.ts
```diff
@@ -0,0 +1,91 @@
...
+      });
+    return result;
+  }
+
+  async set(key: string, value: any): Promise<any> {
+    return await this.firebase.ref(`${STORE_SERVER_KEY}/${key}`).set(value);
+  }
+}
+
+const HOST = "localhost";
```
snario: Again store these somewhere else

packages/node/src/node.ts
```diff
@@ -75,6 +76,27 @@ export default class Node {
...
 
+  // Note: The following getter/setter method will become private
+  // once the machine becomes embedded in the Node.
+  /**
+   * Retrieves the value that's mapped to the given key through the provided
+   * When the given key doesn't map to any value, `null` is returned.
+   * store service.
+   * @param key
+   */
+  async get(key: string): Promise<any> {
```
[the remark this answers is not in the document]
cf19drofxots: The functionality that primarily uses the store service is the machine storing commitments, which is why this PR is a prerequisite for embedding the machine.

packages/node/test/integration/services/firebase-service.ts
cf19drofxots: This is specifically under the test scope though, and so it'll be the same across deployments.

For services that are actually injected into the Node, those should probably pull this stuff from the environment.

PR #312, opened 2018-12-04
[dapp-high-roller] Integrate with Playground + CF.js + NodeProvider

joelalejandro:
This PR allows to execute the handshake protocol needed to connect the Playground to the dApp via MessageChannel.

⚠️  Symlinks are required for this implementation to work. Since we have some issues with Rollup imports/exports, I'm using IIFE versions to make things work:

```
cf.js -> ../../../cf.js/dist/index-iife.js
ethers.js -> ../../../../node_modules/ethers/dist/ethers.js
eventemitter3.js -> ../../../node-provider/node_modules/eventemitter3/umd/eventemitter3.js
node-provider.js -> ../../../node-provider/dist/index-iife.js
```

PR #316, opened 2018-12-05
[monorepo] Adds arbitrary order build script

joelalejandro:
According to findings on #312, while we work on IIFE mode, we need to use symlinks to connect the dependencies. To keep `lerna` from breaking the build, we now use a bash script to arbitrarily build the packages in the order we need them to.

:warning: This should be reverted as soon as we fix Rollup.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,5 +5,5 @@ for package in $packages; do
   echo ">>> Building package: $package"
   cd packages/$package
   yarn build
-  cd ../..
+  cd -
 done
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin
+packages="contracts common-types machine cf.js node node-provider playground dapp-high-roller"
+
+for package in $packages; do
+  echo ">>> Building package: $package"
+  cd packages/$package
+  yarn build
+  cd ../..
```
cf19drofxots: Protip: you can also do `cd -` here to jump back

joelalejandro: Good one!

PR #321, opened 2018-12-07
[playground] Implements account and network tunnels for state

patience-tema-baron:
Adds an `AccountTunnel` and `NetworkTunnel` to keep track of app state. Currently breaks tests due to https://github.com/ionic-team/stencil-state-tunnel/issues/4.


.merge_file_LmODxR
```diff
@@ -0,0 +1,43 @@
+import { newE2EPage } from "@stencil/core/testing";
```
snario: Don't think this file should be committed

packages/dapp-high-roller/src/components/app-wager/app-wager.e2e.ts.orig
```diff
@@ -0,0 +1,11 @@
+import { newE2EPage } from "@stencil/core/testing";
```
snario: Don't think should be committed

packages/playground/src/components/account/account-edit/account-edit.tsx
```diff
@@ -9,6 +10,11 @@ import { UserChangeset } from "../../../types";
   shadow: true
 })
 export class AccountEdit {
+  @Element() el!: HTMLStencilElement;
+  @Prop() address: string = "";
+  @Prop() email: string = "";
```
snario: Why is there an email field

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -1,33 +1,53 @@
...
+import NetworkTunnel, { State as NetworkState } from "../../data/network";
+
 @Component({
   tag: "app-root",
   styleUrl: "app-root.scss",
   shadow: true
 })
 export class AppRoot {
+  @State() accountState: AccountState = {};
+  @State() networkState: NetworkState = {
```
snario: What's the representation of a network state supposed to be atm?

packages/playground/src/components/account/account-edit/account-edit.tsx
Alonski: @snario Shouldn't the account edit page allow editing of email? If not then I can remove this

packages/playground/src/components/app-root/app-root.tsx
Alonski: @snario Currently it is hardcoded to be "Lorem".
This will need to be inserted according to the actual network.

packages/playground/src/components/account/account-edit/account-edit.tsx
joelalejandro: @snario @Alonski  Let's just leave this for now.

PR #324, opened 2018-12-10
[CFjs] Propose install of app instances

mvanderh:
### Description

Add a proposeInstall method to AppFactory that returns app instance ID. 

Also:
- add off() method to Provider to unsubscribe from events


packages/cf.js/src/provider.ts
```diff
@@ -87,6 +91,26 @@ export class Provider {
     });
   }
 
+  async getOrCreateAppInstance(
+    id: AppInstanceID,
+    info?: AppInstanceInfo
+  ): Promise<AppInstance> {
+    if (!this.appInstances[id]) {
```
joelalejandro: @mvanderh How do you feel about flipping the condition so the behavior is better reflected in the function's name?

```ts
if (id in this.appInstances) {
  return this.appInstances[id];
}

// Rest of the code is the "createAppInstance" part...
```

packages/cf.js/src/app-instance.ts
```diff
@@ -1,8 +1,28 @@
...
+  AppInstanceID,
+  AppInstanceInfo,
+  BlockchainAsset
+} from "@counterfactual/common-types";
+import { BigNumber } from "ethers/utils";
 
 export class AppInstance {
   readonly id: AppInstanceID;
-  constructor(readonly info: AppInstanceInfo) {
+  readonly appId: Address;
```
cf19drofxots: So `initialState` is not being set here?

packages/cf.js/src/provider.ts
mvanderh: I don't think that works in Javascript? Or is it new?

packages/cf.js/src/app-instance.ts
cf19drofxots: So the idea here is to not include the state in an `AppInstance` and call the Node for that AppInstance's state, hence why it's in the proposal params?

packages/cf.js/src/provider.ts
joelalejandro: @mvanderh It's part of [ES5](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in), it should work both in Node and the browser:

image

image

packages/cf.js/src/app-instance.ts
mvanderh: Good point.

For starters, I don't reckon any dapp would need access to the *initial* state. However, I can imagine wanting immediate access to the latest state, instead of asking for it asynchronously every time you need it (or subscribing to state update events).

@ebryn @joelalejandro thoughts?

packages/cf.js/src/provider.ts
mvanderh: Oooh cool, TIL. I like your suggestion then

packages/cf.js/src/app-instance.ts
joelalejandro: I believe it makes sense to have the latest state ready as soon as possible.

mvanderh: Do you need it on-demand or could it be loaded async? I'm just a bit afraid of race conditions if you rely on a `state` field being up-to-date at all times.

mvanderh: @joelalejandro

joelalejandro: It can be `async` @mvanderh

packages/cf.js/src/provider.ts
joelalejandro: I actually meant to reverse the code blocks, not only to apply the `in` operator :P But don't worry, let's merge this PR and deal with that later :)

PR #329, opened 2018-12-11
[node] Propose install & install across Nodes

cf19drofxots:
### Description

This adds support in the Node to make install proposal and install calls across Nodes. 

https://github.com/counterfactual/monorepo/blob/4255016971f27799c681e74678e9970c93a33779/packages/node/test/integration/propose-install.spec.ts#L61-L119


packages/node/src/channels.ts
```diff
@@ -0,0 +1,328 @@
...
+  static initialFreeBalances(
+    multisigOwners: Address[],
+    initialAppsNonce: Nonce
+  ): {
+    [assetType: number]: FreeBalance;
+  } {
+    // TODO: extend to all asset types
+    const ethFreeBalance = new FreeBalance(
+      multisigOwners[0],
+      ethers.utils.bigNumberify("0"),
```
joelalejandro: @cf19drofxots Might be interesting to have a constant for this. I've noticed we use it several times.

packages/node/src/channels.ts
```diff
@@ -0,0 +1,328 @@
...
+
+  private async getChannelFromPeerAddress(
+    peerAddress: Address
+  ): Promise<Channel> {
+    const owners = Channels.canonicalizeAddresses([
+      this.selfAddress,
+      peerAddress
+    ]);
+    const multisigAddress = this.ownersToMultisigAddress[owners];
+    const channel = await this.store.get(
```
joelalejandro: @cf19drofxots Should we maybe abstract this into a `getChannelFromStore()` function?

packages/node/src/channels.ts
```diff
@@ -0,0 +1,328 @@
...
+      channel.freeBalances,
+      channel.appInstances,
+      channel.proposedAppInstances
+    );
+  }
+
+  /**
+   * Gets all appInstances across all of the channels open on this Node.
+   */
+  async getAllApps(): Promise<AppInstanceInfo[]> {
```
joelalejandro: @cf19drofxots We could probably have a private function that solves most of the logic and swap the key to inspect in the `channel` object, something like:

```ts
private async getAppsFrom(appsListKey: string) {
    const apps: AppInstanceInfo[] = [];
    const channels = await this.getAllChannels();
    Object.values(channels).forEach((channel: Channel) => {
      if (channel[appsListKey]) {
        apps.push(...Object.values(channel[appsListKey]));
      }
    });
    return apps;
}
```

And then the public API is reduced to:

```ts
async getAllApps(): Promise<AppInstanceInfo[]> {
  return this.getAppsFrom("appInstances");
}

async getAllPendingApps(): Promise<AppInstanceInfo[]> {
  return this.getAppsFrom("proposedAppInstances");
}
```

What do you think?

packages/node/src/channels.ts
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  private async getChannelFromAppInstanceId(
+    appInstanceId: string
+  ): Promise<Channel> {
+    const multisigAddress = this.appInstanceIdToMultisigAddress[appInstanceId];
+    const channel = await this.store.get(
+      `${this.multisigKeyPrefix}/${multisigAddress}`
+    );
+    return new Channel(
```
joelalejandro: @cf19drofxots Does TypeScript takes this type and automatically wraps it into a Promise?

packages/node/src/methods/method-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { Channels } from "../channels";
+import { NodeMessage } from "../node";
+import { IMessagingService } from "../service-interfaces";
+
+import { addAppInstance, install, proposeInstall } from "./installations";
+import { addMultisig, createMultisig } from "./multisig-creation";
+
+export class MethodHandler {
+  private METHODS = new Map();
```
joelalejandro: @cf19drofxots Any particular reason for using upper-case here?

packages/node/src/methods/method-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+    this.mapEvents();
+    this.EVENTS.forEach((eventHandler: Function, eventName: string) => {
+      this.outgoing.on(eventName, async (msg: NodeMessage) => {
+        await eventHandler(this.channels, this.messagingService, msg);
+      });
+    });
+  }
+}
+
+async function getAppInstances(
```
joelalejandro: @cf19drofxots Should this function live elsewhere?

packages/node/src/methods/multisig-creation.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function createMultisig(
+  channels: Channels,
+  messagingService: IMessagingService,
+  params: Node.CreateMultisigParams
+): Promise<Node.CreateMultisigResult> {
+  const result = {
+    multisigAddress: await channels.createMultisig(params)
+  };
+  const selfAddress = channels.selfAddress;
+  const peerAddress = params.owners.filter(owner => owner !== selfAddress)[0];
```
joelalejandro: @cf19drofxots You could also do:

```ts
const [ peerAddress ] = params.owners.filter(...);
```

Looks _prettier_ :P

packages/node/src/node.ts
```diff
@@ -32,13 +32,25 @@ export default class Node {
...
-    this.registerListeners();
-    this.registerConnection();
+    this.channels = new Channels(
+      this.signer.address,
+      this.storeService,
+      // naive, account-based multisig indexing
+      `${nodeConfig.MULTISIG_KEY_PREFIX}/${this.signer.address}`
+    );
+    this.registerMessagingConnection();
+    new MethodHandler(
```
joelalejandro: @cf19drofxots This is a bit weird. Since all of the `MethodHandler` class members are private, maybe it should be a function instead of a class?

packages/node/src/service-interfaces.ts
```diff
@@ -1,8 +1,8 @@
 import { Address } from "@counterfactual/common-types";
 
 export interface IMessagingService {
-  send(peerAddress: Address, msg: object);
-  receive(address: Address, callback: (msg: object) => void);
+  send(peerAddress: Address, msg: any);
```
joelalejandro: @cf19drofxots Why do we go with `any` here? I think it'd be nice if  `msg` had a well-defined type here.

packages/node/test/integration/services/firebase-service.ts
```diff
@@ -36,7 +40,16 @@ class FirebaseMessagingService implements IMessagingService {
           );
           return;
         }
-        callback(snapshot.val());
+        const msg = snapshot.val();
+        const msgKey = JSON.stringify(msg);
+        if (msg !== null) {
```
joelalejandro: @cf19drofxots We can avoid if-nesting by flipping the condition and using `return`:

```ts
// Case 1: Message is null, don't do anything.
if (msg === null) {
  return;
}

// Case 2: Message is already served, delete it.
if (msgKey in this.servedMessages) {
  delete this.servedMessages[msgKey];
  return;
}

// Case 3: Message must be served, do it.
this.servedMessages[msgKey] = true;
callback(msg);
```

packages/node/src/channels.ts
cf19drofxots: Yeah makes sense. I hadn't gotten to retrieving the proposed app instances yet, so didn't need to generalize it just yet.

packages/node/src/channels.ts
cf19drofxots: Yup

packages/node/src/methods/method-handler.ts
cf19drofxots: Wasn't a dynamic map at first. Casing is stale, will update.

packages/node/src/methods/method-handler.ts
cf19drofxots: Yup, moved.

packages/node/src/node.ts
cf19drofxots: This is also stale :P

Not all its method are private now.

packages/node/src/service-interfaces.ts
cf19drofxots: I left it loose here to not restrict usage and wanted to see what the use case pattern would be more concretely.

Will update to a more type-safe signature.

packages/node/test/integration/services/firebase-service.ts
cf19drofxots: Good call!

packages/common-types/src/node-protocol.ts
```diff
@@ -20,6 +20,7 @@ export namespace Node {
   // SOURCE: https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#public-methods
   export enum MethodName {
     GET_APP_INSTANCES = "getAppInstances",
+    GET_PROPOSED_APP_INSTANCES = "getProposedAppInstances",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: does this PR need to add this to the list of public methods? I don't see it there currently

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * The schema of a channel is below.
+ * The following Channels class encapsulates the state and persistence of all
+ * the channels in a Node.
+ */
+
+/**
+ * The fully expanded schema for a channel:
+ * multisigAddress: {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I assume this isn't a literal string - does this mean "keyed by multisigAddress"? what is a schema - is it a firebase concept?

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+ *        assetType: AssetType,
+ *        token?: Address
+ *      },
+ *      deposits: Map<Address, BigNumber>
+ *    }
+ *  },
+ *  proposedAppInstances: same schema as appInstances,
+ *  freeBalances: Map<AssetType,
+ *    freeBalance: {
+ *      alice: Address,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this requires checking if it breaks anything legacy-ish depending on it, but I would like to make the `{alice, bob}.*` parameters private so that people don't accidentally access them

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+ *      aliceBalance: BigNumber,
+ *      bob: Address,
+ *      bobBalance: BigNumber,
+ *      uniqueId: number,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: we should not use this name

packages/node/src/channels.ts
cf19drofxots: Yes, keyed by multisig address.

What is the object representation encapsulating the concept of a state channel?

packages/common-types/src/node-protocol.ts
cf19drofxots: I didn't see it there, so I added it.

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+ * The following Channels class encapsulates the state and persistence of all
+ * the channels in a Node.
+ */
+
+/**
+ * The fully expanded schema for a channel:
+ * multisigAddress: {
+ *  multisigAddress: Address,
+ *  multisigOwners: Address[],
+ *  appsNonce: Nonce,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: what is `appsNonce`?

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+ *  freeBalances: Map<AssetType,
+ *    freeBalance: {
+ *      alice: Address,
+ *      aliceBalance: BigNumber,
+ *      bob: Address,
+ *      bobBalance: BigNumber,
+ *      uniqueId: number,
+ *      localNonce: number,
+ *      timeout: number,
+ *      dependencyNonce: {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: we should not use this name - it should be called `uninstallationNonce`

packages/node/src/channels.ts
cf19drofxots: sgtm

This current schema is rather loosely defined tbh. Whatever schema ends up being the finalized definition that the machine understands can be reflected here.

packages/node/src/channels.ts
cf19drofxots: Should be more accurately named root nonce

Not sure why @snario had a preference to put `app` in there

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  static canonicalizeAddresses(addresses: Address[]): string {
+    addresses.sort((addrA: Address, addrB: Address) => {
+      return new ethers.utils.BigNumber(addrA).lt(addrB) ? -1 : 1;
+    });
+    return ethers.utils.hashMessage(addresses.join(""));
+  }
+
+  static getMultisigAddress(owners: Address[]): Address {
+    // TODO: implement this using CREATE2
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: confused about the naming of this. as it stands this function should be called `generateNewMultisigAddress` or something since `get.*` functions should be idempotent.

packages/node/src/channels.ts
cf19drofxots: I'm actually not even sure this field is required.

What is its purpose? If the freeBalance is indexed by asset type, and there are a known number of asset types, keeping track of the nonce of this object is achieved via `localNonce`. So what should `uniqueId` do, if anything?

packages/node/src/channels.ts
cf19drofxots: You are correct.

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+ * purposes of updating and retrieving a channel's state from the store.
+ *
+ * An instance is by itself stateless and effectively reflects the state of an
+ * according channel in the store.
+ */
+class Channel {
+  constructor(
+    readonly multisigAddress: Address,
+    readonly multisigOwners: Address[],
+    readonly appsNonce: Nonce = new Nonce(true, 0, 0),
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: we should not have a default value here and specify it explicitly at constructor call time instead

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+      channel.multisigOwners,
+      channel.appsNonce,
+      channel.freeBalances,
+      channel.appInstances,
+      channel.proposedAppInstances
+    );
+  }
+
+  async getAppInstances(
+    status: APP_INSTANCE_STATUS
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The API docs seems to suggest that `getAppInstances` takes 0 arguments?

packages/node/src/channels.ts
cf19drofxots: How come?

This only exists to load a pre-existing `appsNonce` upon querying the store for that channel. If no pre-existing channel exists then this default seems natural?

packages/node/src/channels.ts
cf19drofxots: It does not.

This is not exposed in the public API, it's internal to the Node to distinguish between proposed and installed apps.

packages/node/test/integration/cross-node-multisig-creation.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+import { IMessagingService, IStoreService, Node, NodeConfig } from "../../src";
+
+import { A_PRIVATE_KEY, B_PRIVATE_KEY } from "../env";
+
+import FirebaseServiceFactory from "./services/firebase-service";
+import { getChannelAddresses, getNewMultisig } from "./utils";
+
+dotenv.config();
+
+describe("Node can create multisig, other owners get notified", () => {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: 👍

packages/node/test/integration/cross-node-multisig-creation.spec.ts
```diff
@@ -0,0 +1,57 @@
+import dotenv from "dotenv";
+import FirebaseServer from "firebase-server";
+
+import { IMessagingService, IStoreService, Node, NodeConfig } from "../../src";
+
+import { A_PRIVATE_KEY, B_PRIVATE_KEY } from "../env";
+
+import FirebaseServiceFactory from "./services/firebase-service";
+import { getChannelAddresses, getNewMultisig } from "./utils";
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: AFAICT there are two different things named `getChannelAddresses`; can we disambiguate them

packages/common-types/src/node-protocol.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: same comment for CREATE_MULTISIG and GET_CHANNEL_ADDRESSES

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `cf.legacy.utils.FreeBalance.uniqueId` is actually not something I'm very familiar with. It seems to be unused.

packages/node/test/integration/cross-node-multisig-creation.spec.ts
cf19drofxots: What are the two different things?

Are you referring to these two:
- https://github.com/counterfactual/monorepo/pull/329/files#diff-aee0f6276c67926a5915f7fb7f49a8dcR53
- https://github.com/counterfactual/monorepo/pull/329/files#diff-888623f31f995189e8abe5b3979358e8R33

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: is it a property that you must specify a default (or a default-constructable type, or something) in the store schema?

packages/node/src/channels.ts
cf19drofxots: Thought so. I'll drop it. If we need it (or something similar) later can add it back upon reassessment.

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: ah, this is an internal function not exactly the same as the node method?

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: > What is the object representation encapsulating the concept of a state channel?

Are you answering me in Jeopardy! syntax?

packages/node/src/channels.ts
cf19drofxots: Correct.

All the public methods get mapped here: https://github.com/counterfactual/monorepo/pull/329/files#diff-54ab0b7918c282f97b5fc1d791ddbdd4R56

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I see - yes, agree that this schema (due to transitive dependencies) reflects quite a bit of legacy stuff

packages/common-types/src/node-protocol.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: as in, doesn't this PR need to edit `cf.js/API_REFERENCE.md` to keep it in sync with this change?

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+      `${this.multisigKeyPrefix}/${
+        channel.multisigAddress
+      }/proposedAppInstances`,
+      channel.proposedAppInstances
+    );
+  }
+
+  // Utility methods
+
+  static canonicalizeAddresses(addresses: Address[]): string {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: can we call this `hashUnorderedListOfAddresses` or similar? IMO this name is not really accurate

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  async getAddresses(): Promise<Address[]> {
+    const channels = await this.getAllChannels();
+    return Object.keys(channels);
+  }
+
+  private async getChannelFromPeerAddress(
+    peerAddress: Address
+  ): Promise<Channel> {
+    const owners = Channels.canonicalizeAddresses([
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: can we call this `ownersHash`

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+/**
+ * Note: this class itelf does not hold any meaningful state either.
+ * It encapsulates the operations performed on relevant appInstances and
+ * abstracts the persistence to the store service.
+ */
+export class Channels {
+  /**
+   * A convenience lookup table from a set of owners to multisig address.
+   */
+  private readonly ownersToMultisigAddress = {};
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: can we call this `ownerHashToMultisigAddress`

packages/node/src/channels.ts
cf19drofxots: Not right now.

Though what's the use case for not having a default here when a new channel is being created? What's the benefit of having the caller supply it when they're creating a _new_ channel?

packages/node/src/channels.ts
```diff
@@ -0,0 +1,387 @@
...
+   *        environment.
+   */
+  constructor(
+    public readonly selfAddress: Address,
+    private readonly store: IStoreService,
+    private readonly multisigKeyPrefix: string
+  ) {}
+
+  /**
+   * Called when a new multisig is created for a set of owners.
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: AIUI this is called *to* create a new multisig (i.e. not in response to someone else creating a multisig)

packages/node/src/channels.ts
cf19drofxots: Yeah I don't like this name either. Will update.

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: hmm, can we write "firebase schema" (maybe link to a reference) early on in the comments before abbreviating it to "schema"

packages/node/src/channels.ts
cf19drofxots: Correct.

Does the description throw you off from that interpretation? If so, what would be a better description?

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: but it is very good that this is written out in expanded form somewhere

packages/node/src/channels.ts
cf19drofxots: What makes this inherently tied to firebase?

cf19drofxots: Any key/value store should be able to store an object with this representation, no?

packages/node/src/methods/request-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+  getProposedAppInstances
+} from "./app-instance-operations";
+import { addAppInstance, install, proposeInstall } from "./install-operations";
+import {
+  addMultisig,
+  createMultisig,
+  getChannelAddresses
+} from "./multisig-operations";
+
+export class RequestHandler {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Can we have a docstring for this? This seems pretty tightly coupled to node and I can't tell by reading this code what a request handler is

cf19drofxots: Damn, forgot this. Good catch.

packages/node/src/node.ts
```diff
@@ -61,83 +77,51 @@ export default class Node {
...
-  }
-
-  /**
-   * This sets up all the listeners for the methods the Node is expected to have
-   * as described at https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#node-protocol
-   *
-   * The responses to these calls are the events being listened on
-   * https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#events
+   * Enables making a direct call to the Node for a specific method instead
+   * of registering a callback for it.
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: what does "instead of registering a callback" mean?

packages/node/src/channels.ts
cf19drofxots: Actually yeah, the wording can be better. Will update to:
```
Called to create a new multisig for a set of owners.
```

packages/node/test/integration/cross-node-multisig-creation.spec.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: yes, the wrapper in `utils`

actually, that wrapper is super thin and does "the expected thing", so maybe the same name is fine. in that case a short docstring in `utils` would help

packages/node/src/node.ts
cf19drofxots: As opposed to the `on` method above this which accepts a callback that can't be waited on

Maybe dropping the "instead of" is better:
```
   * Makes a direct call to the Node for a specific method.
```

Since the call is awaitable it'd be implied that the call is async, and does not necessarily need a callback.

packages/node/src/channels.ts
cf19drofxots: Actually, that makes sense for app instances, not so much for free balances.

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The root nonce timeout for a new channel shouldn't be 0, for one

packages/node/src/channels.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It wasn't clear to me whether "schema" is meant to refer to a firebase-specific concept or a more general concept

PR #334, opened 2018-12-12
[WIP] [CFjs] Add AppInstance methods: getState(), takeAction(), uninstall() 

mvanderh:
### Description

Make AppInstance class functional enough to be used in the Playground demo.

PR #335, opened 2018-12-12
[CFjs] Support events on AppInstances

mvanderh:
### Description

Support for events on AppInstances.

~Don't mind the terrible commit history. Will rebase once #334 is merged into master.~

~### UNIT TESTS IN PROGRESS~

Depends on #334


.gitignore
```diff
@@ -23,3 +23,5 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
+
+jest-cache/
```
snario: Let's keep this only in the `.gitignore` files of the individual packages.

packages/cf.js/src/provider.ts
```diff
@@ -214,4 +244,26 @@ export class Provider {
     }
     this.eventEmitter.emit(event.type, event);
```
snario: I would just replace the `break`s with `return this.eventEmitter.emit( ... )`

packages/cf.js/src/types/events.ts
```diff
@@ -30,9 +31,15 @@ export interface UpdateStateEventData extends AppEventData {
   action?: AppAction;
 }
 
+export interface CreateMultisigEventData {
```
snario: I think this should be a `type` not an `interface`

packages/cf.js/src/types/events.ts
```diff
@@ -30,9 +31,15 @@ export interface UpdateStateEventData extends AppEventData {
...
+  owners: Address[];
+  multisigAddress: Address;
+}
+
 export interface ErrorEventData {
   errorName: string;
   message?: string;
+  appInstanceId?: string;
   extra?: { [k: string]: string | number | boolean | object };
```
snario: I think we should enforce these fields to exist and not use `?`

packages/cf.js/test/app-instance.spec.ts
```diff
@@ -100,4 +102,116 @@ describe("CF.js AppInstance", () => {
       await appInstance.uninstall();
     });
   });
+
+  describe("Node events", () => {
+    it("fires update state events", async () => {
+      expect.assertions(4);
+
+      const testOldState = "1000";
```
snario: I think state should be typed as an object

packages/cf.js/test/app-instance.spec.ts
```diff
@@ -100,4 +102,116 @@ describe("CF.js AppInstance", () => {
...
+      };
+      nodeProvider.simulateMessageFromNode(event);
+      nodeProvider.simulateMessageFromNode(event);
+      setTimeout(done, 50);
+    });
+
+    it("can unsubscribe from events", done => {
+      const callback = () => {
+        done.fail();
+      };
```
snario: I think you can replace this with `const callback = done.fail` or just use `done.fail` below

packages/common-types/src/node-protocol.ts
```diff
@@ -43,7 +43,7 @@ export namespace Node {
     UNINSTALL = "uninstallEvent",
     PROPOSE_STATE = "proposeStateEvent",
     REJECT_STATE = "rejectStateEvent",
-    MULTISIG_CREATED = "multisigCreatedEvent"
+    CREATE_MULTISIG = "createMultisigEvent"
```
snario: does this enum _need_ to use strings ?

packages/common-types/src/node-protocol.ts
```diff
@@ -166,12 +166,17 @@ export namespace Node {
     myPayout: BigNumber;
     peerPayout: BigNumber;
   }
+  export interface CreateMultisigEventData {
```
snario: I think this should be a `type`

packages/cf.js/test/app-instance.spec.ts
```diff
@@ -100,4 +102,116 @@ describe("CF.js AppInstance", () => {
...
     });
   });
+
+  describe("Node events", () => {
+    it("fires update state events", async () => {
+      expect.assertions(4);
+
+      const testOldState = "1000";
+      const testAction = "200";
+      const testNewState = "1200";
```
mvanderh: Should be ABI-decoded already i.e. BigNumber

packages/common-types/src/node-protocol.ts
mvanderh: I mean preferably yes. Otherwise the node would have to send integer enum values over the wire, which is much less debuggable

packages/cf.js/src/types/events.ts
mvanderh: We should. For now we don't know what errors we have though. 

I could write a couple and strictly type this for now, until we think more about the errors that the Node can return.

packages/cf.js/src/provider.ts
mvanderh: Ah yea that does read nicer

.gitignore
mvanderh: I had a jest-cache/ folder in my root folder for some reason. Git kept asking me to add it so i gitignored it.

cf19drofxots: Sounds like that folder might have been stale? We don't seem to be running jest from the root so not sure how that could happen currently

packages/cf.js/src/app-instance.ts
```diff
@@ -9,8 +9,16 @@ import {
   Node
 } from "@counterfactual/common-types";
 import { BigNumber } from "ethers/utils";
+import EventEmitter from "eventemitter3";
 
 import { Provider } from "./provider";
+import { CounterfactualEvent } from "./types";
+
+export enum AppInstanceEventType {
```
cf19drofxots: Why not just name it `AppInstanceEvent`?

`AppInstanceEventType.UPDATE_STATE` vs `AppInstanceEvent.UPDATE_STATE`

packages/cf.js/src/types/events.ts
cf19drofxots: What does an `extra` field do?

I'd imagine if the `errorName` was an enum then the necessary information could be conveyed to the client through just the `message` right?

packages/cf.js/test/app-instance.spec.ts
```diff
@@ -100,4 +101,116 @@ describe("CF.js AppInstance", () => {
       await appInstance.uninstall();
     });
   });
+
+  describe("Node events", () => {
+    it("fires update state events", async () => {
+      expect.assertions(4);
+
+      const testOldState = bigNumberify(1000);
```
cf19drofxots: Do you find the prefix `test` helpful in these variable names? I'd think it'd be implied given that this is in the context of a test

Just implicit/explicit comparison thoughts

What do you think about the `expected` prefix?

This would lead to:
```
expect(action).toBe(expectedAction)
```

packages/cf.js/test/app-instance.spec.ts
```diff
@@ -100,4 +101,116 @@ describe("CF.js AppInstance", () => {
...
+      nodeProvider.simulateMessageFromNode(event);
+      nodeProvider.simulateMessageFromNode(event);
+      setTimeout(done, 50);
+    });
+
+    it("can unsubscribe from events", done => {
+      const callback = () => {
+        done.fail();
+      };
+      appInstance.on(AppInstanceEventType.UNINSTALL, callback);
```
cf19drofxots: Why subscribe then unsubscribe? Would it not have the same effect to not call `on` and `off` at all?

packages/cf.js/src/provider.ts
```diff
@@ -169,29 +176,28 @@ export class Provider {
   }
 
   private async handleNodeEvent(nodeEvent: Node.Event) {
```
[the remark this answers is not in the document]
mvanderh: I find that an eventHandlers mapping adds more complication than is worth. I do like extracting handlers to methods though, so I've done that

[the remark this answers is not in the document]
joelalejandro: Fair enough. As long as we don't have a mile-long method, I'm good with it :D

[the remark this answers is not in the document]
snario: I agree that mile long methods suck

cf19drofxots:
@mvanderh none of my comments are blockers and this is already approved, though in general it's ideal to resolve outstanding comments before merging

packages/cf.js/src/app-instance.ts
mvanderh: It's inconsistent and ambiguous. Though I too would like a shorter name for it.

packages/cf.js/test/app-instance.spec.ts
mvanderh: Yes, that's what I'm testing: whether `on` followed by `off` does what it's expected to do.

packages/cf.js/test/app-instance.spec.ts
mvanderh: Yeah that reads nicer, just a bit long.

packages/cf.js/src/app-instance.ts
cf19drofxots: I may be missing some context.
- Where does the ambiguity arise? The enum is defining events directly relevant to an AppInstance
- And inconsistent with what?

packages/cf.js/src/types/events.ts
mvanderh: Maybe, but there could still be extra (structured) data that's important for CFjs. I could remove it & re-add once it becomes clear

packages/cf.js/src/app-instance.ts
mvanderh: Ambiguous - is AppInstanceEvent an enum or the actual event object?
Inconsistent - We have EventType for the type of CounterfactualEvent, but AppInstanceEvent for the type of ... app instance event?

PR #345, opened 2018-12-14
[CFjs] Uninstall method return value

mvanderh:
### Description

@snario brought up the need to discuss this: https://github.com/counterfactual/monorepo/pull/334#discussion_r241628649

What should the "uninstall" method return? Both the Node method and the CFjs method.


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
I think "OK" is fine

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
That should be parameters in the uninstall protocol

mvanderh:
Sure, but how do you imagine the dapp developer using CFjs gets this information?

PR #283, opened 2018-11-29
[dapp-high-roller] High Roller Smart Contract

Alonski:
### Description

Build out High Roller Smart Contract

### Related issues

#271 #242

PR #346, opened 2018-12-15
[contracts lint] Add solium max-len -> 80 and fixed linting

Alonski:
### Description

Currently line width is declared in `.editorconfig` at 80 chars. However, this isn't enforced in Solium linting.
This PR adds `max-len: 80` as a Solium rule.

In addition, this PR fixes all current violations of that rule.

Note that there are various ways to fix line length rules. If you see something that is a bit off and can be written in a better way I would love the feedback.

### Related issues

Fixes lint issues #283


packages/contracts/contracts/lib/StaticCall.sol
```diff
@@ -100,7 +100,9 @@ library StaticCall {
     private
     view
   {
-    require(to.isContract(), "Attempted to make a static call on non-conract address");
+    require(
+      to.isContract(), "Attempted to make a static call on non-conract address"
```
cf19drofxots: `non-contract`

snario:
```suggestion
      to.isContract(), "Attempted to make a static call on non-contract address"
```

PR #352, opened 2018-12-18
bold and emojis in build.sh

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
from https://github.com/counterfactual/monorepo/pull/163/files#diff-0b83f9dedf40d7356e5ca147a077acb4R9


snario:
This is more trouble than it is worth... going to just get rid of it. File is being removed soon anyway.

PR #364, opened 2018-12-25
Replace ethers usage to always import from module

snario:
Across the entire repo.

### DIFF
--- a/packages/node/test/integration/utils.ts
+++ b/packages/node/test/integration/utils.ts
@@ -5,7 +5,7 @@ import {
   Node as NodeTypes
 } from "@counterfactual/common-types";
 import cuid from "cuid";
-import { ethers } from "ethers";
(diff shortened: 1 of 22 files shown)


packages/node/test/integration/utils.ts
```diff
@@ -95,9 +95,9 @@ export function makeProposalRequest(
     asset: {
       assetType: AssetType.ETH
     },
-    myDeposit: ethers.utils.bigNumberify("1"),
-    peerDeposit: ethers.utils.bigNumberify("1"),
-    timeout: ethers.utils.bigNumberify("1"),
+    myDeposit: One,
```
joelalejandro: @snario I'm not particularly sure about this one. I think it's clearer to use `bigNumerify("1")`, but that's just me, take it or leave it :P

PR #365, opened 2018-12-25
[contracts] ethereum-waffle testing

snario:
Uses `ethereum-waffle` to test the `contracts` repo.


snario:
I think this is blocking too https://github.com/EthWorks/Waffle/issues/33

packages/contracts/test/app-registry.spec.ts
```diff
@@ -211,4 +216,81 @@ contract("AppRegistry", (accounts: string[]) => {
...
+    //   status,
+    //   latestSubmitter,
+    //   appStateHash,
+    //   disputeCounter,
+    //   disputeNonce,
+    //   finalizesAt,
+    //   nonce
+    // } = await appRegistry.functions.appStates(appInstance.id);
+
+    // expect(status).to.be.eq(1);
```
mvanderh: Does this not work?

mvanderh:
Does this obviate Ganache? Or am I misunderstanding what Waffle does apart from simpler syntax?

snario:
@mvanderh Truffle does a lot of things: its a testing framework, a migrations tools, and a compilation tool. Waffle is just a testing framework and a compilation tool, and not a migration tool. Waffle is very small as a dependancy and easy to reason about and natively uses `ethers` which makes the code cleaner. Also, it has a `--fastCompile` flag which uses `solc` directly (not `solc-js`) which we can enable on CI after this is merged.

PR #368, opened 2018-12-26
[dapp-high-roller] Clean up tsconfig.json

snario:
~Removes `dom` from the `lib` key on the `tsconfig.json` inside `dapp-high-roller` because it is already declared in the root level `tsconfig.json`:~

Removes some duplicated fields from the root `tsconfig.json`


snario:
Oh, it seems like `lib` actually **overrides** the extended config instead of inheriting from it. So, I should add `dom` back in. However I'm curious why `dapp-high-roller` needs `es2015` and `es2017` isn't good enough @Alonski?

PR #370, opened 2018-12-26
Rename truffle-config.js to truffle.js

snario:
Renamed the file so it is consistent in `machine` and in `contracts` (as of #163)


joelalejandro:
@snario Yeah, since they now defaulted their config file name to `truffle-config.js`, there's no more name conflict :P

snario:
Ah. 

Okay, I renamed the one in `machine` then: 8314c176

PR #373, opened 2018-12-26
[machine] Remove node_modules/.bin/ prefix from npm script

snario:


joelalejandro:
Is there a reason to do this? Doesn't seem like a non-breaking change. Can you clarify a bit, please?

snario:
Ah yeah actually ignore this, I didn't realize we needed to this.

PR #379, opened 2018-12-26
[cf.js] Remove all usage of cf.js/legacy

snario:
There are only usages left:

1. [This test file](https://github.com/counterfactual/monorepo/blob/liam/a-new-era-has-begun/packages/cf.js/test/legacy/app-instance.spec.ts#L3) testing a `generateAbiEncodings` function 

2. The `Node` has a class called `Channel` which uses `legacy.utils.Nonce` and `legacy.utils.FreeBalance` as members but does not actually use any of the methods. Also, this class is going to be removed in a PR coming in the next few days.


mvanderh:
@snario Is `generateAbiEncodings` actually _used_ anywhere besides the test?

snario:
> @snario Is `generateAbiEncodings` actually _used_ anywhere besides the test?

nope

mvanderh:
Just remove it then IMO.

mvanderh:
Also lol @ that branch name

PR #383, opened 2018-12-26
[machine] Add an integration test for InstallCommitment

snario:
Unfortunately it breaks parallelism so I had to add `--runInBand` back. I tried listening to events to uniquely identify transactions being submitted to ensure the callback in the integration tests runs for the correct multisig, but it seems like not all events are returned from `ethers.Contract.prototype.on`'s event emitter.

Also this PR includes two fixes that needed to be there:
- https://github.com/counterfactual/monorepo/pull/381
- https://github.com/counterfactual/monorepo/pull/382

### DIFF
--- a/packages/machine/test/integration/install-then-set-state.spec.ts
+++ b/packages/machine/test/integration/install-then-set-state.spec.ts
@@ -1,6 +1,7 @@
 import AppRegistry from "@counterfactual/contracts/build/contracts/AppRegistry.json";
 import ETHBucket from "@counterfactual/contracts/build/contracts/ETHBucket.json";
 import MinimumViableMultisig from "@counterfactual/contracts/build/contracts/MinimumViableMultisig.json";
+import MultiSend from "@counterfactual/contracts/build/contracts/MultiSend.json";
 import NonceRegistry from "@counterfactual/contracts/build/contracts/NonceRegistry.json";
 import ProxyFactory from "@counterfactual/contracts/build/contracts/ProxyFactory.json";
 import StateChannelTransaction from "@counterfactual/contracts/build/contracts/StateChannelTransaction.json";
@@ -13,12 +14,13 @@ import {
   BigNumber,
   hexlify,
   Interface,
+  parseEther,
   randomBytes,
   SigningKey
 } from "ethers/utils";
 
-import { SetStateCommitment, SetupCommitment } from "../../src/ethereum";
-import { StateChannel } from "../../src/models";
+import { InstallCommitment, SetStateCommitment } from "../../src/ethereum";
+import { AppInstance, StateChannel } from "../../src/models";
 
 // To be honest, 30000 is an arbitrary large number that has never failed
 // to reach the done() call in the test case, not intelligency chosen
@@ -28,12 +30,13 @@ const JEST_TEST_WAIT_TIME = 30000;
 // gas needed, so we hard-code this number to ensure the tx completes
 const CREATE_PROXY_AND_SETUP_GAS = 6e9;
 
-// Similarly, the SetupCommitment is a `delegatecall`, so we estimate
-const SETUP_COMMITMENT_GAS = 6e9;
-
 // The AppRegistry.setState call _could_ be estimated but we haven't
 // written this test to do that yet
 const SETSTATE_COMMITMENT_GAS = 6e9;
+
+// Also we can't estimate the install commitment gas b/c it uses
+// delegatecall for the conditional transaction
+const INSTALL_COMMITMENT_GAS = 6e9;
 
 let networkId: number;
 let provider: JsonRpcProvider;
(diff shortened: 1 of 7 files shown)


packages/machine/test/integration/install-then-set-state.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  dotenv.config();
+
+  // Can use ! because dotenv-safe ensures value is set
+  const host = process.env.DEV_GANACHE_HOST!;
+  const port = process.env.DEV_GANACHE_PORT!;
+  const mnemonic = process.env.DEV_GANACHE_MNEMONIC!;
+
+  function pathForAccount(idx: number) {
+    return `m/44'/60'/0'/0/${idx}`;
+  }
```
snario: This uses the HD Path format to get the equivalent of `accounts[idx]` in Truffle tests. I did this when trying to get parallelism to work (avoiding transaction nonce being out of sync) but ended up not needing to do it.

packages/machine/test/integration/install-then-set-state.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    interface Matchers<R> {
+      toBeEq(expected: BigNumber): BigNumber;
+    }
+  }
+}
+expect.extend({
+  toBeEq(received: BigNumber, argument: BigNumber) {
+    return {
+      pass: received.eq(argument),
+      message: () => `expected ${received} not to be equal to ${argument}`
```
mvanderh: Ooh nice. Ive been using `.toString()` but it sucks

packages/machine/test/integration/install-then-set-state.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+ * and trigger the InstallCommitment on-chain to resolve that app and verify
+ * the balances have been updated on-chain.
+ */
+describe("Scenario: install app, set state, put on-chain", () => {
+  jest.setTimeout(JEST_TEST_WAIT_TIME);
+
+  it("returns the funds the app had locked up", async done => {
+    const signingKeys = [
+      new SigningKey(hexlify(randomBytes(32))),
+      new SigningKey(hexlify(randomBytes(32)))
```
mvanderh: Why randomize? I prefer predefined constants.

packages/machine/test/integration/install-then-set-state.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+} from "ethers/utils";
+
+import {
+  InstallCommitment,
+  SetStateCommitment
+} from "../../src/middleware/protocol-operation";
+import { AppInstance, StateChannel } from "../../src/models";
+
+// To be honest, 30000 is an arbitrary large number that has never failed
+// to reach the done() call in the test case, not intelligency chosen
```
mvanderh: "Intelligency" isn't a word lol.

packages/machine/test/integration/install-then-set-state.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+
+    const users = signingKeys.map(x => x.address);
+
+    const proxyFactory = new Contract(
+      ProxyFactory.networks[networkId].address,
+      ProxyFactory.abi,
+      wallet
+    );
+
+    proxyFactory.on("ProxyCreation", async proxy => {
```
mvanderh: There's no way to await this instead? Callback hell & unnecessary indentation sucks.

packages/machine/test/integration/install-then-set-state.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      let stateChannel = new StateChannel(proxy, users).setupChannel(network);
+      let freeBalanceETH = stateChannel.getFreeBalanceFor(AssetType.ETH);
+
+      const state = {
+        alice: stateChannel.multisigOwners[0],
+        bob: stateChannel.multisigOwners[1],
+        aliceBalance: WeiPerEther,
```
mvanderh: This would be easier to read if you instead wrote:
```suggestion
        aliceBalance: parseEther("1.0"),
        bobBalance: parseEther("1.0")
```

packages/machine/test/integration/install-then-set-state.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+        freeBalanceETH.timeout,
+        app.appSeqNo
+      );
+
+      const installTx = installCommitment.transaction([
+        signingKeys[0].signDigest(installCommitment.hashToSign()),
+        signingKeys[1].signDigest(installCommitment.hashToSign())
+      ]);
+
+      await wallet.sendTransaction({ to: proxy, value: WeiPerEther.mul(2) });
```
mvanderh:
```suggestion
      await wallet.sendTransaction({ to: proxy, value: WeiPerEther.mul(2) });
      await wallet.sendTransaction({ to: proxy, value: parseEther("2") });
```

packages/machine/test/integration/install-then-set-state.spec.ts
snario: Aiming for unique values without hard-coding arbitrary strings. Ideally `randomBytes` would take a "seed" value.

PR #390, opened 2018-12-27
[monorepo] regenerate yarn.lock

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


Alonski:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll Why did you need to regenerate the `yarn.lock`? This will update all dependencies as per their semver definition in `package.json`

If we want to update packages we should IMO be using [yarn upgrade interactive](https://yarnpkg.com/lang/en/docs/cli/upgrade-interactive/)

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
@Alonski I'm trying to rebase #163 which changes yarn.lock. AFAICT we pin packages and use greenkeeper enough that no direct dependencies get upgraded.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
there's a guess here that the intermittent unterminated ifndef errors are caused by node_gyp trying to build downloaded files when the download fails halfway https://github.com/phusion/node-sha3/issues/42#issuecomment-390659874

PR #394, opened 2018-12-27
[playground-server] Implements JSON Web Token protocol + Firebase configuration endpoint

joelalejandro:
This PR implements JWT for the API and also decouples Firebase configuration from the frontend, making it available via a `/firebase` endpoint.

The API client on the `playground` package stores a token provided by the API on the user's local storage. Tokens have an expiration time of 30 seconds (so tokens are requested periodically when expired or invalid for other reasons).

**Points I'd like to have some feedback on:**

- Any ideas on how to restrict the `/token` endpoint so only the Playground uses the API? I was thinking of shortening even more the expiration time (something like 5 seconds or so). Also - does this matter?
- At first I wanted to distinguish a non-authorized token from an expired token, but apparently this isn't possible. Should we tie up any error as a `401 Unauthorized`? The API is currently responding `205 Reset Content` as a way of seeing _"hey, refresh your token!"_.


packages/playground-server/package.json
```diff
@@ -21,27 +21,33 @@
...
   },
   "scripts": {
-    "start": "PLAYGROUND_SERVER_ENV=production netlify-lambda -c webpack.config.js serve src",
-    "build": "PLAYGROUND_SERVER_ENV=production netlify-lambda -c webpack.config.js build src",
-    "dev": "PLAYGROUND_SERVER_ENV=development netlify-lambda -c webpack.config.js serve src",
-    "test": "PLAYGROUND_SERVER_ENV=development jest"
+    "start": "netlify-lambda -c webpack.config.js serve src",
+    "build": "netlify-lambda -c webpack.config.js build src",
+    "dev": "cross-env FIREBASE_API_KEY= FIREBASE_AUTH_DOMAIN= FIREBASE_DATABASE_URL= FIREBASE_PROJECT_ID= FIREBASE_STORAGE_BUCKET= FIREBASE_MESSAGING_SENDER_ID= PLAYGROUND_SERVER_SECRET=foo PLAYGROUND_SERVER_ENV=development netlify-lambda -c webpack.config.js serve src",
+    "test": "cross-env FIREBASE_API_KEY=a FIREBASE_AUTH_DOMAIN=b FIREBASE_DATABASE_URL=c FIREBASE_PROJECT_ID=d FIREBASE_STORAGE_BUCKET=e FIREBASE_MESSAGING_SENDER_ID=f PLAYGROUND_SERVER_SECRET=foo PLAYGROUND_SERVER_ENV=development jest"
```
snario: I feel like these scripts are getting super long. Maybe we can find a way to shorten this? I used `dotenv-safe` in other packages to enforce the existence of the variables in the environment.

packages/playground-server/src/methods/token.ts
```diff
@@ -0,0 +1,8 @@
+import { Context } from "koa";
+
+export default function token() {
+  return async function(ctx: Context, next: () => Promise<any>) {
+    // TODO: This should restrict who can request a JSON Web Token.
```
snario: I'm happy with the suggestion of rate-limiting tokens.

packages/playground-server/src/middlewares/expired-token-control.ts
```diff
@@ -0,0 +1,14 @@
+import { Context } from "koa";
+
+export default function expiredTokenControl() {
+  return async function(ctx: Context, next: () => Promise<any>) {
+    return next().catch(err => {
+      if (401 === err.status) {
```
snario:
```suggestion
       if (err.status === 401) {
```

packages/playground-server/src/middlewares/inject-token.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default function injectToken(jwtSecret: string): ConditionalMiddleware {
+  const middleware = async function(
+    this: ConditionalMiddleware,
+    ctx: Context,
+    next: () => Promise<any>
+  ) {
+    ctx.body = {
+      ...ctx.body,
+      token: sign({ ...ctx.body }, jwtSecret, { expiresIn: "30s" })
```
snario:
```suggestion
      token: sign(ctx.body, jwtSecret, { expiresIn: "30s" })
```

packages/playground-server/tsconfig.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "declaration": false
```
snario: We use this a lot -- should it be in the root `tsconfig.json`?

packages/playground-server/package.json
Alonski: Maybe we can do something like this: https://github.com/kentcdodds/cross-env#usage
```
{
  "scripts": {
    "parentScript": "cross-env GREET=\"Joe\" npm run childScript",
    "childScript": "cross-env-shell \"echo Hello $GREET\""
  }
}
```

packages/playground-server/src/methods/token.ts
joelalejandro: @snario I've reduced the lifespan of a token to 5 seconds in d8d27e7. However, I'm leaving this middleware declared for a bit longer. I think I can add a bit more of security without derailing from the main goal.

packages/playground-server/package.json
joelalejandro: :+1: Sounds interesting. If you don't mind, I'd like to work this out on a different PR and come up with a solution for any packages that required env-vars to work.

ebryn:
@snario it needs to be running to receive the firebase messages

PR #395, opened 2018-12-28
[contracts] kill ganache before starting tests

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Previously, if an existing ganache instance is running, `test.sh`'s invocation of ganache-cli would fail silently and the `kill` call would fail.

Why an existing instance might be running: user ran it in another tab, or (in my case) `ganache-cli` gets reparented to PID 1 for some reason.

### DIFF
--- a/packages/contracts/scripts/test.sh
+++ b/packages/contracts/scripts/test.sh
@@ -8,7 +8,7 @@ function clean {
 
 trap clean INT TERM EXIT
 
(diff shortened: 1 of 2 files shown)


snario:
Can we add a log to tell the user we're doing this? Can be annoying if a script kills your other processes without telling you

PR #396, opened 2018-12-28
[monorepo] rename instances of metachannel

ldct:
same motivation as #384


snario:
I think you want to update this to use `Params` not `Data` @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll

PR #398, opened 2018-12-28
[node] Use updated channel and app instance models in Node

cf19drofxots:
### Description

This brings the underlying Node state representation of channels and app instances up to date with that of the machine.

This also bundles up the machine in the Node distribution.


packages/machine/src/index.ts
```diff
@@ -1,3 +1,4 @@
 import { InstructionExecutor } from "./instruction-executor";
 
 export { InstructionExecutor };
+export * from "./models";
```
ebryn: We need to finish reviewing the patterns used in the Machine's models before we expose them as public API

packages/machine/src/models/app-instance.ts
```diff
@@ -67,7 +67,7 @@ export class AppInstance {
     isVirtualApp: boolean,
     appSeqNo: number,
     rootNonceValue: number,
-    latestState: object,
+    latestState: any,
```
ebryn: why `object` to `any`?

packages/machine/src/models/app-instance.ts
```diff
@@ -105,6 +105,10 @@ export class AppInstance {
     return ret;
   }
 
+  public toJson(): AppInstanceJson {
+    return this.json;
```
ebryn: I meant to inquire with @snario about the pattern of using `this.json` to store the state of the AppInstance. Can you explain why this is preferred over setting them on as instance properties?

packages/machine/src/models/app-instance.ts
snario: Actually this seems wrong, it should be the same as `AppState`: https://github.com/counterfactual/monorepo/blob/master/packages/machine/src/protocol-types-tbd.ts#L16

packages/machine/src/models/app-instance.ts
ebryn: We discussed this on a call

cf19drofxots: Keeping it as is?

snario: For now yes though we have an improvement plan for a future PR, not urgent.

packages/node/src/channels.ts
```diff
@@ -80,9 +50,21 @@ export class Channels {
    */
   async createMultisig(params: Node.CreateMultisigParams): Promise<Address> {
     const multisigAddress = this.generateNewMultisigAddress(params.owners);
-    const channel: Channel = new Channel(multisigAddress, params.owners);
-    const ownersHash = orderedAddressesHash(params.owners);
-    await this.saveChannel(channel, ownersHash);
+    let stateChannel: StateChannel = new StateChannel(
```
snario:
```suggestion
    let stateChannel = new StateChannel(
```

snario: This is also repeated in a few other places. It should be implicit from the return value of `StateChannel`'s constructor.

packages/node/src/channels.ts
```diff
@@ -131,271 +124,104 @@ export class Channels {
...
       proposedAppInstance,
-      clientAppInstanceID
+      appInstanceId
     );
+    return appInstanceId;
   }
 
   async install(params: Node.InstallParams): Promise<AppInstanceInfo> {
     if (!params.appInstanceId) {
       return Promise.reject("No AppInstance ID specified to install");
```
snario:
```suggestion
      return Promise.reject("No AppInstanceID specified to install");
```

packages/node/src/channels.ts
```diff
@@ -131,271 +124,104 @@ export class Channels {
...
-    return await this.store.get(
-      `${this.storeKeyPrefix}/${OWNERS_HASH_TO_MULTISIG_ADDRESS}/${ownersHash}`
+    await this.store.addAppInstanceProposal(
+      channel,
+      proposedAppInstance,
+      appInstanceId
     );
   }
 
   // private utility methods
```
snario:
```suggestion
```

snario: Not helpful

packages/node/src/channels.ts
```diff
@@ -407,46 +233,115 @@ export class Channels {
...
-        );
-      }
-    });
-    return apps;
-  }
+  private createAppInstanceFromAppInstanceInfo(
+    proposedAppInstanceInfo: ProposedAppInstanceInfo,
+    channel: StateChannel
+  ): AppInstance {
+    const appInstanceInfo = { ...proposedAppInstanceInfo };
```
snario: Why not just remove this line and use `proposedAppInstanceInfo` below?

packages/node/src/channels.ts
```diff
@@ -407,46 +233,115 @@ export class Channels {
...
+    appInstances: AppInstance[]
+  ): Promise<AppInstanceInfo[]> {
+    const appInstanceInfos: AppInstanceInfo[] = [];
+    for (const appInstance of appInstances) {
+      const appInstanceId = await this.store.getAppInstanceIDFromAppInstanceIdentityHash(
+        appInstance.identityHash
+      );
+      appInstanceInfos.push(await this.store.getAppInstanceInfo(appInstanceId));
+    }
+    return appInstanceInfos;
```
snario: I think this function could be written more clearly using `Array.prototype.map` on `appInstances`.

packages/node/src/channels.ts
```diff
@@ -407,46 +233,115 @@ export class Channels {
...
+  }
+
+  getNonFreeBalanceAppInstancesJSON(stateChannel: StateChannel): AppInstance[] {
+    const appInstances = stateChannel.appInstances;
+    const nonFreeBalanceAppInstances: AppInstance[] = [];
+    appInstances.forEach((appInstance, appInstanceIdentityHash) => {
```
snario: This could also be written in functional style.

packages/node/src/channels.ts
```diff
@@ -407,46 +233,115 @@ export class Channels {
...
+    ]).functions.getTurnTaker.sighash;
+
+    const isStateTerminalSigHash = new Interface([
+      `function isStateTerminal(${stateEncoding})`
+    ]).functions.isStateTerminal.sighash;
+
+    let applyActionSigHash = "0x00000000";
+    if (appInstanceInfo.abiEncodings.actionEncoding !== undefined) {
+      applyActionSigHash = new Interface([
+        `function applyAction(${stateEncoding}, ${
```
snario: I don't think you need to add the token `function` to the string

packages/node/src/channels.ts
```diff
@@ -407,46 +233,115 @@ export class Channels {
...
+      getTurnTaker: appFunctionSigHashes.getTurnTaker,
+      isStateTerminal: appFunctionSigHashes.isStateTerminal,
+      stateEncoding: appInstanceInfo.abiEncodings.stateEncoding,
+      actionEncoding: appInstanceInfo.abiEncodings.actionEncoding
+    };
+
+    const terms: Terms = {
+      assetType: appInstanceInfo.asset.assetType,
+      limit: appInstanceInfo.myDeposit.add(appInstanceInfo.peerDeposit)
+    };
```
snario: This is not typed correctly... weird that linter did not complain.

```
Property 'token' is missing in type '{ assetType: AssetType; limit: BigNumber; }' but required in type 'Terms'.ts(2741)
```

snario: Oh I see you added `?` to `token` in the type.

I don't think it is a good idea to add `?`.

I would recommend just setting the value to `AddressZero` by default.

packages/node/src/store.ts
```diff
@@ -0,0 +1,269 @@
...
+    appInstanceIdentityHash: string
+  ): Promise<string> {
+    return this.storeService.get(
+      `${
+        this.storeKeyPrefix
+      }/${DB_NAMESPACE_APP_INSTANCE_IDENTITY_HASH_TO_APP_INSTANCE_ID}/${appInstanceIdentityHash}`
+    );
+  }
+
+  // setters
```
snario:
```suggestion
```

packages/types/src/app-instance.ts
```diff
@@ -10,7 +10,7 @@ export enum AssetType {
 export interface Terms {
   assetType: AssetType;
   limit: BigNumber;
-  token: Address;
+  token?: Address;
```
snario:
```suggestion
  token: Address;
```

packages/types/src/app-instance.ts
```diff
@@ -30,12 +30,15 @@ export interface AppIdentity {
   defaultTimeout: number;
 }
 
-export interface AppInterface {
-  addr: string;
+export interface AppFunctionsSigHashes {
```
snario: Rename to `AppInterfaceSighashParameters`?

packages/types/src/app-instance.ts
```diff
@@ -48,8 +51,8 @@ export interface SignedStateHashUpdate {
...
 
 export interface ETHBucketAppState {
-  alice: string;
-  bob: string;
-  aliceBalance: BigNumber;
-  bobBalance: BigNumber;
+  readonly alice: string;
+  readonly bob: string;
+  readonly aliceBalance: BigNumber;
+  readonly bobBalance: BigNumber;
```
snario: Curious what the motivation was for this?

packages/types/src/node-protocol.ts
```diff
@@ -79,6 +79,9 @@ export namespace Node {
     timeout: BigNumber;
     initialState: AppState;
   };
+  export interface InterNodeProposeInstallParams extends ProposeInstallParams {
```
snario: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll curious your thoughts here

packages/machine/src/protocol-types-tbd.ts
```diff
@@ -13,7 +13,7 @@ import { Protocol } from "./types";
 // todo(ldct): top-level arrays are probably illegal since they are not
 // structs...
 export type AppState = {
-  [x: string]: string | number | boolean | AppState | AppStateArray;
+  [x: string]: string | BigNumber | boolean | AppState | AppStateArray;
```
snario: I think it can also be a number. Maybe use `BigNumberish`

packages/machine/src/models/state-channel.ts
```diff
@@ -244,4 +274,39 @@ export class StateChannel {
...
+
+    const channelJSON = {
+      multisigAddress: this.multisigAddress,
+      multisigOwners: this.multisigOwners,
+      appInstances: appInstanceJsons,
+      freeBalanceAppIndexes: Array.from(this.freeBalanceAppIndexes.entries()),
+      monotonicNumInstalledApps: this.monotonicNumInstalledApps
+    };
+    return channelJSON;
+  }
```
snario: I think this entire method would be better written in functional style

packages/machine/src/models/state-channel.ts
```diff
@@ -244,4 +274,39 @@ export class StateChannel {
...
+  static fromJson(json: StateChannelJSON): StateChannel {
+    const appInstances: Map<string, AppInstance> = new Map();
+    if (json.appInstances) {
+      Object.entries(json.appInstances).forEach(appInstanceEntry => {
+        appInstances.set(
+          appInstanceEntry[0],
+          AppInstance.fromJson(appInstanceEntry[1])
+        );
+      });
+    }
```
snario: Can't you do `new Map(json.appInstances)`? or perhaps 

```js
new Map(
  json.appInstances.map(
    ([k, v]) => [k, AppInstance.fromJson(v)]
  )
);
```

packages/types/src/app-instance.ts
cf19drofxots: What's the address of the token if the assetType is ETH?

packages/types/src/app-instance.ts
cf19drofxots: Sure

packages/types/src/app-instance.ts
snario: It is the default value for the `address` data type which is `AddressZero`

packages/types/src/app-instance.ts
cf19drofxots: Hmm I don't recall atm. Seems like it's not a strict requirement anyway so will remove.

packages/types/src/node-protocol.ts
cf19drofxots: Only reason this exists is because no interface existed for an initiating Node (who creates an AppInstanceId for a proposed AppInstance) to tell a peer Node about said proposal's AppInstanceId.

Not capturing this need via an interface would get hacky and brittle. I'm not married to this naming either but wanted to emphasize this is strictly a Node-to-Node interface concern.

PR #399, opened 2018-12-28
Add noImplicitAny to tsconfigs that already comply

snario:
Adds the rule `"noImplicitAny": true` to `tsconfig.json` files for packages whose source code already has no implicit `any`s.


joelalejandro:
@snario From the looks of the [CI build], tests on the `node-provider` package can't comply with this rule.

Alonski:
It seems that TSLint isn't checking this. Jest however is.

Issue #401, opened 2018-12-28
[node] "ethers/utils" is being referenced as a global "utils"


joelalejandro:
We should check Rollup config for that module and register the necessary globals. Something like:

```js
const globals = {
  // ...
  "ethers": "ethers",
  "ethers/utils": "ethers.utils"
  // ...
}
```

PR #402, opened 2018-12-28
[node] Fix rollup globals

cf19drofxots:
### Description

Specific module references are causing breakages, such as 
```
node.js:492 Uncaught ReferenceError: utils is not defined
    at node.js:492
```

### Related issues

This fixes: https://github.com/counterfactual/monorepo/issues/401

### DIFF
--- a/packages/node/rollup.config.js
+++ b/packages/node/rollup.config.js
@@ -4,7 +4,11 @@ import pkg from "./package.json";
 
 const globals = {
   "@counterfactual/cf.js": "cfjs",
-  "eventemitter3": "EventEmitter"
+  "@counterfactual/types": "types",
+  "eventemitter3": "EventEmitter",
+  "ethers/constants": "ethers.constants",
+  "ethers/utils": "ethers.utils",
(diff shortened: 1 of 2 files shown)


cf19drofxots:
@mvanderh unless we bundle each package that needs `@counterfactual/types` in the dist file, it seems like we'll need to reference the types package somehow and in the case of the iife format, it'd be the global dependency.

mvanderh:
I suppose that's true. Just as long as we don't make it inconvenient for end-users of Counterfactual (which of course doesn't include the Node) by requiring them to have scripts in the global scope.

joelalejandro:
@cf19drofxots @mvanderh I'm thinking maybe we should bundle `types` as part of any IIFE that consumes it. It feels super awkward to inject a script for its constants alone.

snario:
I agree with @joelalejandro.

cf19drofxots:
@joelalejandro I was inclined to do that as well. Since we're all in agreement here I'll update this to reflect that approach.

cf19drofxots:
@joelalejandro request for re-review

PR #405, opened 2018-12-30
[monorepo] Rename AppInstanceId to AppInstanceIdentityHash

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Due to historical reasons, the word `AppInstanceId` in the monorepo refers to two separate things:

- something generated by cf.js via a call to `uuid.v4()` which looks like `9956b015-46b2-4e05-8528-54071bbcfa38`
- the `keccak256` hash of an `AppIdentity` struct, which looks like `c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`

(notice that they do not even encode the same number of bits)

This PR changes  occurrences of `AppInstanceId` referring to the latter to `AppIdentityHash`

### DIFF
--- a/packages/contracts/contracts/ETHVirtualAppAgreement.sol
+++ b/packages/contracts/contracts/ETHVirtualAppAgreement.sol
@@ -26,7 +26,7 @@ contract ETHVirtualAppAgreement {
     AppRegistry registry;
     Transfer.Terms terms;
     uint256 expiry;
-    bytes32 appInstanceId;
+    bytes32 appIdentityHash;
     uint256 capitalProvided;
     address[2] beneficiaries;
   }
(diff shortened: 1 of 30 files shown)


snario:
I'm not sure I agree with this fully. `AppInstanceId` _is_ the hash of an `AppIdentity`. I think a correct fix would be to remove the usage of `AppInstanceId` for that `uuid` generated thing

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
We should discuss which names to assign each concepts then. Suggestions for what to rename the uuid instances to?

snario:
Following discussion with @ebryn @cf19drofxots and @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:

## AppInstanceId
The client-side identifier for an AppInstance

## AppInstanceIdentityHash
The identifier used when submitting the AppInstance up for dispute on-chain and signing commitments

PR #406, opened 2018-12-31
[playground] Integrates Metamask account

joelalejandro:
This PR will expose the following functionality to the Playground:

- [x] Request Metamask permission to connect from the Playground (`web3.currentProvider.enable()`).
- [x] Detect the current Ethereum network being used
- [x] Store the current Ethereum address
- [x] Show the address in the Account Register form 

image

### DIFF
--- a/packages/playground/src/components/widgets/widget-connection/widget-connection.tsx
+++ b/packages/playground/src/components/widgets/widget-connection/widget-connection.tsx
@@ -4,9 +4,9 @@ import NetworkTunnel from "../../../data/network";
 
 const NETWORK_NAMES = {
   "1": "Ethereum (Mainnet)",
-  "3": "Ropsten (Private)",
+  "3": "Ropsten (Testnet)",
   "42": "Kovan (Testnet)",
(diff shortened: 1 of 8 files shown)


packages/playground/src/components/widgets/widget-connection/widget-connection.tsx
```diff
@@ -2,6 +2,12 @@ import { Component, Element, Prop } from "@stencil/core";
 
 import NetworkTunnel from "../../../data/network";
 
+const NETWORK_NAMES = {
+  "1": "Ethereum (Mainnet)",
+  "3": "Ropsten (Private)",
+  "42": "Kovan (Testnet)",
+  "4": "Rinkeby (Private)"
```
ebryn: Not sure where `Private` terminology is coming from. They're all Testnets.

joelalejandro: You're right. For some odd reason, the Spanish translation only showed "Kovan" as a testnet, and the others as "Privada" (Private). I'll modify it accordingly.

PR #407, opened 2018-12-31
[playground-server] [playground] Implements account registration flow

joelalejandro:
This PR allows to create an account to use the Playground.

Supported features:

- [x] API endpoint: `/api/create-account`
- [x] Creation of a multisig between a real Ethereum address and a real Node address
- [x] Playground call to the API upon user registration
- [x] Playground frontend configuration to know the API's URL (shouldn't be hardcoded)
- [x] User should sign the operation with Metamask
- [x] Save the user somewhere (shouldn't just store the multisig address in local storage)
- [x] Should prevent an user from registering twice with the same eth_address
- [x] Should show an error message in the UI for each error case
- [x] Documentation

Signature request example:

image


packages/playground-server/src/index.ts
```diff
@@ -0,0 +1,11 @@
+import mountApi from "./api";
+import node from "./node";
+
+(async () => {
+  const api = mountApi(node);
+  await api.listen(9000);
```
snario: Can we make this an env var?

joelalejandro: Sure! :D

packages/playground-server/src/middleware/create-account.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    if (!response.ok) {
+      ctx.body = response;
+      if (response.error) {
+        ctx.status = response.error.status;
+      }
+      return next();
+    }
+
+    const multisigResponse = await node.call(
```
ebryn: Is there a reason we don't use the method on the Channel object?

packages/playground/src/components/account/account-register/account-register.tsx
```diff
@@ -33,9 +38,32 @@ export class AccountRegister {
...
   }
 
-  formSubmitionHandler() {
-    console.log(this.changeset);
-    this.updateAccount(this.changeset);
+  async formSubmitionHandler() {
+    const data = this.changeset;
+
+    const httpResponse = await fetch(
+      "http://localhost:9000/api/create-account",
```
ebryn: env var?

packages/playground-server/src/middleware/create-account.ts
joelalejandro: @ebryn AFAIK, the intended usage of the Node API is through this method. Everything else is private.

packages/playground/src/components/account/account-register/account-register.tsx
joelalejandro: Yeah, sure

packages/playground-server/src/index.ts
```diff
@@ -0,0 +1,12 @@
+import mountApi from "./api";
+import node from "./node";
+
+(async () => {
+  const api = mountApi(node);
+  const { PLAYGROUND_SERVER_PORT } = process.env;
```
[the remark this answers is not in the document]
joelalejandro: Sure. This is a leftover of the _Netlify function era_ :P didn't know if `PORT` would collide with something else.

packages/playground-server/src/middleware/create-account.ts
ebryn: @cf19drofxots can you confirm?

packages/playground/src/components/account/account-register/account-register.tsx
ebryn: @joelalejandro to be clear, i mean specifically for the host

packages/playground-server/src/middleware/create-account.ts
cf19drofxots: Yeah, it follows the same pattern as other method calls to the Node

https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#method-createmultisig

packages/playground/src/components/account/account-register/account-register.tsx
```diff
@@ -41,27 +37,12 @@ export class AccountRegister {
...
 
-    if (!apiResponse.ok) {
-      const error = apiResponse.error as ErrorResponse;
-      alert(`Something went wrong: ${error.status}`);
-      return;
-    }
-
-    const { multisigAddress } = apiResponse.data as CreateAccountResponseData;
-    window.localStorage.setItem("playground:multisig", multisigAddress);
+    window.localStorage.setItem(
```
snario: Do we have any place in the codebase where we define explicitly the kinds of data stored in `window.localStorage`? I'm always a fan of having a single place where schemas are defined rather than places scattered across the codebase that set arbitrary key-value pairs.

packages/playground/src/data/playground-api-client.ts
```diff
@@ -0,0 +1,46 @@
...
+  CreateAccountRequest,
+  CreateAccountResponseData,
+  ErrorResponse
+} from "@counterfactual/playground-server";
+
+const BASE_URL = `ENV:API_HOST`;
+
+type JsonBody = {
+  [key: string]: string | number | boolean | JsonBody | undefined;
+};
```
snario: Seems like a super common type. Is this not in some generic typings package or Typescript built-in somewhere?

cf19drofxots: IIRC `undefined` is not a valid value in JSON, the equivalent is `null`.

joelalejandro: @snario It's a type generic enough to avoid doing.

```ts
type ApiResponse = CreateAccountResponseData | LoginResponseData | ...ResponseData;
```

The standard is to use `any`, `object` or `{ [key: string]: any }` - but they seem too permissive to me.

packages/playground/src/components/account/account-register/account-register.tsx
joelalejandro: @snario This is the first time we use `localStorage` at all, so we haven't defined anything yet. I also do not know what else we're going to store at the moment, hence the `playground:*` namespace.

snario: OK. Perhaps we should just make a note in the code that, if we ever decide to set this value in some other file or read from it in another file, to abstract it to a more general db accessor? Just want to avoid the scenario where some other `.tsx` file starts overwriting the value for some reason and we end up being unsure how it happens

packages/playground/src/data/playground-api-client.ts
joelalejandro: @cf19drofxots Actually `null` isn't equivalent to `undefined`. `undefined` means a key isn't defined at all in the object, while `null` means a key is present and has a null value.

joelalejandro: In this case, it's used because a key attempting to be accessed maybe undefined.

packages/playground/src/components/account/account-register/account-register.tsx
joelalejandro: @snario Talking with @ebryn we might define a schema to use `localStorage` more like a DB.

packages/playground-server/src/middleware/create-account.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  const providedSignature = params.signature;
+  const expectedMessage = buildSignaturePayload(params);
+  const expectedAddress = verifyMessage(expectedMessage, providedSignature);
+
+  if (
+    // We compare the addresses case-insensitively.
+    // verifyMessage() returns mixed upper and lower case characters,
+    // while eth.accounts[0] always returns lower-case characters.
+    params.address.localeCompare(expectedAddress, "en", {
```
snario: I believe you can use `ethers` here: https://docs.ethers.io/ethers.js/html/api-utils.html#addresses

```js
getAddress(params.address) === expectedAddress
```

joelalejandro: @snario Great, this is way better. I'll implement it right away :)

packages/playground-server/test/api.spec.ts
```diff
@@ -12,18 +15,107 @@ const client = axios.create({
...
+          expect(response.data).toEqual({
+            ok: false,
+            error: {
+              status: 400,
+              errorCode: ErrorCode.SignatureRequired
+            }
+          });
+        });
+    });
+    it("fails when an invalid signature is passed to the request", async () => {
```
snario: Minor nitpick but it would be nice if we had some whitespace between the `it` statements

packages/playground/src/components/account/account-register/account-register.tsx
```diff
@@ -1,10 +1,32 @@
+declare var web3: {
```
snario: I think this is a good candidate to be placed inside `typescript-typings`

snario: https://www.npmjs.com/package/@types/web3 also exists

snario: Unless this is so insignificant so as to not warrant a dependency you think?

joelalejandro: @snario Yeah, I agree. The `@types/web3` package isn't compatible with the Metamask web3 version. It's a shame :man_shrugging:

packages/playground/src/data/playground-api-client.ts
cf19drofxots: @joelalejandro you're right. I was just pointing out that `undefined` isn't supported in JSON and we seem to be using it as part of the JsonBody definition.

joelalejandro: @cf19drofxots Yeah, that's true, but it's valid in terms of a proper result for an unaccessible member.

packages/playground-server/.env-cmdrc
```diff
@@ -0,0 +1,14 @@
+{
```
snario: Can we add a note in the README regarding what this file is for / why we aren't using standard `.env`

joelalejandro: @snario Sure.

snario:
I rebased to master.

One thing I noticed is that the GitHub UI adds both of our profile pics on the commits but the underlying `.git` file shows the correct author.. :s Feel free to force push to undo if it is a problem.

snario:
Sorry @joelalejandro I messed up the GitHub UI history; was trying to resolve a `yarn.lock` conflict. It's back to how it was before I did anything.

PR #413, opened 2019-01-02
[CFjs] Add proposeVirtualInstall and installVirtual methods to Node API spec

mvanderh:
### Description

Add proposeVirtualInstall and installVirtual methods to Node API. This enables installation of virtual apps via metachannels.

### DIFF
--- a/packages/cf.js/API_REFERENCE.md
+++ b/packages/cf.js/API_REFERENCE.md
@@ -11,6 +11,8 @@
             - [Node method](#method-getappinstances)
         - `async install(appInstanceId: AppInstanceID): Promise<AppInstance>`
             - [Node method](#method-install)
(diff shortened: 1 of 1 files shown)


packages/cf.js/API_REFERENCE.md
```diff
@@ -11,6 +11,8 @@
             - [Node method](#method-getappinstances)
         - `async install(appInstanceId: AppInstanceID): Promise<AppInstance>`
             - [Node method](#method-install)
+        - `async installVirtual(appInstanceId: AppInstanceID, intermediaries: Address[]): Promise<AppInstance>`
```
cf19drofxots: Would `intermediaries` be specified again? Could they not just be looked up via the `appInstanceId`?

Cuz I'm imagining the proposal happens through a specified intermediary, so doing the actual install via a possibly different intermediary would deviate from the proposal.

PR #415, opened 2019-01-02
[machine] Add seq number check

cf19drofxots:
### Description

This adds a check for the right sequence number being specified for an AppInstance being installed into a StateChannel.

### DIFF
--- a/packages/machine/src/models/state-channel.ts
+++ b/packages/machine/src/models/state-channel.ts
@@ -33,7 +33,9 @@ const ERRORS = {
   FREE_BALANCE_IDX_CORRUPT: (idx: string) =>
     `Index ${idx} used to find ETH Free Balance is broken`,
   INSUFFICIENT_FUNDS:
-    "Attempted to install an appInstance without sufficient funds"
+    "Attempted to install an appInstance without sufficient funds",
+  INVALID_SEQ_NO:
+    "Attempted to install an appInstance with invalid sequence number"
 };
 
 function createETHFreeBalance(
@@ -115,7 +117,7 @@ export class StateChannel {
     return this.appInstances.get(idx!)!;
   }
 
-  public setupChannel(network: NetworkContext) {
+  public setupChannel(network: NetworkContext): StateChannel {
     const size = this.appInstances.size;
 
     if (size > 0) throw Error(`${ERRORS.APPS_NOT_EMPTY(size)})`);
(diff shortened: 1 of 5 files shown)


packages/machine/src/models/state-channel.ts
```diff
@@ -173,7 +175,11 @@ export class StateChannel {
     appInstance: AppInstance,
     aliceBalanceDecrement: BigNumber,
     bobBalanceDecrement: BigNumber
-  ) {
+  ): StateChannel {
+    if (appInstance.appSeqNo !== this.monotonicNumInstalledApps + 1) {
+      throw Error(`${ERRORS.INVALID_SEQ_NO}`);
```
snario:
```suggestion
      throw Error(ERRORS.INVALID_SEQ_NO);
```

packages/machine/src/models/state-channel.ts
```diff
@@ -115,7 +117,7 @@ export class StateChannel {
     return this.appInstances.get(idx!)!;
   }
 
-  public setupChannel(network: NetworkContext) {
+  public setupChannel(network: NetworkContext): StateChannel {
```
snario: I don't think we need this here

packages/machine/test/unit/models/state-channel/set-state.spec.ts
```diff
@@ -47,7 +47,7 @@ describe("StateChannel::setState", () => {
         token: AddressZero
       },
       false,
-      Math.ceil(Math.random() * 2e10),
+      2,
```
snario: Shouldn't this be 1?

ETHFreeBalance = 0
TestApp = 1

?

packages/machine/src/models/state-channel.ts
cf19drofxots: The function returns a `StateChannel` instance. Not having that in the function signature to indicate the return type is misleading, as one would think the function returns void.

packages/machine/test/unit/models/state-channel/set-state.spec.ts
cf19drofxots: That's what I thought as well. It fails with `1` though, and I didn't spend the time investigating why.

packages/machine/src/models/state-channel.ts
snario: It would be implicitly caught by the type checker

packages/machine/test/unit/models/state-channel/set-state.spec.ts
snario: Ok, I'll look into it

packages/machine/src/models/state-channel.ts
cf19drofxots: Of course, my point was for more targeted towards legibility.

PR #416, opened 2019-01-02
[node] Add InstructionExecutor to requests to methods

snario:
Adds an instance of an `InstructionExecutor` to the `Node` and threads it through into all `methods` by adding an `InstructionExecutor` as the 3rd param in all `RequestHandler`-compliant methods.

### DIFF
--- a/packages/node/src/node.ts
+++ b/packages/node/src/node.ts
@@ -55,15 +55,7 @@ export class Node {
       // account-address-based indexing
       `${nodeConfig.STORE_KEY_PREFIX}/${this.signer.address}`
     );
-    this.instructionExecutor = new InstructionExecutor({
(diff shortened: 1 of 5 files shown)


packages/node/src/node.ts
```diff
@@ -44,12 +49,22 @@ export class Node {
       // account-address-based indexing
       `${nodeConfig.STORE_KEY_PREFIX}/${this.signer.address}`
     );
+    this.instructionExecutor = new InstructionExecutor({
+      // TODO: Pass in NetworkContext into Node constructor
```
cf19drofxots: This should be updated now that `NetworkContext` is being passed in.

snario: Done

PR #430, opened 2019-01-03
[CFjs] Support for installVirtual and proposeInstallVirtual

mvanderh:
### Description

Support for installVirtual and proposeInstallVirtual, required methods for installing virtual apps.


packages/cf.js/API_REFERENCE.md
```diff
@@ -427,8 +427,6 @@ An instance of an installed app.
     - Amount of the asset deposited by the counterparty
 - `timeout: BigNumber`
     - Number of blocks until a submitted state for this app is considered finalized
-- `isVirtual: boolean`
```
mvanderh: Adding this boolean would break a bunch of Node code, which I thought was unecessary since you can derive it from the `intermediaries` field.

packages/cf.js/src/app-factory.ts
```diff
@@ -57,11 +59,7 @@ export class AppFactory {
     const timeout = parseBigNumber(params.timeout, "timeout");
     const myDeposit = parseBigNumber(params.myDeposit, "myDeposit");
     const peerDeposit = parseBigNumber(params.peerDeposit, "peerDeposit");
-    try {
-      getAddress(params.peerAddress);
```
mvanderh: It's not valuable to check whether these addresses are valid since the Node must do this anyway, and will likely return the exact same error as CFjs would here. Extra code complexity with no gain

PR #366, opened 2018-12-25
[contracts] Use ethlint not solium

snario:
Solium was [renamed](https://medium.com/solium/renaming-solium-to-ethlint-18b3cf043d15) and [6 hours ago they did a version bump](https://github.com/duaraghav8/Ethlint/commit/7d2bb796bb697c5e6a114328c7468c041e43cf68) which wasn't put on `npm` under the name `solium` so it [broke our build](https://circleci.com/gh/counterfactual/monorepo/12364) on master.

PR #431, opened 2019-01-03
[contracts] Upgrade Solidity code to 0.5

snario:
This is blocked on [a bug](https://github.com/EthWorks/Waffle/issues/30) that I think is happening in `ethereum-waffle`.

A nice benefit is that Truffle supports Typescript natively now so we don't need to run `tsc`.

In a future PR I'll take advantage of `abi.decode`.


snario:
Currently blocked until #366 is merged

```
@counterfactual/contracts: contracts/libs/Transfer.sol
@counterfactual/contracts:   39:23    error    Syntax error: unexpected token t    
```

packages/apps/contracts/CommitRevealApp.sol
```diff
@@ -37,15 +37,15 @@ contract CommitRevealApp {
     bytes32 actionHash;
   }
 
-  function isStateTerminal(AppState state)
+  function isStateTerminal(AppState memory state)
```
mvanderh: Is `memory` required now for params?

snario: Yes

snario: https://solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html#explicitness-requirements

mvanderh: Cool, I like it

PR #433, opened 2019-01-04
coin fall

patience-tema-baron:
Adds a coin fall animation when the player wins.

Note: the gif is janky, but the animations are smooth when you run high roller yourself.

coin fall


Alonski:
Make it rain!

PR #447, opened 2019-01-07
[machine] rename UpdateData -> UpdateParams

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This change was reverted in #425 due to it being a dependency of #417

### DIFF
--- a/packages/machine/src/protocol/update.ts
+++ b/packages/machine/src/protocol/update.ts
@@ -65,7 +65,7 @@ function proposeStateTransition(
   state: StateChannel
 ) {
   const { appIdentityHash, newState } = message.params as UpdateParams;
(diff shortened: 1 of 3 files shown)


packages/machine/src/protocol/update.ts
```diff
@@ -64,8 +64,8 @@ function proposeStateTransition(
   context: Context,
   state: StateChannel
 ) {
-  const { appIdentityHash, newState } = message.params as UpdateData;
-  context.stateChannel = state.setState(appIdentityHash, newState);
+  const { appIdentityHash, newState } = message.params as UpdateParams;
+  context.stateChannel = state.setState(appInstanceId, newState);
```
snario:
```suggestion
  context.stateChannel = state.setState(appInstanceHash, newState);
```

PR #455, opened 2019-01-07
[node] Add default env vars

cf19drofxots:
### Description

This replaces `dotenv` with `dotenv-extended` to include defaults for env vars used in testing.


packages/node/.env.schema
```diff
@@ -0,0 +1,7 @@
+A_PRIVATE_KEY=
```
snario: You should look into `dotenv-safe`

PR #457, opened 2019-01-07
[node] Refactor request handling

cf19drofxots:
### Description

This slightly refactors how the requests are handled to remove unneeded parameters being passed around.


packages/node/src/node.ts
```diff
@@ -51,17 +49,15 @@ export class Node {
...
     );
-    this.instructionExecutor = new InstructionExecutor(networkContext);
     this.registerMessagingConnection();
     this.requestHandler = new RequestHandler(
       this.incoming,
       this.outgoing,
       this.channels,
       this.messagingService,
-      this.instructionExecutor
+      new InstructionExecutor(networkContext)
```
snario: As context, my initial motivation for having it be a property on `Node` was that I had intended for middlewares to be registered at this scope rather than inside the `RequestHandler`. NBD, makes sense to do this for now.

PR #459, opened 2019-01-07
[node] Refactor node internals

cf19drofxots:
### Description

This destructs the `Channels` class as it was getting too bloated adding extra functionality to it. This spreads the functionality respectively.


packages/node/src/methods/app-instance-operations.ts
```diff
@@ -1,25 +1,72 @@
...
+        return await store.getAppInstanceInfo(appInstanceId);
+      }
+    )
+  );
+}
+
+function getNonFreeBalanceAppInstancesJSON(
+  stateChannel: StateChannel
+): AppInstance[] {
+  return [...stateChannel.appInstances.values()].filter(
```
snario:
```suggestion
  return stateChannel.appInstances.values().filter(
```

packages/node/src/methods/app-instance-operations.ts
```diff
@@ -1,25 +1,72 @@
...
+    (appInstance: AppInstance) => {
+      return !stateChannel.appInstanceIsFreeBalance(appInstance.identityHash);
+    }
```
snario: You can reduce this from `=> { return X }` to `=> X`

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+  store: Store,
+  params: Node.InstallParams
+): Promise<AppInstanceInfo> {
+  if (!params.appInstanceId) {
+    return Promise.reject("No AppInstanceId specified to install");
+  }
+
+  const stateChannel = await store.getChannelFromAppInstanceID(
+    params.appInstanceId
+  );
```
snario: Stylistically, you can do `const { appInstanceId } = params;` above and then this and the following line won't need to be 3 lines long

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+  await store.addAppInstanceProposal(channel, proposedAppInstance);
+  return appInstanceId;
+}
+
+async function install(
+  store: Store,
+  params: Node.InstallParams
+): Promise<AppInstanceInfo> {
+  if (!params.appInstanceId) {
+    return Promise.reject("No AppInstanceId specified to install");
```
snario: Can you add a test that hits this case?

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+  }
+
+  const stateChannel = await store.getChannelFromAppInstanceID(
+    params.appInstanceId
+  );
+
+  const appInstanceInfo = await store.getProposedAppInstanceInfo(
+    params.appInstanceId
+  );
+  const appInstance: AppInstance = createAppInstanceFromAppInstanceInfo(
```
snario:
```suggestion
  const appInstance = createAppInstanceFromAppInstanceInfo(
```

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+    token: proposedAppInstanceInfo.asset.token
+      ? proposedAppInstanceInfo.asset.token
+      : AddressZero
+  };
+
+  return new AppInstance(
+    channel.multisigAddress,
+    // TODO: generate ephemeral app-specific keys
+    channel.multisigOwners,
+    proposedAppInstanceInfo.timeout.toNumber(),
```
snario: Why is it a BigNumber to begin with? My philosophy has been to keep things as `number` types as long as possible for things that rationally should be small numbers (e.g., the timeout or dependency nonce value) and as `BigNumber` the entire way through for things that rationally could be large numbers (e.g., the value in Wei of some free balance value).

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+    // TODO: this should be thread-safe
+    channel.numInstalledApps,
+    channel.rootNonceValue,
+    proposedAppInstanceInfo.initialState,
+    0,
+    proposedAppInstanceInfo.timeout.toNumber()
+  );
+}
+
+function getAppFunctionSigHashes(
```
snario: I might recommend this function _not_ be a method on this class but rather a utility function in some other place. In general I am of the opinion that class methods should be kept to only those that are absolutely necessary to be on the class (mutate internal state).

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+      })`
+    ]).functions.applyAction.sighash;
   }
+
+  return {
+    resolve: resolveSigHash,
+    applyAction: applyActionSigHash,
+    getTurnTaker: getTurnTakerSigHash,
+    isStateTerminal: isStateTerminalSigHash
+  } as AppInterfaceSighashParameters;
```
snario:
```suggestion
  };
```

You can set the return type as `AppInterfaceSighashParameters` and it will type check this object automatically. In this scenario doing type casting actually _weakens_ the type system because `{ blah: blah} as AppInterfaceSighashParameters` would not throw a type error.

packages/node/src/methods/multisig-operations.ts
```diff
@@ -38,13 +60,33 @@ export async function createMultisig(
...
+
+  const state = {
+    alice: stateChannel.multisigOwners[0],
+    bob: stateChannel.multisigOwners[1],
+    aliceBalance: bigNumberify(0),
+    bobBalance: bigNumberify(0)
+  };
+
+  stateChannel = stateChannel.setState(freeBalanceETH.identityHash, state);
+  await this.store.saveChannel(stateChannel);
```
snario: Can we rename this method to `saveStateChannel`

packages/node/src/methods/app-instance-operations.ts
cf19drofxots: `.values()` returns an `Iterator`, which can't be `filter`ed.

snario: Oh wow, weird. I would have thought `Map` would have a `filter` method.

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+ * @param appInstanceInfo The AppInstanceInfo to convert
+ * @param channel The channel the AppInstanceInfo belongs to
+ */
+function createAppInstanceFromAppInstanceInfo(
+  proposedAppInstanceInfo: ProposedAppInstanceInfo,
+  channel: StateChannel
+): AppInstance {
+  const appFunctionSigHashes = getAppFunctionSigHashes(proposedAppInstanceInfo);
+
+  const appInterface: AppInterface = {
```
cf19drofxots: @snario are you specifically referring to return values from functions whose signature explicitly indicate the return type? Or all explicit typings for variables?

For example here: not explicitly defining the type for `appInterface` won't ensure the object has the properties you need it to have, and the error might throw downstream.

packages/node/src/methods/install-operations.ts
cf19drofxots: I kept this in parity with how an `AppInstanceInfo` is defined: https://github.com/counterfactual/monorepo/blob/master/packages/types/src/data-types.ts#L14

packages/node/src/methods/install-operations.ts
cf19drofxots: 👍

packages/node/src/methods/install-operations.ts
snario: In this case it is necessary to specify the type because it is an object literal, but there are other cases where the type can be inferred

cf19drofxots: Right, so the former case then. Cool

packages/node/src/methods/install-operations.ts
```diff
@@ -82,8 +104,175 @@ export async function addAppInstance(
...
+    appInterface,
+    terms,
+    // TODO: pass correct value when virtual app support gets added
+    false,
+    // TODO: this should be thread-safe
+    channel.numInstalledApps,
+    channel.rootNonceValue,
+    proposedAppInstanceInfo.initialState,
+    0,
+    proposedAppInstanceInfo.timeout.toNumber()
```
cf19drofxots: @snario also I'm pretty sure setting this field to be this value isn't right. What should the `latestTimeout` be for an `AppInstance` being instantiated?

packages/node/src/methods/install-operations.ts
cf19drofxots: It's not a method on a class though. It's a standalone function in the same context (read: file) as the related code using it.

packages/node/src/methods/install-operations.ts
snario: Yeah it is kind of ambiguous... it is the value for which the most recent signed state update was signed as. If the most recent state update does not exist (this is an install) then the value should probably be the default timeout value.

We're waiting on @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll to define the parameters in the install protocol for defining initial state too, which I think should include state, nonce, and timeout (this val).

packages/node/src/methods/install-operations.ts
cf19drofxots: Done: https://github.com/counterfactual/monorepo/pull/459/files#diff-f1c40cc9129a886fba505158a6b8d1f3

packages/node/src/methods/install-operations.ts
cf19drofxots: I'll leave this for now to unblock merge.

packages/node/test/unit/install.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+import { AddressZero, One, Zero } from "ethers/constants";
+import { instance, mock, when } from "ts-mockito";
+import { v4 as generateUUID } from "uuid";
+
+import { ERRORS, install } from "../../src/methods/install-operations";
+import { openStateChannel } from "../../src/methods/multisig-operations";
+import { ProposedAppInstanceInfo } from "../../src/models";
+import { Store, STORE_ERRORS } from "../../src/store";
+import { EMPTY_NETWORK } from "../integration/utils";
+import MEMORY_STORE_SERVICE from "../services/memory-store-service";
```
snario: Does it make sense for this to be capitalized?

packages/node/test/unit/install.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+dotenv.load();
+
+describe("Can handle correct & incorrect installs", () => {
+  const storeKeyPrefix = "store";
+
+  it("fails to install without appInstanceId", () => {
+    const store = new Store(MEMORY_STORE_SERVICE, storeKeyPrefix);
+    const params = { appInstanceId: "" };
+    expect(install(store, params)).rejects.toEqual(ERRORS.NO_APP_INSTANCE_ID);
+  });
```
snario: Perhaps we should have two tests -- one for invalid appInstanceId (not bytes32, empty string, etc) and one for empty (undefined or null)

packages/node/test/unit/install.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+import { EMPTY_NETWORK } from "../integration/utils";
+import MEMORY_STORE_SERVICE from "../services/memory-store-service";
+
+dotenv.load();
+
+describe("Can handle correct & incorrect installs", () => {
+  const storeKeyPrefix = "store";
+
+  it("fails to install without appInstanceId", () => {
+    const store = new Store(MEMORY_STORE_SERVICE, storeKeyPrefix);
```
snario: You can instantiate this store in a `beforeAll`

packages/node/test/unit/install.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    when(mockedStore.getProposedAppInstanceInfo(appInstanceId)).thenResolve(
+      proposedAppInstanceInfo
+    );
+    when(mockedStore.getChannelFromAppInstanceID(appInstanceId)).thenReject(
+      STORE_ERRORS.NO_MULTISIG_FOR_APP_INSTANCE_ID
+    );
+
+    const params = { appInstanceId };
+    try {
+      await install(store, params);
```
snario:
```suggestion
      await install(store, { appInstanceId });
```

packages/node/test/unit/install.spec.ts
cf19drofxots: Come to think of it, probably not. I'll camel case it.

packages/node/test/unit/install.spec.ts
cf19drofxots: Yeah that's a fair point. This only covers an empty string because the type checker enforced 
 that on the argument being passed in, but this behavior can't be enforced when this is transpiled to JS so I'll add a second case and override the type checker.

packages/node/test/unit/install.spec.ts
cf19drofxots: Since the second half of the test cases use a mocked store, the `store` variable actually loses uniform typing across the test cases,  so if I were to move anything to a `beforeAll` it'd just be the variable declaration, and there's not much benefit from doing that.

snario: I mean, you can always use multiple `describe` blocks

cf19drofxots: I could, though generally I lean towards having one describe block to describe a test suite per file.

PR #460, opened 2019-01-07
[Node] Store & retrieve app instance commitments

mvanderh:
### Description

Task: https://app.clubhouse.io/counterfactual/story/221/store-generated-commitments-into-the-node-s-storeservice-by-registering-a-callback-to-the-node-s-instructionexecutor-on-state

Couple of points:
- Is it safe to store commitments for appInstanceHash? (i.e. is it unique enough? should I use appInstanceID instead?)

- Install protocol doesn't have an appInstanceHash param to read. It's technically possible to generate it from the other params (signingKeys, appInterface, etc) but that's inconvenient. Can we add an appInstanceHash field to the machine's Context somehow?

- How do I get the counterparty's signature so that I can create and store a commitment? Right now I only have my own signature.

**TODO**: Tests


cf19drofxots:
> Is it safe to store commitments for appInstanceHash? (i.e. is it unique enough? should I use appInstanceID instead?)

I'm assuming you're referring to [`appInstanceIdentityHash`](https://github.com/counterfactual/monorepo/blob/master/packages/machine/src/models/app-instance.ts#L116-L135), in which case it should be safe as each hash is mapped to an `appInstanceId`.

> Install protocol doesn't have an appInstanceHash param to read. It's technically possible to generate it from the other params (signingKeys, appInterface, etc) but that's inconvenient. Can we add an appInstanceHash field to the machine's Context somehow?

Can the functions linked to above be used for this?

> How do I get the counterparty's signature so that I can create and store a commitment? Right now I only have my own signature.

For now this can probably be worked around by providing an ad-hoc signature from the counterparty as presumably we have their private key too (for testing / iteration reasons that is). When the machine is embedded this workaround can be removed.

packages/node/src/channels.ts
```diff
@@ -285,12 +285,34 @@ export class Channels {
...
+  ) {
+    return this.store.setCommitmentForAppInstanceHash(
+      appInstanceHash,
+      protocol,
+      commitment
+    );
+  }
+
   getNonFreeBalanceAppInstancesJSON(stateChannel: StateChannel): AppInstance[] {
     return [...stateChannel.appInstances.values()].filter(
```
snario:
```suggestion
    return stateChannel.appInstances.values().filter(
```

snario: Actually furthermore, the `Map` class has native filter ability. Why are we converting to JSON?

packages/node/src/db-schema.ts
```diff
@@ -59,7 +59,11 @@ export const DB_NAMESPACE_APP_INSTANCE_ID_TO_PROPOSED_APP_INSTANCE =
  * that was installed.
  */
 export const DB_NAMESPACE_APP_INSTANCE_ID_TO_APP_INSTANCE_INFO =
-  "AppInstanceIdToAppInstanceInfo";
+  "appInstanceIdToAppInstanceInfo";
+
+// TODO: description
```
snario: TODO

packages/node/src/node.ts
```diff
@@ -55,6 +56,21 @@ export class Node {
       `${nodeConfig.STORE_KEY_PREFIX}/${this.signer.address}`
     );
     this.instructionExecutor = new InstructionExecutor(networkContext);
+    this.instructionExecutor.middlewares.add(
```
snario: I think there is a `register` method on `InstructionExecutor`

packages/node/src/node.ts
```diff
@@ -55,6 +56,21 @@ export class Node {
...
+      Opcode.STATE_TRANSITION_COMMIT,
+      async (message, next, context) => {
```
snario: I think you need to add the types here? Does type inference work?

packages/node/src/node.ts
```diff
@@ -55,6 +56,21 @@ export class Node {
...
+        const { appIdentityHash } = message.params as any; // TODO: not every params has an appIdentityHash field
+        const transaction = context.commitment!.transaction([
```
snario: If you do type checking you won't need to add `!`

packages/node/src/store.ts
```diff
@@ -263,4 +264,33 @@ export class Store {
...
+
+  async getCommitmentForAppInstanceHash(
+    appInstanceHash: string,
+    protocol: string
+  ): Promise<string> {
+    const key = this.computeCommitmentKey(appInstanceHash, protocol);
+    return this.storeService.get(key);
+  }
+
+  private computeCommitmentKey(appInstanceHash: string, protocol: string) {
```
snario: I think we can type check `protocol` here to be in an enum, probably one already defined in the `types`

packages/node/src/node.ts
mvanderh: It does! Though I prefer adding types anyway. My IDE autogen'd this

packages/node/src/node.ts
mvanderh: Ah right

packages/node/src/channels.ts
mvanderh: Don't know. @cf19drofxots

cf19drofxots:
> For what: adding an appIdentityHash to Context or computing it from Install params?

Computing it from the Install params. If it's too inconvenient then it might be worth it to add the `appIdentityHash` to the machine context.

> What exactly are you suggesting? Should the Node have a temporary COUNTERPARTY_KEY field that it uses to sign commitments before it stores them?

Yeah something like that. This can then be removed once more appropriate plumbing is in place to get the right counterparty signature.

packages/node/src/node.ts
snario: Actually you're right; but it means we should be checking if the value is undefined instead of muting the type checker probably

snario:
This can help with the conversion: https://github.com/counterfactual/monorepo/blob/master/packages/machine/src/ethereum/utils/app-identity.ts

packages/node/src/node.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I would really prefer passing them through return types rather than binding to `STATE_TRANSITION_COMMIT`. If this can be done the opcode should be deleted.

mvanderh: What does "passing them through return types" mean?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
It means to include the commitment in eg `runSetupProtocol`'s return type

Le 10 janv. 2019 12:26 a.m., "Mitchell Van Der Hoeff" <
notifications@github.com> a écrit :

*@mvanderh* commented on this pull request.
------------------------------

In packages/node/src/node.ts
<https://github.com/counterfactual/monorepo/pull/460#discussion_r246443285>:

> @@ -55,6 +56,21 @@ export class Node {
       `${nodeConfig.STORE_KEY_PREFIX}/${this.signer.address}`
     );
     this.instructionExecutor = new InstructionExecutor(networkContext);
+    this.instructionExecutor.middlewares.add(
+      Opcode.STATE_TRANSITION_COMMIT,
+      async (message, next, context) => {

What does "passing them through return types" mean?

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/counterfactual/monorepo/pull/460#discussion_r246443285>,
or mute the thread

packages/node/src/node.ts
snario: This is a separate issue that has been discussed in the past; I'd recommend we open an issue to discuss it, probably to be done after the Playground release.

PR #463, opened 2019-01-08
[ttt] add mocking for proposeInstall and install

patience-tema-baron:
Adds a MockNodeProvider, which supports `proposeVirtualInstall`, `virtualInstall`, and `getAppInstanceDetails`. Additionally, it now routes the user through a waiting room while the playground connects them with another user.


packages/dapp-tic-tac-toe/src/Waiting.jsx
```diff
@@ -0,0 +1,135 @@
...
+    this.countDown();
+    this.initialize();
+  }
+
+  componentWillUnmount() {
+    clearTimeout(this.countdownTimeout);
+  }
+
+  async initialize() {
+    await this.props.nodeProvider.connect();
```
ebryn: any reason you don't do this up in the `App` class? there's no need for anything but the `cfProvider` to have a reference to it

packages/dapp-tic-tac-toe/src/Node.js
```diff
@@ -0,0 +1,32 @@
...
+  EventName: {
+    INSTALL: "installEvent",
+    REJECT_INSTALL: "rejectInstallEvent",
+    UPDATE_STATE: "updateStateEvent",
+    UNINSTALL: "uninstallEvent",
+    PROPOSE_STATE: "proposeStateEvent",
+    REJECT_STATE: "rejectStateEvent",
+    CREATE_MULTISIG: "createMultisigEvent"
+  }
+}
```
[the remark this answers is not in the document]
ebryn: Why do we use them if we're not using TS?

packages/dapp-tic-tac-toe/src/MockNodeProvider.js
```diff
@@ -0,0 +1,112 @@
...
+  sendMessage(message) {
+    if (!this.isConnected) {
+      // We fail because we do not have a messagePort available.
+      throw new Error(
+        "It's not possible to use postMessage() before the NodeProvider is connected. Call the connect() method first."
+      );
+    }
+    console.log("message in", message);
+    switch (message.type) {
+      case Node.MethodName.PROPOSE_INSTALL_VIRTUAL: {
```
ebryn: just use the string values instead of the enums

packages/dapp-tic-tac-toe/src/Waiting.jsx
patience-tema-baron: It was to prevent the Waiting component from attempting to send a message before a connection was established. I went ahead and moved `connect` to App, which I agree makes more sense architecturally. Since the child components may still attempt to send messages before a connection is established, though, I compromised by locking the screen until after `connect` resolves.

patience-tema-baron: Removed the types, though unfortunately we've got to keep the package as a dep because other packages depend on it.

PR #465, opened 2019-01-08
[playground-server] Implements Matchmaking API

joelalejandro:
This PR adds the POST `/api/matchmake` endpoint to the Playground Server.

- If there is only one user to matchmake, it'll return that particular user.
- If there are more users, it'll select a random user from the database, excluding the requesting user, and return its username and address.
- If there are no users, it'll return a `no_users_available` error.

It also refactors the error control system, putting in place a proper middleware for that.

:information_source: **This PR does not add an API call from the Playground front-end.**

### Request format
```json
{
  "userAddress": "0x0f693cc956df59dec24bb1c605ac94cadce6014d"
}
```

#### When a user is matched
```json
{
  "ok": true,
  "data": {
    "username": "bob",
    "peerAddress": "0x93678a4828d07708ad34272d61404dd06ae2ca64"
  }
}
```

#### If there are no users
```json
{
  "ok": false,
  "error": {
    "status": 400,
    "errorCode": "no_users_available"
  }
}
```

#### If something unexpectedly fails
```json
{
  "ok": false,
  "error": {
    "status": 500,
    "errorCode": "matchmake_failed"
  }
}
```


snario:
There is one feature that I think we'll need right, which is the "matchmake by filter" e.g., only matchmake for High Roller

ebryn:
> There is one feature that I think we'll need right, which is the "matchmake by filter" e.g., only matchmake for High Roller

No, I don't think so. Either way it's out of scope at this moment. The next thing we'd probably want is more of a filter by who is currently online.

packages/playground-server/src/api.ts
```diff
@@ -23,6 +25,18 @@ export default function mountApi() {
   return api;
 }
 
+export function createErrorResponseForDatabase(
```
ebryn: can we move this elsewhere?

packages/playground-server/src/middleware/matchmake.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (!params.userAddress) {
+    return createErrorResponse(400, ErrorCode.UserAddressRequired);
+  }
+
+  return { ok: true };
+}
+
+export default function matchmake() {
+  return async (ctx: Context, next: () => Promise<void>) => {
+    // TODO: Do we need signature validation here?
```
ebryn: You need to be logged in to call `matchmake`. Let's add this when we add the login route.

packages/playground-server/src/middleware/matchmake.ts
```diff
@@ -0,0 +1,59 @@
...
+import {
+  ApiResponse,
+  ErrorCode,
+  MatchmakeRequest,
+  MatchmakeResponseData,
+  PlaygroundUserData
+} from "../types";
+
+function validateRequest(params: MatchmakeRequest): ApiResponse {
+  if (!params.userAddress) {
```
ebryn: this shouldn't be necessary when this becomes an authenticated request

packages/playground-server/src/middleware/matchmake.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    response.data = {
+      username: matchedUser.username,
+      peerAddress: matchedUser.address
+    } as MatchmakeResponseData;
+
+    ctx.status = 200;
+    ctx.body = response;
+
+    return next();
```
ebryn: it'd be cool to have a `try { } catch() { } finally { }` pattern here to avoid having to explicitly call `next()` twice?

packages/playground-server/src/types.ts
```diff
@@ -48,7 +61,7 @@ export type GetAppsResponseData = {
 };
 
 export type PlaygroundUserData = {
-  email: string;
+  email?: string;
```
ebryn: shouldn't be optional

packages/playground-server/src/middleware/matchmake.ts
joelalejandro: Fixed via d404f19. We now have a proper error middleware, so no more `try...catch` needed in the controllers.

packages/playground-server/src/types.ts
joelalejandro: Fixed via 3bdd92d. Matchmaking now has its own `MatchmakeUserData` type.

packages/playground-server/src/middleware/matchmake.ts
joelalejandro: Will be cleared on the Login route PR.

packages/playground-server/src/api.ts
joelalejandro: Done via d404f19. We now have a proper error middleware, so no more try...catch needed in the controllers. This function is no longer necessary.

PR #469, opened 2019-01-08
[node] Remove need for tsc in Node

snario:
Don't think we need this


packages/node/rollup.config.js
```diff
@@ -62,9 +62,7 @@ export default [
       nodeResolve({
         only: [...bundledDependencies]
       }),
-      typescript({
-        tsconfig: "tsconfig.rollup.json"
-      }),
+      typescript(),
```
cf19drofxots: If the `tsconfig.json` is removed then what configuration is rollup looking at?

cf19drofxots: 👍

cf19drofxots:
How does this affect an IDE's type checking that's based on a local installation of `tsc`?

snario:
> How does this affect an IDE's type checking that's based on a local installation of `tsc`?

Shouldn't have an effect

cf19drofxots:
@snario this breaks transpilation for the tests so a `yarn build` doesn't catch errors anymore.

Not sure if this optimization is worth the hassle of figuring out a workaround. The previous configuration was working just fine.

snario:
> @snario this breaks transpilation for the tests so a `yarn build` doesn't catch errors anymore.
> 
> Not sure if this optimization is worth the hassle of figuring out a workaround. The previous configuration was working just fine.

https://github.com/counterfactual/monorepo/pull/527

PR #477, opened 2019-01-09
[WIP] [contracts & apps] Upgrade App definition to use abi.decode

snario:
This creates a new abstract contract called `CounterfactualApp` which is an interface for all applications that adhere to the Counterfactual interface.

Since `abi.decode` exists in Solidity 0.5, we can pass in `bytes` instead of contract-specific structs to the apps which is what enables the new abstract contract (i.e., interface) to exist.

- Support for dynamic types in application state structs
- More strict requirements for the interface of an application
- Compile-time checks for adherence to Counterfactual interface
- Can get rid of `LibStaticCall` since `staticcall` is now used by default in Solidity 0.5 for functions with the `pure` modifier
- No more need to store sighashes in `AppInterface`
- `getTurnTaker` can return an `address` in the `signingKeys` array instead of an index

- [ ] Remove `LibStaticCall`
- [ ] Update the app definitions in the `apps` folder
- [ ] Remove the need to store any sighashes in contracts
- [ ] Remove all usage of sighashes in the rest of the codebase
- [ ] Write an example app with dynamic types (e.g., MixEth)


packages/contracts/contracts/default-apps/ETHBalanceRefundApp.sol
```diff
@@ -12,11 +13,13 @@ contract ETHBalanceRefundApp {
     uint256 threshold;
   }
 
-  function resolve(AppState memory state, Transfer.Terms memory terms)
+  function resolve(bytes memory encodedState, Transfer.Terms memory terms)
     public
     view
     returns (Transfer.Transaction memory)
   {
+    AppState memory state = abi.decode(encodedState, (AppState));
```
mvanderh: Very cool!

Why do you need brackets round (AppState)?

packages/contracts/contracts/libs/Transfer.sol
```diff
@@ -40,7 +40,9 @@ library Transfer {
       uint256 value = txn.value[i];
 
       if (txn.assetType == uint8(Transfer.Asset.ETH)) {
-        to.transfer(value);
+        // solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html
+        address payable receiver = address(uint160(to));
```
mvanderh: Can we not change the type of `to` to `address payable` instead?

PR #480, opened 2019-01-10
[Documentation] Fix README files regarding yarn version and package links

Alonski:
### Description

I updated the wording for **Make sure you have Yarn v1.10.1 installed or higher** a few days ago but didn't update it for all of the README.md files, this completes that.

Also I noticed that Header links for each README was pointing to an incorrect URL that was returning a 404. Fixed that as well.

There is still an issue with images in the top right of each README. I am not sure what image is supposed to be there. The src points to https://static1.squarespace.com/static/59ee6243268b96cc1fb2b14a/t/5af73bca1ae6cf80fc1cc250/1529369816810/?format=1500w

However this returns a `Website Expired` page when visited.


snario:
We can point to the logo in the repo root:

```markdown
![](../../logo.svg)
```

https://github.com/counterfactual/monorepo/blame/master/README.md#L3

PR #493, opened 2019-01-10
[playground] Fixes Node/dApp communication

joelalejandro:
DappContainer was echoing messages instead of actually feeding messages to the Node back and forth.


ebryn:
_(For others paying attention, this is intentionally WIP code because we need to debug in our staging env due to our local dev env setup not being fully baked yet)_

PR #494, opened 2019-01-10
Remove sighashes from AppInterface in TS and AppInterface in Solidity

snario:
One of the benefits of #477


mvanderh:
This doesn't break any code in the Machine?

snario:
Nope

PR #496, opened 2019-01-10
Change import / export usage of machine types

snario:
Exports made explicit by the machine and imported precisely without labels by the node.

### DIFF
--- a/packages/node/src/methods/request-handler.ts
+++ b/packages/node/src/methods/request-handler.ts
@@ -1,8 +1,10 @@
+import { InstructionExecutor } from "@counterfactual/machine";
 import {
-  InstructionExecutor,
-  types as machineTypes
-} from "@counterfactual/machine";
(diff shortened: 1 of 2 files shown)


packages/node/src/methods/request-handler.ts
```diff
@@ -1,8 +1,10 @@
...
 import {
-  InstructionExecutor,
-  types as machineTypes
-} from "@counterfactual/machine";
-import { Protocol } from "@counterfactual/machine/dist/src/types";
+  Context,
+  Opcode,
+  Protocol,
+  ProtocolMessage
+} from "@counterfactual/machine/dist/src/types";
```
mvanderh: Wasn't the whole point of this PR to import all these things from "@counterfactual/machine"?

snario: Oh woops my bad, fixing

PR #497, opened 2019-01-10
Make instructionExecutor member of Node

snario:
Requirement in order to use `this.signer` for registering a middleware on `OP_SIGN`


cf19drofxots:
I'd err on the side of not having PRs that are _too_ simple. This change doesn't effectively do much, as there are no tests or anything for it.

PR #498, opened 2019-01-10
[Node] Add registerOpSignMiddleware to Node

snario:


packages/node/src/node.ts
```diff
@@ -65,6 +72,24 @@ export class Node {
...
+            "Reached OP_SIGN middleware without generated commitment."
+          );
+        }
+        context.signature = this.signer.signDigest(
+          context.commitment.hashToSign()
+        );
+        next();
+      }
+    );
+    console.info("Successfully registered OP_SIGN middleware.");
```
cf19drofxots: No need for this log

PR #499, opened 2019-01-10
[machine] Replace OP_SIGN_VALIDATE with function

snario:
It turns out the implementation is the same for all protocols.


cf19drofxots:
This doesn't actually validate a signature though. Is it a WIP?

snario:
> This doesn't actually validate a signature though. Is it a WIP?

It does -- it verifies that `fromAddress` signed `commitment.hashToSign()` which was generated earlier in the machine's execution path.

cf19drofxots:
Ah I stand correct - missed the last `if` statement. 👍

PR #505, opened 2019-01-10
[node] Create api-spec.ts

snario:
Some file restructuring to make explicit the public API.

### DIFF
--- a/packages/node/src/api-spec.ts
+++ b/packages/node/src/api-spec.ts
@@ -1,0 +1,23 @@
+import { Node } from "@counterfactual/types";
+
+import {
+  createMultisig,
+  getAllChannelAddresses,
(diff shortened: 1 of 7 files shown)


packages/node/src/api-spec.ts
```diff
@@ -0,0 +1,21 @@
+import { Node } from "@counterfactual/types";
```
cf19drofxots: The name of this file made me think it's a test file at first glance lol

snario: Maybe api router is a better name?

cf19drofxots: 👍

cf19drofxots: I thought the name was gonna be changed before merging? @snario

PR #506, opened 2019-01-10
[node] Add runInstallProtocol in install method

snario:
~Currently hits an error until we merge in #498.~

EDIT: Rebased after it was merged

### DIFF
--- a/packages/node/src/methods/app-instance/install/operation.ts
+++ b/packages/node/src/methods/app-instance/install/operation.ts
@@ -1,4 +1,8 @@
-import { AppInstance, StateChannel, InstructionExecutor } from "@counterfactual/machine";
+import {
+  AppInstance,
+  InstructionExecutor,
+  StateChannel
+} from "@counterfactual/machine";
 import {
   AppInstanceInfo,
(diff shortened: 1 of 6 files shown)


packages/node/src/methods/app-instance/install/operation.ts
```diff
@@ -13,6 +17,9 @@ import { ERRORS } from "../../errors";
 
 export async function install(
   store: Store,
+  instructionExecutor: InstructionExecutor,
+  initiatingAddress: string,
+  respondingAddress: string,
```
cf19drofxots: The rest of the code base uses `peerAddress` to the refer to the other Node - let's remain consistent

snario: This is a bit tricky because the parameters are named in the protocol specification as of the latest update by @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll

cf19drofxots: Alright, then gotta update the rest of the codebase to use `respondingAddress` then.

PR #514, opened 2019-01-12
[node] Update file structure

cf19drofxots:
### Description

This destructs the internal file structure of the Node as it was becoming unwieldy and introduces more logical placements.


cf19drofxots:
I did do a `git mv` - the github UI doesn't show it as only a rename because the file contents were also modified, but git kept the correct history and you can still blame: https://github.com/counterfactual/monorepo/blame/node_internal_refactor/packages/node/src/api-router.ts

PR #521, opened 2019-01-13
[node] Define Node message payload types & extend Node events

cf19drofxots:
### Description

This explicitly defines the payload types for the messages that get passed between Nodes.

It also extends the Node events to make a more sensible events API.


packages/node/src/events/propose-install/controller.ts
```diff
@@ -0,0 +1,17 @@
+import { RequestHandler } from "../../request-handler";
+import { ProposeMessage } from "../../types";
+
+import { setAppInstanceIDForProposeInstall } from "./operation";
+
+export async function proposeInstallEventController(
+  this: RequestHandler,
```
snario: The pattern of using the variable name `this` when its a bit of a lie / hack seems like it could be a problem. Perhaps just name the variable `requestHandler`?

packages/node/src/types.ts
```diff
@@ -0,0 +1,49 @@
+import { Address, Node } from "@counterfactual/types";
+
+// TODO: the contents of this file should probably be in the types package
```
cf19drofxots: @snario thoughts here?

packages/node/src/events/propose-install/operation.ts
```diff
@@ -7,15 +7,17 @@ import { getChannelFromPeerAddress } from "../../utils";
 export async function setAppInstanceIDForProposeInstall(
   selfAddress: Address,
   store: Store,
-  params: Node.InterNodeProposeInstallParams
+  params: Node.ProposeInstallParams,
+  appInstanceId: string,
+  respondingAddress: Address
```
snario: I think we should be consistent with using `string` and not `Address`

packages/node/src/methods/app-instance/install/controller.ts
```diff
@@ -29,12 +29,13 @@ export default async function installAppInstanceController(
...
-  const installApprovalMsg: NodeMessage = {
+  const installApprovalMsg: InstallMessage = {
     from: this.address,
-    event: Node.EventName.INSTALL,
+    event: NODE_EVENTS.INSTALL,
     data: {
-      appInstanceId: appInstance.id,
-      proposal: false
+      params: {
+        appInstanceId: appInstance.id
```
snario: I just realized after having been confused by the variable name here the confusion that arises from the subtle difference between `AppInstance` and `AppInstanceInfo`. I think we should rename all of these variables to `appInstanceInfo` until we find a way to make the distinction cleaner through abstraction.

Namely, `AppInstanceInfo.prototype.id` is a _property_ to fetch a declared UUID style identifier but `AppInstance.prototype.identityHash` (previously `.id`) is a getter and a _computed property_ for the hash of the `AppInstance.prototype.identity` (an `AppIdentity`).

packages/node/src/methods/state-channel/add.ts
```diff
@@ -2,18 +2,20 @@ import { StateChannel } from "@counterfactual/machine";
...
-  const params = nodeMsg.data;
+export async function addMultisig(
+  this: RequestHandler,
+  nodeMsg: CreateMultisigMessage
+) {
   let stateChannel = new StateChannel(
-    params.multisigAddress,
-    params.owners
+    nodeMsg.data.multisigAddress,
+    nodeMsg.data.params.owners
```
snario: It seems to me like the notion of a "Node Message" / `nodeMsg` / `NodeMessage` should not be something that the internal implementation of the `Node` (i.e., the "methods") should know about. Perhaps it makes sense that it is only ever _received_ by the `RequestHandler` and then _passed_ to a "controller" and at that point it is dissolved into arguments passed into the actual function implementations.

I think we mostly have this right now, but maybe a way to clean up the code a bit would be to have a single location where `NodeMessages` are converted into function arguments with no other logic involved. For example:

```typescript
function reducer(requestHandler: RequestHandler, msg: NodeMessage) {
    switch (msg.type) {
        case NodeMessage.CREATE_MULTISIG:
            addMultisig({
                multisigAddress: msg.data.multisigAddress,
                owners: msg.data.params.owners,
            }, {
                networkContext: requestHandler.networkContext
            });
            break;
        ...
    }
}
```

This way we would benefit from:
- Easier unit testing for all of the "message handlers"
- Clean separation of `NodeMessage` and the actual functional implementation of the `Node`

packages/node/src/events/propose-install/controller.ts
cf19drofxots: Sure

packages/node/src/events/propose-install/operation.ts
cf19drofxots: This code base is consistently using `Address` (even though it's just a string type alias)

packages/node/src/methods/app-instance/install/controller.ts
cf19drofxots: Are you specifically referring to `appInstanceId`?

Following previous discussions, the term `appInstanceId` is explicitly said to refer to the ID of the AppInstanceInfo (which you're referring to as `AppInstanceInfo.prototype.id`) and `appInstanceIdentityHash` refers to the computed property which you're referring to as `AppInstance.prototype.identityHash`.

This is explained here: https://github.com/counterfactual/monorepo/blob/master/packages/node/src/db-schema.ts#L12-L30

cf19drofxots: Can you further clarify which variables you're referring to for renaming though? Because the params don't return an `AppInstanceInfo`, they return an `AppInstanceInfo`'s ID.

packages/node/src/methods/state-channel/add.ts
cf19drofxots: Sgtm

packages/node/src/methods/app-instance/install/controller.ts
snario: Right, I agree with the parameter `appInstanceId`, but I just mean the variable in the code above named `appInstance` I think we should name `appInstanceInfo` because it is an `AppInstanceInfo` object, not an `AppInstance` object

packages/node/src/events/propose-install/operation.ts
snario: OK -- maybe we can discuss later, not in scope for this pr.

packages/node/src/types.ts
snario: If it is not used in any other package, I'd just keep it in this package. I think the requirement for `@counterfactual/types` should be "it is used in >= 2 other packages"

cf19drofxots: But it'll be used by the playground, as it'll need to know what events to register callbacks for and what arguments those callbacks receive.

cf19drofxots:
@joelalejandro specifically the test cases (*.spec.ts) will be the most relevant changes for you in this change set.

PR #529, opened 2019-01-14
[WIP] [PoC] [node] Messaging service spike

snario:
## Notes

- [x] There is an error in the tests for `propose-virtual-install` which shows `thrown: undefined` for all tests in the test suite. I dug into it, and it seems to be failing at this line:

https://github.com/counterfactual/monorepo/blob/8c66c5efd486b032b2c3e31d2902b21d3c049cea/packages/node/src/services.ts#L66-L68

However also, when I googled it, [I found this](https://github.com/facebook/jest/issues/7172).

- [x] I feel like there are way too many `async` and `awaits` being used across the `IO_WAIT` work. Would love some thoughts on perhaps a better interface. Maybe `IMessageService` should have `onReceive` instead of `receive`.

### DIFF
--- a/packages/machine/src/protocol/install.ts
+++ b/packages/machine/src/protocol/install.ts
@@ -33,7 +33,8 @@ export const INSTALL_PROTOCOL = {
     Opcode.IO_WAIT,
 
     // Verify they did indeed countersign the right thing
-    validateSignature,
+    (message: ProtocolMessage, context: Context, stateChannel: StateChannel) =>
+      validateSignature(context.inbox[0], context, stateChannel),
 
     // Consider the state transition finished and commit it
     Opcode.STATE_TRANSITION_COMMIT
@@ -51,6 +52,17 @@ export const INSTALL_PROTOCOL = {
 
     // Wrap the signature into a message to be sent
     prepareToSendSignature,
+
+    function prepareToSendSignature(
+      message: ProtocolMessage,
+      context: Context
+    ) {
+      context.outbox[0] = {
+        ...context.outbox[0],
+        fromAddress: context.outbox[0].toAddress,
+        toAddress: context.outbox[0].fromAddress
+      };
+    },
 
     // Send the message to your counterparty
     Opcode.IO_SEND,
--- a/packages/node/src/services.ts
+++ b/packages/node/src/services.ts
@@ -6,10 +6,7 @@ import { NodeMessage } from "./types";
 
 export interface IMessagingService {
   send(respondingAddress: Address, msg: NodeMessage): Promise<void>;
-  onReceive(
-    address: Address,
-    callback: (msg: NodeMessage) => void
-  ): Promise<void>;
+  onReceive(address: Address, callback: (msg: NodeMessage) => void);
 }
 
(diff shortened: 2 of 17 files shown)


snario:
Noticed this warning message in the tests (despite them all passing). @cf19drofxots can you have a look?

```
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  TCPSERVERWRAP

      16 |
      17 |   createServer() {
    > 18 |     return new FirebaseServer(this.port, this.host);
         |            ^
      19 |   }
      20 | }
      21 |

      at new WebSocketServer (../../node_modules/ws/lib/websocket-server.js:68:20)
      at new FirebaseServer (../../node_modules/firebase-server/index.ts:146:14)
      at TestFirebaseServiceFactory.createServer (test/integration/services/firebase-service.ts:18:12)
      at Object.beforeAll (test/integration/node-store.spec.ts:19:49)
```

packages/node/src/services.ts
```diff
@@ -56,41 +61,56 @@ class FirebaseMessagingService implements IMessagingService {
     private readonly messagingServerKey: string
   ) {}
 
-  async send(respondingAddress: Address, msg: object) {
-    await this.firebase
+  send(respondingAddress: Address, msg: object) {
+    const sanitizedMsg = JSON.parse(JSON.stringify(msg));
+    this.firebase
```
cf19drofxots: So I thought this returns a promise, and I awaited it because I wanted confirmation on the operation succeeding, and throwing a reject when the operation failed.

This seems like it'd ignore the reject if the operation fails?

snario:
Actually I think it's because `ProtocolMessage` is being massed in, not ` NodeMessage`:

https://github.com/counterfactual/monorepo/blob/3d495cb70d20a59d77bb779f1ee51248aa4bf3b9/packages/node/src/node.ts#L93

cf19drofxots:
I wanna merge this in but I feel like getting the parameters right for sending is directly in scope of hooking up the messaging service properly.

snario:
Yeah. There is also another important item I need to do which is make the machine middleware asynchronous (I think), because the following middleware needs to be blocking:

https://github.com/counterfactual/monorepo/blob/3d495cb70d20a59d77bb779f1ee51248aa4bf3b9/packages/node/src/node.ts#L101-L109

packages/machine/src/protocol/install.ts
```diff
@@ -52,6 +53,17 @@ export const INSTALL_PROTOCOL = {
     // Wrap the signature into a message to be sent
     prepareToSendSignature,
 
+    function prepareToSendSignature(
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: what's this step for?

snario: Swaps `fromAddress` and `toAddress` before sending it to the counterparty. The only reason the function is called `prepareToSendSignature` is that I copy and pasted that function into this location and was moving too quickly to care to delete the function name, so it makes no sense out of context sorry.

snario:
Closing as better pull requests are incoming.

PR #532, opened 2019-01-15
[monorepo] cleanup and fixes

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- rename `free-balance.ts` to `eth-bucket.ts`
- rename some functions defined inside
- make `setupChannel` static
- rename `aliceBalanceDecrement` and `bobBalanceDecrement` 
- remove `terms` from `InstallVirtualAppParams`; instead calculate it from other params
- change public API `runSetupProtocol` to not require any `StateChannel`s to be passed in
- add `multisigOwners` to `SetupParams`


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
marking as [WIP] while I comment out some failing tests

PR #535, opened 2019-01-15
Persist workplace for built packages to speed up CircleCI

mvanderh:


.circleci/config.yml
```diff
@@ -15,6 +15,10 @@ jobs:
           key: dependency-cache-{{ checksum ".dependencies_checksum" }}
           paths:
             - node_modules
+      - save_cache:
+          key: build-cache-{{ .Revision }}
+          paths:
+            - packages/**/dist
```
snario: I think the path for `contracts` is `build`

.circleci/config.yml
```diff
@@ -36,8 +42,9 @@ jobs:
       - run: cat packages/*/package.json | shasum > .dependencies_checksum
       - restore_cache:
           key: dependency-cache-{{ checksum ".dependencies_checksum" }}
+      - attach_workspace:
+          at: /home/circleci/project
       - run: yarn # symlink packages' node_modules
```
snario:
```suggestion
```

.circleci/config.yml
```diff
@@ -24,8 +29,9 @@ jobs:
...
```
snario:
```suggestion
```

.circleci/config.yml
mvanderh: Don't need it anymore?

snario: I don't think so right? Wouldn't `yarn build` fail if the dependencies weren't installed?

.circleci/config.yml
cf19drofxots: I put that there to make it explicitly clear why `yarn` is being run

Without context you'd be tempted to remove it because you'd think `yarn` has already ran to pull the dependencies so it doesn't need to be run again.

snario: I think we won't need it though because the workspace is persisted, thus the symlinks are too

mvanderh: In this PR i persist only the dist/ and build/ folders if that matters.

cf19drofxots: Cool - if it works with the updated config then it's all good

This comment was just for _that_ config which seems stale now

snario: Hm ok let's explore this later then, this PR looks good

PR #536, opened 2019-01-15
[node] Sanitize AppInstance proposal 

cf19drofxots:
### Description

This ensures correct behavior whether or not the initial state of an AppInstance is set during a proposal.


packages/node/src/models/proposed-app-instance-info.ts
```diff
@@ -16,7 +16,7 @@ export interface IProposedAppInstanceInfo {
   myDeposit: BigNumber;
   peerDeposit: BigNumber;
   timeout: BigNumber;
-  initialState: AppState;
+  initialState?: AppState;
```
snario: Wait, why is this being made possibly undefined?

packages/node/src/models/proposed-app-instance-info.ts
```diff
@@ -63,16 +63,20 @@ export class ProposedAppInstanceInfo implements AppInstanceInfo {
     appInstanceId: AppInstanceID,
     proposeParams: IProposedAppInstanceInfo
   ) {
+    if (!proposeParams.abiEncodings.actionEncoding) {
+      delete proposeParams.abiEncodings.actionEncoding;
+    }
```
snario: We can just JSON stringify and and JSON parse at firebase-entry-time to handle this I think

packages/node/src/models/proposed-app-instance-info.ts
cf19drofxots: Because it may be possibly undefined when it gets passed in...as in the initial state has literally not been defined?

packages/node/src/models/proposed-app-instance-info.ts
cf19drofxots: Done.

packages/node/src/models/proposed-app-instance-info.ts
```diff
@@ -30,7 +30,7 @@ export interface ProposedAppInstanceInfoJSON {
   myDeposit: string;
   peerDeposit: string;
   timeout: string;
-  initialState: AppState;
+  initialState?: AppState;
```
snario: As mentioned on the call, I don't think we can allow this

PR #537, opened 2019-01-15
[node] Node uses stable address across instantiations

cf19drofxots:
### Description

This allows the Node to be instantiated while having a stable address, using the store to keep track of whether the Node has created a private key or not.


snario:
What is the benefit of using `static create` over `new`?

cf19drofxots:
In this case, `Await`able calls can't be made in a constructor.

Unless we modify the store service interface, I didn't see any other way to achieve this functionality.

PR #544, opened 2019-01-16
[machine] Update InstructionExecutor with messaging and sig validation logic

snario:
Summary:

Necessary to make the message a counterparty sends in response (to trigger `IO_WAIT`s callback) distinct from the initial message sent to the counterparty which is intended to trigger `dispatchReceivedMessage`.

This had previously been ambiguous. Now, we assume that any `IO_WAIT` middleware will add to the `context.inbox` array of `ProtocolMessage`s. For most protocols, it explicitly adds a `ProtocolMessage` with a `signature` on it that corresponds to the counterparties signature on the commitment generated by `proposeStateTransition`.

We verify that `message.toAddress` is the signer of `context.commitment` with the signature from `context.inbox[0].signature`.


packages/machine/src/protocol/utils/signature-validator.ts
```diff
@@ -1,26 +1,21 @@
...
-import { StateChannel } from "../../models";
-import { ProtocolMessage } from "../../protocol-types-tbd";
-import { Context } from "../../types";
+import { EthereumCommitment } from "../../ethereum/types";
 
 export function validateSignature(
-  message: ProtocolMessage,
-  context: Context,
-  state: StateChannel
+  expectedSigner: string,
```
mvanderh: Use Address type

packages/machine/test/unit/protocol/signature-validator.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  it("throws if the signature is wrong", () => {
+    expect(() =>
+      validateSignature(
+        signer.address,
+        commitment,
+        signer.signDigest(HashZero.replace("00", "11")) // 0x111111...
+      )
+    ).toThrow("Received invalid signature on validateSignature");
+  });
```
mvanderh: You could add a test for when the signature is for the right hash, but by the wrong signer.

packages/machine/src/protocol/utils/signature-validator.ts
snario: I'm actually against using `Address` because although it signals to the developer that this field should be an ethereum address, it creates a false sense of safety. I think instead a better practice would be to always write some kind of decorator for functions that expect addresses and instead get either empty string or strings which are not hex values or strings which aren't exactly 42 characters long.

packages/machine/test/unit/protocol/signature-validator.spec.ts
snario: That's actually what this is testing and I think it is technically impossible to do the inverse

packages/machine/src/protocol/utils/signature-validator.ts
mvanderh: Yeah. This is the catch of Typescript: there's no "real" guarantees of type safety.

packages/machine/src/protocol/utils/signature-validator.ts
```diff
@@ -1,26 +1,21 @@
...
+  commitment?: EthereumCommitment,
+  signature?: Signature
```
cf19drofxots: Why is this optional? The function shouldn't even be called if a signature can't be given for validation.

packages/machine/src/protocol/utils/signature-validator.ts
```diff
@@ -1,26 +1,21 @@
...
-import { ProtocolMessage } from "../../protocol-types-tbd";
-import { Context } from "../../types";
+import { EthereumCommitment } from "../../ethereum/types";
 
 export function validateSignature(
-  message: ProtocolMessage,
-  context: Context,
-  state: StateChannel
+  expectedSigner: string,
+  commitment?: EthereumCommitment,
```
cf19drofxots: Same comment as below.

packages/machine/test/unit/instruction-executor.spec.ts
```diff
@@ -29,44 +29,46 @@ describe("InstructionExecutor", () => {
...
+      });
+
+      // Ensure validateSignature in Setup Protocol does not throw
+      instructionExecutor.register(Opcode.IO_WAIT, (_, __, context) => {
+        context.inbox.push({
+          signature: responder.signDigest(context.commitment!.hashToSign())
+        } as ProtocolMessage);
+      });
 
       const scm = new Map<string, StateChannel>([
```
cf19drofxots: This should have a better name.

packages/machine/test/unit/protocol/signature-validator.spec.ts
cf19drofxots: You can have a signature for the wrong hash with right signer.

The signature would just be for a different commitment than the one being validated against.

snario: My point is that there is no way to distinguish between those two cases

packages/machine/src/protocol/install.ts
```diff
@@ -24,16 +29,25 @@ export const INSTALL_PROTOCOL = {
...
-    prepareToSendSignature,
+    addSignedCommitmentToOutboxForSeq1,
 
     // Send the message to your counterparty
     Opcode.IO_SEND,
 
     // Wait for them to countersign the message
     Opcode.IO_WAIT,
 
+    // Verify a message was received
```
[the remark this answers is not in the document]
snario: They could be but I like the idea of each step doing exactly one specific check.

packages/machine/src/protocol/utils/signature-forwarder.ts
```diff
@@ -1,13 +1,24 @@
 import { ProtocolMessage } from "../../protocol-types-tbd";
 import { Context } from "../../types";
 
-export function prepareToSendSignature(
+export function addSignedCommitmentToOutboxForSeq1(
```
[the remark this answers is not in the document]
snario: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll and I discussed this. It represents the specific arrow on a sequence diagram between 2 parties and the best way to represent it, we think, it numerically. Message 0, message 1, etc. Candidate replacement names are `Message.InitialMessage`, `Message.AckMessage`, etc but in some cases there are many messages and many acks.

packages/machine/src/protocol/utils/signature-validator.ts
snario: To reduce code reuse in the protocol flow files and this file.

packages/machine/src/protocol/install.ts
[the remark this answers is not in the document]
cf19drofxots: The checks are directly related to each other though, but this is basically a non-issue. Moving on.

packages/machine/src/protocol/utils/signature-forwarder.ts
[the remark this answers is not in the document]
cf19drofxots: I figured - can we have docstrings on these function definitions to make it a bit more clear though

packages/machine/src/protocol/utils/signature-validator.ts
cf19drofxots: Doesn't sound like a clean separation but non-blocking for now.

PR #546, opened 2019-01-16
[monorepo] Remove unknown rollup options

snario:
These aren't recognized by rollup

### DIFF
--- a/packages/node/rollup.config.js
+++ b/packages/node/rollup.config.js
@@ -50,7 +50,6 @@ export default [
         globals: globals
       }
     ],
-    sourceMap: true,
     external: external,
     plugins: [
       commonjs({
--- a/packages/node-provider/rollup.config.js
+++ b/packages/node-provider/rollup.config.js
@@ -18,7 +18,6 @@ export default [
     ],
     plugins: [
       typescript()
-    ],
-    exclude: ["test"]
+    ]
   }
 ];


packages/node/rollup.config.js
```diff
@@ -50,7 +50,6 @@ export default [
         globals: globals
       }
     ],
-    sourceMap: true,
```
cf19drofxots: Where did you see rollup not recognizing this?

According to its docs, it does: https://rollupjs.org/guide/en#big-list-of-options

snario: That's only if its inside one of the things inside the `output` key, not in the top level

snario: And I saw this in the output of running `yarn build`

```
⚙️  Building package: node
$ rollup -c
(!) You have passed an unrecognized option
Unknown input option: sourceMap. Allowed options: acorn, acornInjectPlugins, cache, chunkGroupingSize, context, experimentalCacheExpiry, experimentalOptimizeChunks, experimentalTopLevelAwait, external, inlineDynamicImports, input, manualChunks, moduleContext, onwarn, perf, plugins, preserveModules, preserveSymlinks, shimMissingExports, treeshake, watch
```

cf19drofxots: Then move it to `output`?

snario: It is already there

PR #547, opened 2019-01-16
[node] Miscellaneous code cleanup

snario:


packages/node/src/events/index.ts
```diff
@@ -1,10 +1,11 @@
-export { default as installEventController } from "./install/controller";
+import { installEventController } from "./install/controller";
```
cf19drofxots: No need to import these like this if that file only has one thing being exported (i.e. `export default`)

packages/node/src/events/propose-install-virtual/operation.ts
```diff
@@ -19,15 +19,8 @@ export async function setAppInstanceIDForProposeInstallVirtual(
   appInstanceId: string,
   incomingAddress: Address
 ) {
-  const channel = await getChannelFromPeerAddress(
-    selfAddress,
-    incomingAddress,
-    store
+  await store.addAppInstanceProposal(
```
cf19drofxots: I'm generally not a fan of doing this because _unless you look at the function signature_ to see what it's expecting (which is too tedious when you're just _looking_ at its invocation), these inliners are unnecessary optimizations that take away from easy legibility.

In the previous case, it's extremely obvious immediately what's being passed in.

packages/node/src/store.ts
```diff
@@ -59,7 +59,9 @@ export class Store {
    * Returns the StateChannel instance with the specified multisig address.
    * @param multisigAddress
    */
-  async getStateChannel(multisigAddress: Address): Promise<StateChannel> {
+  public async getStateChannel(
```
cf19drofxots: It's `public` by default. No need to specify unless making it `private`.

packages/node/src/events/index.ts
cf19drofxots: The style of importing things one by one then exporting them all together is nice though.

packages/node/src/events/install/controller.ts
```diff
@@ -18,3 +18,5 @@ export async function installEventController(
     nodeMsg.data.params
   );
 }
+
+export default installEventController;
```
cf19drofxots: Why not just leave `default` in the function signature?

Saves having to export down here.

snario: Now it is exporting both as a symbol _and_ default, so importers have a choice

packages/node/src/events/propose-install-virtual/operation.ts
snario: I don't understand to be honest, this change only removes the variable names but is otherwise the same thing. I think you can infer that the parameter being passed in is a "channel" if the function name is `getChannelFromPeerAddress` and the second parameter is a "proposedAppInstance" if the value is `new ProposedAppInstanceInfo`

packages/node/src/store.ts
snario: This is just a preference of always biasing to be explicit about things when possible

packages/node/src/events/install/controller.ts
cf19drofxots: I mean, this is what `defaults` are for really - to indicate that a file only has one thing to import from, so the `import { x } from './x'` seems like an antipattern to `default` usage.

snario: Hmmm, ok I'll remove this

packages/node/src/events/propose-install-virtual/operation.ts
cf19drofxots: In this case you can infer, but it might not always be super clear.

And because it's not always clear, then you'd end up with sometimes having this style and sometimes having that style, thus introducing stylistic inconsistencies.

So the explicit variable naming is consistently erring on the side of easier legibility

packages/node/src/store.ts
cf19drofxots: 👍

packages/node/src/events/propose-install-virtual/operation.ts
cf19drofxots: Yeah I'd be in favor of that

That'd also remove parameter ordering concerns

PR #548, opened 2019-01-17
[node] Implements AppInstance takeAction

cf19drofxots:
### Description

This implements the [`takeAction`](https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#method-takeaction) public API method.

Note that this is only for AppInstances, not Virtual AppInstances. The PR for Virtual AppInstances will follow this.


packages/machine/src/models/app-instance.ts
```diff
@@ -245,4 +245,11 @@ export class AppInstance {
       latestTimeout: timeout
     });
   }
+
+  public decodeState(encodedAppState: string): any {
```
snario: `decodeAppState`?

packages/node/src/methods/app-instance/take-action/controller.ts
```diff
@@ -0,0 +1,49 @@
...
+import { ERRORS } from "../../errors";
+
+import { generateNewAppInstanceState } from "./operation";
+
+export default async function takeActionController(
+  requestHandler: RequestHandler,
+  params: Node.TakeActionParams
+): Promise<Node.TakeActionResult> {
+  const { appInstanceId, action } = params;
+  if (!appInstanceId) {
```
snario:
```suggestion

  if (!appInstanceId) {
```

(nitpick sorry)

packages/node/src/methods/app-instance/take-action/controller.ts
```diff
@@ -0,0 +1,49 @@
...
+        newState
+      }
+    }
+  };
+
+  const appInstanceInfo = await requestHandler.store.getAppInstanceInfo(
+    appInstanceId
+  );
+
+  requestHandler.messagingService.send(
```
snario:
```suggestion
  await requestHandler.messagingService.send(
```

packages/node/src/methods/app-instance/take-action/operation.ts
```diff
@@ -0,0 +1,41 @@
+import { AppInstance } from "@counterfactual/machine";
+import { AppState } from "@counterfactual/types";
+import { Contract } from "ethers";
+import { Provider } from "ethers/providers";
+
+import { ERRORS } from "../../errors";
+
+export async function generateNewAppInstanceState(
+  appInstance: AppInstance,
+  action: any,
```
snario: We can define a new type alias here I think. `type AppAction = AppState`.

However at the same time I recommend renaming `AppState` to `ABIEncoderV2Struct` and reusing that here too.

packages/node/src/methods/app-instance/take-action/operation.ts
```diff
@@ -0,0 +1,41 @@
...
+  provider: Provider
+): Promise<AppState> {
+  if (
+    !appInstance.appInterface.addr ||
+    appInstance.appInterface.addr.trim() === ""
+  ) {
```
snario: Could we wrap this logic in a util file and call it `isEmptyString`? There may even be an alias on the `Sting` prototype

packages/node/src/methods/app-instance/take-action/operation.ts
```diff
@@ -0,0 +1,41 @@
...
+    return Promise.reject(ERRORS.NO_APP_CONTRACT_ADDR);
+  }
+
+  const abi = [
+    `function applyAction(${appInstance.appInterface.stateEncoding}, ${
+      appInstance.appInterface.actionEncoding
+    }) pure returns (bytes)`
```
snario: Two things:

1. Probably it would be nice to put this string construction into a file called `function-signatures.ts` or something like that and have it be called e.g., `getApplyActionSignature(stateEncoding, actionEncoding)`.

2. Inside that function you can use newlines, e.g.,

```typescript
const signature = `
    function applyAction(
        ${appInstance.appInterface.stateEncoding} state,
        ${appInstance.appInterface.actionEncoding} action
    )
        pure
        returns (bytes)
`;
```

3. Furthermore, this is actually in flux at the moment. Very soon everything will use the exact same interface:

`function applyAction(bytes, bytes) pure returns (bytes)`

packages/node/src/methods/app-instance/take-action/operation.ts
```diff
@@ -0,0 +1,41 @@
...
+  try {
+    return appInstance.decodeState(
+      await appContract.functions.applyAction(appInstance.state, action)
+    );
+  } catch (e) {
+    const sanitizedError = e
+      .toString()
+      .replace("s: VM Exception while processing transaction: revert");
+    return Promise.reject(`${ERRORS.INVALID_ACTION}: ${sanitizedError}`);
+  }
```
snario: This also could be put inside a functional unit-test-able util file, e.g., `eth_call.ts` or something more descriptive

packages/node/src/methods/errors.ts
```diff
@@ -7,5 +7,9 @@ export const ERRORS = {
   NO_PROPOSED_APP_INSTANCE_FOR_APP_INSTANCE_ID:
     "No proposed AppInstance exists for the given appInstanceId",
   NULL_INITIAL_STATE_FOR_PROPOSAL:
-    "A proposed AppInstance cannot have an empty initial state"
+    "A proposed AppInstance cannot have an empty initial state",
+  NO_APP_INSTANCE_FOR_TAKE_ACTION:
+    "No AppInstanceId specified to takeAction on",
+  NO_APP_CONTRACT_ADDR: "The App Contract address is empty",
+  INVALID_ACTION: "Invalid action taken"
```
snario: You could include the error from the blockchain here. Usually it will actually revert with a message like: "cannot place X here!"

packages/node/test/integration/cant-take-action.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    firebaseServiceFactory = new TestFirebaseServiceFactory(
+      process.env.FIREBASE_DEV_SERVER_HOST!,
+      process.env.FIREBASE_DEV_SERVER_PORT!
+    );
+    firebaseServer = firebaseServiceFactory.createServer();
+    messagingService = firebaseServiceFactory.createMessagingService(
+      process.env.FIREBASE_MESSAGING_SERVER_KEY!
+    );
+    nodeConfig = {
+      STORE_KEY_PREFIX: process.env.FIREBASE_STORE_PREFIX_KEY!
```
snario: Minor comment but it might be cleaner stylistically to have a single file called `env.spec.ts` or something where we fetch all these variables once and them import them symbolically as `TestEnvironment`

packages/node/test/integration/cant-take-action.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    };
+
+    privateKey = new SigningKey(hexlify(randomBytes(32))).privateKey;
+    provider = new Web3Provider(
+      ganache.provider({
```
snario: Does this make a new blockchain per test suite?

I tried using `setupFiles` in `jest` before with Ganache but couldn't figure it out.

snario: Also `ethereum-waffle` has some nice helper methods for this kind of stuff. e.g., `getMockProvider`

packages/node/test/integration/cant-take-action.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+      TicTacToeApp.bytecode,
+      wallet
+    ).deploy();
+  });
+
+  beforeEach(async () => {
+    storeServiceA = firebaseServiceFactory.createStoreService(
+      process.env.FIREBASE_STORE_SERVER_KEY! + generateUUID()
+    );
+    nodeA = await Node.create(
```
snario: I do kind of agree with @ebryn that we should name it explicitly as `createAndConnectToRemoteDB` or something like that just to be very clear.

packages/node/test/integration/cross-node-multisig-creation.spec.ts
```diff
@@ -15,6 +16,7 @@ describe("Node can create multisig, other owners get notified", () => {
   let nodeB;
   let storeServiceB: IStoreService;
   let nodeConfig: NodeConfig;
+  let provider: Provider;
```
snario:
```suggestion
  let provider: JsonRpcProvider;
```

packages/node/test/integration/get-app-instances.spec.ts
```diff
@@ -19,6 +20,7 @@ describe("Node method follows spec - getAppInstances", () => {
   let storeService: IStoreService;
   let node: Node;
   let nodeConfig: NodeConfig;
+  let provider: Provider;
```
snario:
```suggestion
  let provider: JsonRpcProvider;
```

packages/node/test/integration/get-app-instances.spec.ts
```diff
@@ -32,14 +34,17 @@ describe("Node method follows spec - getAppInstances", () => {
     nodeConfig = {
       STORE_KEY_PREFIX: process.env.FIREBASE_STORE_PREFIX_KEY!
     };
+    // fake provider as nothing is listening on this URL
+    provider = new JsonRpcProvider("localhost:8545");
```
snario: Maybe use `ts-mockito` instead?

packages/node/test/integration/get-state.spec.ts
```diff
@@ -20,6 +21,7 @@ describe("Node method follows spec - getAppInstances", () => {
   let storeService: IStoreService;
   let node: Node;
   let nodeConfig: NodeConfig;
+  let provider: Provider;
```
snario:
```suggestion
  let provider: JsonRpcProvider;
```

packages/node/test/integration/get-state.spec.ts
```diff
@@ -91,9 +95,13 @@ describe("Node method follows spec - getAppInstances", () => {
 
     const getStateResult = await node.call(getStateReq.type, getStateReq);
     const state = (getStateResult.result as NodeTypes.GetStateResult).state;
-    expect(state).toEqual(
+    expect(state.foo).toEqual(
+      // @ts-ignore
```
snario: Instead of ignoring, use type casting

packages/node/test/integration/take-action.spec.ts
```diff
@@ -0,0 +1,225 @@
...
+      })
+    );
+
+    const wallet = new Wallet(privateKey, provider);
+    tttContract = await new ContractFactory(
+      TicTacToeApp.interface,
+      TicTacToeApp.bytecode,
+      wallet
+    ).deploy();
+  });
```
snario: You can abstract this whole `beforeAll` thing into some kind of `blockchain-tests-setup-helper` file, it is duplicated many times

packages/node/test/integration/utils.ts
```diff
@@ -123,12 +135,11 @@ export function makeInstallProposalRequest(
   if (!nullInitialState) {
     initialState = {
       foo: AddressZero,
-      bar: 0
+      bar: bigNumberify(0)
```
snario: `Zero`

packages/node/test/integration/get-state.spec.ts
cf19drofxots: Why restrict the type of provider that can be passed in?

packages/node/src/methods/app-instance/take-action/operation.ts
cf19drofxots: I tried `function applyAction(bytes, bytes) pure returns (bytes)`

It failed.

cf19drofxots: Whenever that ends up working we can update.

packages/node/src/methods/errors.ts
cf19drofxots: It does - that's how the error sanitation behaves right now. So are you suggesting we make this a function instead of a constant?

packages/node/test/integration/cant-take-action.spec.ts
cf19drofxots: I think so.

cf19drofxots: But I don't want a mock provider here, I want an actual provider.

packages/node/test/integration/cant-take-action.spec.ts
cf19drofxots: I don't understand where this notion of "connecting" is coming from.

The only reason there's an `init` function is to make an async get on the store, which has nothing to do with connecting to anything.

@snario can you elaborate?

packages/node/test/integration/cross-node-multisig-creation.spec.ts
cf19drofxots: Curious what the benefit of this is?

cf19drofxots:
> I recommend we add a simple validator on the caller-side when passing in the action AppState object which does a quick encoding of the object relative to the AppInstance for which it is based on and if the encoding fails, return an error to the client (e.g., INVALID_PROPOSED_ACTION)

Seems like you're referring to two different kinds of error.

`INVALID_PROPOSED_ACTION` can't be short-circuited - it has to make the eth call as the action itself would be invalid even though the action object (and its encoding) would be fine.

What you're referring to is probably more like `INVALID_ACTION_OBJECT` which would indicate the failure to encode.

snario:
That sounds right.

packages/node/src/methods/app-instance/take-action/operation.ts
cf19drofxots: Sounds good. I'll make a follow up PR for that.

packages/node/test/integration/cant-take-action.spec.ts
cf19drofxots: Yeah, that's cleaner - I'll put up another PR for that.

packages/node/test/integration/take-action.spec.ts
cf19drofxots: Yeah this duplication is undesirable. I'll refactor this.

cf19drofxots: I'll clean this up along with https://github.com/counterfactual/monorepo/pull/548/files#r248885645

cf19drofxots:
The error for no encoding info gets thrown before this function is called.

PR #549, opened 2019-01-17
[WIP] Compile contracts using native solc on CI

mvanderh:
This will speed up CI pretty drastically.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,9 @@ jobs:
       - image: circleci/node:10
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


mvanderh:
@snario Yep, agree. I'm getting it working first.

mvanderh:
This isn't worth doing. It saves 5 seconds during the build process on contract compilation but brings with it a whole bunch of extra complexity (new CI workflow steps & caching, breaks with apps/, need separate waffle.json config files and package.json scripts). Even with more contract code I don't think the speedup would be significant enough.

PR #550, opened 2019-01-17
[node] Small additions to Node to add a wrapped ProtocolMessage type

snario:


packages/node/src/events/protocol-message/controller.ts
```diff
@@ -0,0 +1,20 @@
+import { StateChannel } from "@counterfactual/machine";
+
+import { RequestHandler } from "../../request-handler";
+import { NodeMessageWrappedProtocolMessage } from "../../types";
+
+/**
```
cf19drofxots: Stale docstring :P

packages/node/src/events/protocol-message/controller.ts
```diff
@@ -3,10 +3,6 @@ import { StateChannel } from "@counterfactual/machine";
 import { RequestHandler } from "../../request-handler";
 import { NodeMessageWrappedProtocolMessage } from "../../types";
 
-/**
```
cf19drofxots: Don't remove it...Update it!

PR #551, opened 2019-01-17
[node] Replace state transitions in Node to occur inside machine

snario:


packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -52,6 +59,7 @@ export default async function createMultisigController(
 }
 
 function generateNewMultisigAddress(owners: Address[]): Address {
+  // FIXME: Even before CREATE2 this is incorrect
```
cf19drofxots: It was a placeholder, not actually creating any sort of multisig.

PR #552, opened 2019-01-18
[node] Initial implementation of IO_SEND

snario:
This implements `IO_SEND` and adds some supporting mocks to ensure
it actually works.

So what is going on here is that real messages are being sent _from_ nodes however when they are expecting to receive the message, they aren't actually listening yet (that is a forthcoming PR or commit for `IO_WAIT`). Instead, that behaviour is mocked out.

### DIFF
--- a/packages/machine/src/protocol/setup.ts
+++ b/packages/machine/src/protocol/setup.ts
@@ -80,13 +80,28 @@ export const SETUP_PROTOCOL: ProtocolExecutionFlow = {
 };
 
 function proposeStateTransition(message: ProtocolMessage, context: Context) {
-  const { multisigAddress } = message.params as SetupParams;
-  const sc = context.stateChannelsMap.get(multisigAddress)!;
-  if (sc === undefined) {
-    console.log("sc keys=", context.stateChannelsMap.keys());
-    throw Error(`no such channel at multisig address ${multisigAddress}`);
+  const {
+    multisigAddress,
+    initiatingAddress,
+    respondingAddress
+  } = message.params as SetupParams;
+
+  if (!context.stateChannelsMap.has(multisigAddress)) {
+    context.stateChannelsMap.set(
+      multisigAddress,
+      new StateChannel(
+        multisigAddress,
+        [initiatingAddress, respondingAddress].sort((a, b) =>
+          parseInt(a, 16) < parseInt(b, 16) ? -1 : 1
+        )
+      )
+    );
   }
+
+  const sc = context.stateChannelsMap.get(multisigAddress)!;
+
   const newStateChannel = sc.setupChannel(context.network);
+
   context.stateChannelsMap.set(multisigAddress, newStateChannel);
   context.commitment = constructSetupOp(context.network, newStateChannel);
 }
--- a/packages/node/src/events/protocol-message/controller.ts
+++ b/packages/node/src/events/protocol-message/controller.ts
@@ -12,6 +12,17 @@ export default async function protocolMessageEventController(
   requestHandler: RequestHandler,
   nodeMsg: NodeMessageWrappedProtocolMessage
 ) {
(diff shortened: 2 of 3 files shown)


packages/machine/src/protocol/setup.ts
```diff
@@ -80,13 +80,28 @@ export const SETUP_PROTOCOL: ProtocolExecutionFlow = {
...
-  if (sc === undefined) {
-    console.log("sc keys=", context.stateChannelsMap.keys());
-    throw Error(`no such channel at multisig address ${multisigAddress}`);
+  const {
+    multisigAddress,
+    initiatingAddress,
+    respondingAddress
+  } = message.params as SetupParams;
+
+  if (!context.stateChannelsMap.has(multisigAddress)) {
```
snario: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll question for you. Should this function do the logic below this comment? Namely, creating a new `StateChannel` object in the case where a Setup Protocol message is received from a counterparty and it is responding?

packages/node/src/events/protocol-message/controller.ts
```diff
@@ -12,6 +12,17 @@ export default async function protocolMessageEventController(
...
+  // FIXME: Take into account the protocols sequence length. More
+  //        generally this if statement is trying to capture the notion:
+  //        "do not restart a protocol if a message is received from another
+  //         client mid-protocol execution (assume IO_WAIT is listening)"
+  //
+  //        An alternative might be that IO_WAIT is hooked onto toms event
+  //        listeniner inside the instructionExecutor that emits noise
+  //        when dispatchReceivedMessage receives a message for an in-progress
+  //        protocol execution.
+  if (nodeMsg.data.seq === 2) return;
```
snario: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll Question. Should there be some kind of way for `dispatchReceivedMessage` to pass messages into `IO_WAIT` callbacks specifically?

packages/machine/src/protocol/setup.ts
```diff
@@ -80,13 +80,28 @@ export const SETUP_PROTOCOL: ProtocolExecutionFlow = {
...
+    context.stateChannelsMap.set(
+      multisigAddress,
+      new StateChannel(
+        multisigAddress,
+        [initiatingAddress, respondingAddress].sort((a, b) =>
```
cf19drofxots: Can we make this sorting part of the StateChannel constructor?

That way
1) the client doesn't have to keep doing it
2) the sorting implementation becomes standardized

packages/node/src/events/protocol-message/controller.ts
cf19drofxots: Maybe log something here to record the notion described above?

packages/node/src/node.ts
```diff
@@ -90,21 +90,41 @@ export class Node {
...
     );
   }
 
   private registerIoMiddleware() {
-    // TODO:
-    this.instructionExecutor.register(Opcode.IO_SEND, () => {});
+    this.instructionExecutor.register(
+      Opcode.IO_SEND,
+      async (message: ProtocolMessage, next: Function, context: Context) => {
+        await this.messagingService.send(context.outbox[0].toAddress, {
```
cf19drofxots: Just to confirm, a protocol is never restarted mid-execution, right?

So there's never a case where this `context.outbox` array is gonna be empty?

packages/machine/src/protocol/setup.ts
snario: I don't think it should be in the constructor actually because it hides some important complexity, but perhaps there could be a _validator_ in the constructor?

packages/node/src/node.ts
snario: That is correct

PR #556, opened 2019-01-18
[node] Rename init

snario:
Not a priority but thought I'd put this here

### DIFF
--- a/packages/node/src/node.ts
+++ b/packages/node/src/node.ts
@@ -53,7 +53,7 @@ export class Node {
       provider
     );
 
-    return await node.asyncronouslyConnectRemoteServices();
+    return await node.asyncronouslySetupUsingRemoteServices();
   }
 
   private constructor(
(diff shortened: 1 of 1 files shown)


packages/node/src/node.ts
```diff
@@ -70,7 +70,7 @@ export class Node {
     this.registerIoMiddleware();
   }
 
-  private async init() {
+  private async asyncronouslyConnectRemoteServices(): Promise<Node> {
```
cf19drofxots: This isn't the most accurate either as it's entirely connecting to remote services.

A more accurate name would be `asyncronouslySetupUsingRemoteServices`

cf19drofxots:
```suggestion
  private async asyncronouslySetupUsingRemoteServices(): Promise<Node> {
```

packages/node/src/node.ts
```diff
@@ -52,8 +52,8 @@ export class Node {
       nodeConfig,
       provider
     );
-    await node.init();
-    return node;
+
+    return await node.asyncronouslyConnectRemoteServices();
```
cf19drofxots:
```suggestion
    return await node.asyncronouslySetupUsingRemoteServices();
```

PR #558, opened 2019-01-18
[node] Implements rejectInstall

cf19drofxots:
### Description

This implements the [`rejectInstall`](https://github.com/counterfactual/monorepo/blob/master/packages/cf.js/API_REFERENCE.md#method-rejectInstall) public API method.

Note this is only for non-virtual AppInstances. Implementation for rejecting a Virtual AppInstance will follow this.

### DIFF
--- a/packages/node/src/types.ts
+++ b/packages/node/src/types.ts
@@ -66,3 +66,9 @@ export interface TakeActionMessage extends NodeMessage {
     appInstanceId: string;
   };
 }
+
+export interface RejectProposalMessage extends NodeMessage {
+  data: {
+    appInstanceId: string;
+  };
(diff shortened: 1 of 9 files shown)


packages/node/src/types.ts
```diff
@@ -66,3 +66,9 @@ export interface TakeActionMessage extends NodeMessage {
     appInstanceId: string;
   };
 }
+
+export interface RejectProposalMessage extends NodeMessage {
+  data: {
```
cf19drofxots: There should probably be another field in the data section (maybe: `reason` or something) so the rejecting Node can indicate why they rejected the proposal.

PR #559, opened 2019-01-19
[wip] [playground-server] Refactors the API to use a more robust implementation of JSON API

joelalejandro:
Do not merge yet. The Playground and the dApps need a bit of refactoring.


packages/playground-server/src/api.ts
```diff
@@ -1,34 +1,34 @@
...
-          registryPath: path.resolve(__dirname, "../registry.json")
-        })
-        .inject(UsersController, "users")
-        .inject(SessionController, "session")
-        .inject(MatchmakingController, "matchmaking")
-        .routes()
+      processors
+        .register("app", AppProcessor)
+        .register("matchmakingRequest", MatchmakingRequestProcessor)
+        .register("sessionRequest", SessionRequestProcessor)
```
ebryn: I think we can probably just name this `session`

packages/playground-server/src/resources/app/resource.ts
```diff
@@ -0,0 +1,10 @@
+import { Resource } from "@ebryn/jsonapi-ts";
+
+export default interface App extends Resource {
```
ebryn: I wonder if we need to extend Resource?

PR #567, opened 2019-01-21
[node, machine] Deep machine integration into the Node (including IO_SEND, IO_WAIT)

snario:
This pull request incorporates changes initially spiked out in #529.

1. Updates the protocols to correctly and safely do signature validation based on the initial message that kicked off the machine's execution and the commitment generated and stored inside context.

2. Attaches the `messagingService` to the `IO_SEND` opcode as middleware hook.

3. Attached the `IO_WAIT` opcode to a middleware which checks for a promise of a received value from the counterparty. Due the nature of the middleware architecture and wanting to only have a single "messaging listener" attached on the Node (to avoid duplication and opt instead for filtering) I introduced a `Deferred` type which extracts the `resolve` callback inside a `Promise` to an external method. Thus, the messaging receiver _resolves_ the `Deferred` and the `IO_WAIT` middleware hook listens for the resolution before resolving itself and the `IO_SEND` registers the `Deferred` itself _before_ sending the message to ensure it won't get caught in a race condition (where the message is received before the hook is set). 

4. Introduces a new event type, `protocolMessageEventController` which dispatches a new message type `NodeWrappedProtocolMessage` to the `InstructionExecutor`'s `receiveDispatchedMessage` handler.

5. Updates to the `playground-server` to support the requirements of messaging (namely: signing by the correct address which is the public facing address of the Node and correct checksummed addresses as the network addresses & signed message data addresses)

### DIFF
--- a/packages/machine/src/models/state-channel.ts
+++ b/packages/machine/src/models/state-channel.ts
@@ -43,8 +43,9 @@ const ERRORS = {
     "multisigOwners parameter of StateChannel must be sorted"
 };
 
-const sortAddresses = (addrs: string[]) =>
-  addrs.sort((a, b) => (parseInt(a, 16) < parseInt(b, 16) ? -1 : 1));
+function sortAddresses(addrs: string[]) {
+  return addrs.sort((a, b) => (parseInt(a, 16) < parseInt(b, 16) ? -1 : 1));
+}
 
 export type StateChannelJSON = {
   readonly multisigAddress: string;
--- a/packages/machine/src/models/state-channel.ts
+++ b/packages/machine/src/models/state-channel.ts
@@ -43,6 +43,9 @@ const ERRORS = {
     "multisigOwners parameter of StateChannel must be sorted"
 };
 
+const sortAddresses = (addrs: string[]) =>
+  addrs.sort((a, b) => (parseInt(a, 16) < parseInt(b, 16) ? -1 : 1));
+
 export type StateChannelJSON = {
   readonly multisigAddress: string;
   readonly multisigOwners: string[];
@@ -104,9 +107,7 @@ export class StateChannel {
     private readonly monotonicNumInstalledApps: number = 0,
     public readonly rootNonceValue: number = 0
   ) {
-    const sortedMultisigOwners = multisigOwners.sort((a, b) =>
-      parseInt(a, 16) < parseInt(b, 16) ? -1 : 1
-    );
+    const sortedMultisigOwners = sortAddresses(multisigOwners);
     multisigOwners.forEach((owner, idx) => {
       if (owner !== sortedMultisigOwners[idx]) {
         throw new Error(ERRORS.MULTISIG_OWNERS_NOT_SORTED);
@@ -182,9 +183,11 @@ export class StateChannel {
     multisigAddress: string,
     multisigOwners: string[]
   ) {
+    const sortedMultisigOwners = sortAddresses(multisigOwners);
+
     const fb = createETHFreeBalance(
       multisigAddress,
-      multisigOwners,
+      sortedMultisigOwners,
       network.ETHBucket
     );
 
@@ -196,7 +199,7 @@ export class StateChannel {
 
     return new StateChannel(
       multisigAddress,
-      multisigOwners,
+      sortedMultisigOwners,
       appInstances,
       new Map<string, ETHVirtualAppAgreementInstance>(),
       freeBalanceAppIndexes,
--- a/packages/machine/src/models/state-channel.ts
+++ b/packages/machine/src/models/state-channel.ts
@@ -180,7 +180,7 @@ export class StateChannel {
   }
 
   public static setupChannel(
-    network: NetworkContext,
+    ethBucketAddress: string,
     multisigAddress: string,
     multisigOwners: string[]
   ) {
@@ -189,7 +189,7 @@ export class StateChannel {
     const fb = createETHFreeBalance(
       multisigAddress,
       sortedMultisigOwners,
-      network.ETHBucket
+      ethBucketAddress
     );
 
     const appInstances = new Map<string, AppInstance>();
--- a/packages/machine/src/models/state-channel.ts
+++ b/packages/machine/src/models/state-channel.ts
@@ -38,7 +38,9 @@ const ERRORS = {
   FREE_BALANCE_IDX_CORRUPT: (idx: string) =>
     `Index ${idx} used to find ETH Free Balance is broken`,
   INSUFFICIENT_FUNDS:
-    "Attempted to install an appInstance without sufficient funds"
+    "Attempted to install an appInstance without sufficient funds",
+  MULTISIG_OWNERS_NOT_SORTED:
+    "multisigOwners parameter of StateChannel must be sorted"
 };
 
 export type StateChannelJSON = {
@@ -101,7 +103,16 @@ export class StateChannel {
     > = new Map<AssetType, string>([]),
     private readonly monotonicNumInstalledApps: number = 0,
     public readonly rootNonceValue: number = 0
-  ) {}
+  ) {
+    const sortedMultisigOwners = multisigOwners.sort((a, b) =>
+      parseInt(a, 16) < parseInt(b, 16) ? -1 : 1
+    );
+    multisigOwners.forEach((owner, idx) => {
+      if (owner !== sortedMultisigOwners[idx]) {
+        throw new Error(ERRORS.MULTISIG_OWNERS_NOT_SORTED);
+      }
+    });
+  }
 
   public get numInstalledApps() {
     return this.monotonicNumInstalledApps;
@@ -113,7 +124,7 @@ export class StateChannel {
 
   public getAppInstance(appInstanceIdentityHash: string): AppInstance {
     if (!this.appInstances.has(appInstanceIdentityHash)) {
-      throw Error(`${ERRORS.APP_DOES_NOT_EXIST(appInstanceIdentityHash)}`);
+      throw Error(ERRORS.APP_DOES_NOT_EXIST(appInstanceIdentityHash));
     }
     return this.appInstances.get(appInstanceIdentityHash)!;
   }
@@ -145,7 +156,7 @@ export class StateChannel {
     const idx = this.freeBalanceAppIndexes.get(assetType);
 
     if (!this.appInstances.has(idx!)) {
-      throw Error(`${ERRORS.FREE_BALANCE_IDX_CORRUPT(idx!)}`);
+      throw Error(ERRORS.FREE_BALANCE_IDX_CORRUPT(idx!));
     }
 
     const appInstanceJson = this.appInstances.get(idx!)!.toJson();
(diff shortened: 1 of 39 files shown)


packages/machine/src/models/state-channel.ts
```diff
@@ -38,9 +38,14 @@ const ERRORS = {
   FREE_BALANCE_IDX_CORRUPT: (idx: string) =>
     `Index ${idx} used to find ETH Free Balance is broken`,
   INSUFFICIENT_FUNDS:
-    "Attempted to install an appInstance without sufficient funds"
+    "Attempted to install an appInstance without sufficient funds",
+  MULTISIG_OWNERS_NOT_SORTED:
+    "multisigOwners parameter of StateChannel must be sorted"
 };
 
+const sortAddresses = (addrs: string[]) =>
```
cf19drofxots: Can this be a function instead of a `const`?

packages/machine/src/models/state-channel.ts
```diff
@@ -171,9 +183,11 @@ export class StateChannel {
     multisigAddress: string,
     multisigOwners: string[]
   ) {
+    const sortedMultisigOwners = sortAddresses(multisigOwners);
```
cf19drofxots: To remain consistent with the constructor, this should also throw if the given `multisigOwners` is not sorted.

On a side note: setting up a channel is always going to be the first step because the free balance needs to be setup, which involves the `NetworkContext` needing to be passed in, so why not require the constructor to have the network context as a parameter and do the setup in the constructor?

Needing to instantiate a new state channel object then calling setup on it seems like a sub-optimal creation pattern if it can be avoided.

packages/machine/test/unit/protocol/signature-validator.spec.ts
```diff
@@ -44,6 +44,6 @@ describe("Signature Validator Helper", () => {
         commitment,
         signer.signDigest(HashZero.replace("00", "11")) // 0x111111...
       )
-    ).toThrow("Received invalid signature on validateSignature");
+    ).toThrow();
```
cf19drofxots: Why remove the error message? It's handy to give context on why it failed, no?

packages/node/src/node.ts
```diff
@@ -101,58 +111,56 @@ export class Node {
...
-        await this.messagingService.send(context.outbox[0].toAddress, {
-          from: this.address,
-          event: NODE_EVENTS.PROTOCOL_MESSAGE_EVENT,
-          data: context.outbox[0]
-        });
+        const from = getAddress(this.address);
+        const to = getAddress(context.outbox[0].toAddress);
+        const data = context.outbox[0];
+
+        if (context.outbox[0].seq === 1) {
```
cf19drofxots: Can you put a brief comment here to highlight how seq 1 is relevant to needing a `Deferred` promise?

I can see it being confusing if context isn't provided.

packages/node/src/node.ts
```diff
@@ -101,58 +111,56 @@ export class Node {
...
       Opcode.IO_SEND,
       async (message: ProtocolMessage, next: Function, context: Context) => {
-        await this.messagingService.send(context.outbox[0].toAddress, {
-          from: this.address,
-          event: NODE_EVENTS.PROTOCOL_MESSAGE_EVENT,
-          data: context.outbox[0]
-        });
+        const from = getAddress(this.address);
+        const to = getAddress(context.outbox[0].toAddress);
+        const data = context.outbox[0];
```
cf19drofxots: It's also kind of confusing that the thing being sent to the other party isn't the `message`, but rather `context.outbox[0]` which isn't obvious as to what it actually contains.

packages/node/src/node.ts
```diff
@@ -101,58 +111,56 @@ export class Node {
...
-            multisigAddress: AddressZero
-          },
-          signature: {
-            v: -1,
-            r: HashZero,
-            s: HashZero
+      async (message: ProtocolMessage, next: Function, context: Context) => {
+        await new Promise(async (resolve, reject) => {
+          try {
+            const checksumAddress = getAddress(message.toAddress);
```
cf19drofxots: What if multiple messages were sent to this counter party?

How would this differentiate between the different deferred promises using just the address?

packages/node/src/node.ts
```diff
@@ -101,58 +111,56 @@ export class Node {
...
+            const msg = await this.ioSendDeferrals[checksumAddress].promise;
+            context.inbox.push(msg.data);
+            resolve();
+            next();
+          } catch (e) {
+            reject(e);
```
cf19drofxots: Perhaps this error can be a bit more descriptive to provide context about the failure

packages/node/src/node.ts
snario: This actually reminds me of something I was going to do. As-is, this code only works for single round-trip message exchanges; it will not work for the Install Virtual App Protocol.

packages/machine/src/models/state-channel.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `setupChannel` has been changed to a static function

packages/node/src/node.ts
snario: The problem is an overlapping of terminology. The first message refers to the object passed between opcodes in an instruction flow in the middleware-architecture based instruction executor. The second message refers to the NodeMessages received from your counterparty.

packages/node/src/node.ts
snario: Message in this context refers to one inside the machine which is part of the execution of a protocol and has a continuation back into the machine context unlike a regular message.

packages/node/src/node.ts
```diff
@@ -101,58 +111,56 @@ export class Node {
...
-          from: this.address,
-          event: NODE_EVENTS.PROTOCOL_MESSAGE_EVENT,
-          data: context.outbox[0]
-        });
+        const from = getAddress(this.address);
+        const to = getAddress(context.outbox[0].toAddress);
+        const data = context.outbox[0];
+
+        if (context.outbox[0].seq === 1) {
+          this.ioSendDeferrals[to] = new Deferred<
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: there's one place in the `installVirtualApp` protocol that we do two `SEND`s in a row and then immediately end (without any `WAIT` after)

packages/node/src/store.ts
```diff
@@ -45,13 +45,10 @@ export class Store {
       `${this.storeKeyPrefix}/${DB_NAMESPACE_CHANNEL}`
     )) as { [multisigAddress: string]: StateChannelJSON };
 
-    if (!channelsJSON) {
-      console.log("No channels exist yet");
-    } else {
-      for (const entry of Object.entries(channelsJSON)) {
-        channels[entry[0]] = StateChannel.fromJson(entry[1]);
-      }
+    for (const entry of Object.entries(channelsJSON || {})) {
```
joelalejandro: @snario We can avoid using the sub-index accessors by destructuring directly on the iterator:

```ts
for (const [key, value] of Object.entries(channelsJSON)) {
  channels[key] = StateChannel.fromJson(value);
}
```

Also, it might be best to add the `|| {}` fallback directly in the declaration for `channelsJSON`.

packages/node/src/store.ts
```diff
@@ -45,13 +45,10 @@ export class Store {
...
+      channels[entry[0]] = StateChannel.fromJson(entry[1]);
```
joelalejandro: @snario We should have some casing consistency: variable is named `channelsJSON` (all caps), type is named `StateChannelJSON` (also all caps) but the function is called `fromJson` (word-case).

packages/playground-server/src/node.ts
```diff
@@ -60,9 +67,8 @@ export function getNodeAddress(): string {
 export async function createMultisigFor(
   userAddress: string
 ): Promise<NodeTypes.CreateMultisigResult> {
-  if (!node) {
-    node = await createNodeSingleton();
-  }
+  if (!node) node = await createNodeSingleton();
```
joelalejandro: @snario This change seems unnecessary.

packages/playground-server/src/controllers/decorators/validate-signature.ts
```diff
@@ -46,11 +46,13 @@ function validateSignatureMiddleware(
     }
 
     const expectedMessage = await expectedSignatureMessage(resource);
+
```
joelalejandro: @snario This change seems unnecessary.

packages/playground-server/src/controllers/decorators/validate-signature.ts
```diff
@@ -46,11 +46,13 @@ function validateSignatureMiddleware(
...
     const ethAddress = getAddress(
       (((json as unknown) as APIRequest<SessionAttributes>).data as APIResource<
         SessionAttributes
       >).attributes.ethAddress
     );
+
```
joelalejandro: @snario This change seems unnecessary.

packages/playground/src/data/firebase.ts
```diff
@@ -65,35 +65,35 @@ class FirebaseMessagingService implements IMessagingService {
       .set(JSON.parse(JSON.stringify(msg)));
   }
 
-  receive(address: Address, callback: (msg: object) => void) {
-    if (!this.firebase.app) {
+  onReceive(address: Address, callback: (msg: object) => void) {
```
joelalejandro: @snario Same comments as before for this method.

packages/playground-server/src/node.ts
cf19drofxots: It's also generally not good practice to not use braces for these conditional blocks.

packages/machine/src/models/state-channel.ts
cf19drofxots: So _only_ `setupChannel` should be called then? and the constructor is hidden away?

That's what it sounds like

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The constructor is actually publicly exposed but it is not called publicly

packages/node/src/node.ts
cf19drofxots: I understand.

This seems to imply the Node (and by extension the machine) wouldn't be capable for supporting multiple protocol executions with another party. This would be especially concerning once dealing with an intermediary Node whose proxying virtual messages to an end Node.

Though that's the problem for the next PR supporting virtual protocols in the Node<->machine integration.

packages/machine/src/models/state-channel.ts
snario: I think there are cases where we may want a `StateChannel` that has not been set up yet.

packages/node/src/node.ts
snario: I'll implement `IO_SEND_AND_WAIT` in a separate PR.

packages/node/src/deferred.ts
```diff
@@ -0,0 +1,24 @@
+export class Deferred<T> {
```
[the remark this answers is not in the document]
snario: Due to the nature of having two separate middleware hooks, `IO_SEND` and `IO_WAIT`, we need a way to declare a promise and then in resolve it externally. This is the use case of a `Deferred`; to externalize the `resolve` function.

packages/node/src/store.ts
snario: I think perhaps we should do this in a separate PR

PR #570, opened 2019-01-22
[machine] inline computation of dependency nonce related values

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR removes `dependencyNonceSalt` and `dependencyNonceValue` from `MultisigCommitment`

### DIFF
--- a/packages/machine/src/protocol/install.ts
+++ b/packages/machine/src/protocol/install.ts
@@ -147,7 +147,7 @@ function constructInstallOp(
     freeBalance.hashOfLatestState,
     freeBalance.nonce,
     freeBalance.timeout,
-    freeBalance.appSeqNo,
+    app.appSeqNo,
     freeBalance.rootNonceValue
   );
 }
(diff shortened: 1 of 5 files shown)


packages/machine/src/protocol/install.ts
```diff
@@ -147,7 +147,7 @@ function constructInstallOp(
     freeBalance.hashOfLatestState,
     freeBalance.nonce,
     freeBalance.timeout,
-    freeBalance.appSeqNo,
+    app.appSeqNo,
```
snario: I think this was meant to be `freeBalance.appSeqNo`.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Why?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It goes into `conditionalTransactionInput`

PR #582, opened 2019-01-23
[machine, node] Change IO_WAIT to IO_SEND_AND_WAIT

snario:
Slightly more easy to reason about.

### DIFF
--- a/packages/machine/src/protocol/install-virtual-app.ts
+++ b/packages/machine/src/protocol/install-virtual-app.ts
@@ -44,6 +44,7 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
       });
     },
 
+    // wait for M5
     Opcode.IO_SEND_AND_WAIT
   ],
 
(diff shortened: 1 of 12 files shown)


packages/machine/src/protocol/install-virtual-app.ts
```diff
@@ -112,10 +110,7 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
       context.outbox[0].signature2 = context.signature2; // s7
     },
 
-    Opcode.IO_SEND,
-
-    // wait for M4
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: We should keep this comment

packages/node/src/node.ts
```diff
@@ -226,8 +229,8 @@ export class Node {
    *     solely to the deffered promise's resolve callback.
    */
   private async handleReceivedMessage(msg: NodeMessage) {
-    if (!NODE_EVENTS[msg.event]) {
-      console.error(`Received message with unknown event type`);
+    if (!Object.values(NODE_EVENTS).includes(msg.event)) {
```
joelalejandro: I hate that TypeScript forces the use of Object.values for such a trivial task :(OI hat

snario: Yeah I noticed with the other change it was always evaluating to `true` so have to revert

joelalejandro: Yeah, I re-read it and sorry 'bout that, I was talking non-sense, lol.

PR #583, opened 2019-01-23
[apps] Adds migrations of the apps' contracts to Ropsten

cf19drofxots:
### Description

This adds the migration data (primarily deployed contract addresses) on Ropsten.

It also adds the tooling used to do said migration.


cf19drofxots:
@snario we _could_ but what's the added benefit for our use case?

snario:
> @snario we _could_ but what's the added benefit for our use case?

Based on `ethers` so the code is cleaner; has all the same functionality and features; much smaller dependency

cf19drofxots:
I spent a little bit of time to see if refactoring to use etherlime is super trivial.

It's not quite and it'll take a bit more time, so will come back around to it later - not a priority atm.

packages/apps/README.md
```diff
@@ -0,0 +1,19 @@
...
+| ---------- | --------------- |
+| 1          | Main net        |
+| 3          | Ropsten testnet |
+| 4          | Rinkeby testnet |
+| 42         | Kovan testnet   |
+
+Not all of the networks will be used for the Counterfactual contracts, but you can find a more comprehensive list [here](https://ethereum.stackexchange.com/a/17101).
+To run a migration against a target network:
+
+- make sure the target network configuration exists in `truffle.js`
```
Alonski: Should be `truffle-config.js`?

cf19drofxots: It should be both! good catch

cf19drofxots:
Have to rerun migration since https://github.com/counterfactual/monorepo/pull/584 landed before this PR

PR #585, opened 2019-01-23
no more connecting to undefined

patience-tema-baron:
screenshot from 2019-01-23 11-28-18
 as it should be


packages/playground/src/components/widgets/widget-connection/widget-connection.tsx
```diff
@@ -18,6 +18,7 @@ export class WidgetConnection {
   @Prop() network: string = "";
 
   render() {
+    console.log("network", this.network)
```
Alonski: Is this console.log needed?

patience-tema-baron: Definitely not. Just took it out in the latest commit.

PR #587, opened 2019-01-23
[WIP] [cf.js] Replace node-uuid with util class

Alonski:
### Description

This pretty much just copies the `uuid` code for `v4` into a utility function.

### DIFF
--- a/packages/cf.js/src/provider.ts
+++ b/packages/cf.js/src/provider.ts
@@ -6,7 +6,6 @@ import {
   Node
 } from "@counterfactual/types";
 import EventEmitter from "eventemitter3";
-import { v4 as generateUUID } from "uuid";
 
 import { AppInstance, AppInstanceEventType } from "./app-instance";
 import {
@@ -16,6 +15,8 @@ import {
   UninstallEventData,
   UpdateStateEventData
 } from "./types";
+// import { v4 as generateUUID } from "uuid";
+import generateUUID from "./utils/uuid";
 
 /**
  * Milliseconds until a method request to the Node is considered timed out.
--- a/packages/cf.js/src/utils/uuid.ts
+++ b/packages/cf.js/src/utils/uuid.ts
@@ -1,0 +1,60 @@
+/* tslint:disable */
+const rng = crypto.getRandomValues(new Uint8Array(16));
+
+var byteToHex = [];
+for (var i = 0; i < 256; ++i) {
+  byteToHex[i] = (i + 0x100).toString(16).substr(1);
+}
+
+function bytesToUuid(buf) {
+  var i = 0;
+  var bth = byteToHex;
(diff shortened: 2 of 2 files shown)


packages/cf.js/src/utils/uuid.ts
```diff
@@ -0,0 +1,60 @@
+/* tslint:disable */
+const rng = crypto.getRandomValues(new Uint8Array(16));
```
snario: We assume `crypto` is in scope?

Is it in scope in the browser?

packages/cf.js/src/utils/uuid.ts
```diff
@@ -0,0 +1,60 @@
+/* tslint:disable */
```
snario: I think this whole file can be linted. e.g., `var` vs `const` and typings etc.

Alonski: I had issues with linting that I didn't really want to start getting into. This is highly tested JS code :/ If its important I can look into it tomorrow

packages/cf.js/src/provider.ts
```diff
@@ -16,6 +15,8 @@ import {
   UninstallEventData,
   UpdateStateEventData
 } from "./types";
+// import { v4 as generateUUID } from "uuid";
```
cf19drofxots: Remove instead of commenting out?

packages/cf.js/src/utils/uuid.ts
Alonski: Supported in modern browsers: https://caniuse.com/#feat=getrandomvalues

packages/cf.js/src/provider.ts
Alonski: Yes I will

cf19drofxots:
`uuid` has an MIT license, so we should reference the original source and the license in the copy-over.

ebryn:
we shouldn't be doing this @Alonski

packages/cf.js/src/utils/uuid.ts
Alonski: For example this: `[i++]` throws `[tslint] Forbidden ++ operator [no-increment-decrement]`

snario: Ah.

Yeah maybe we should make it a `.js` file then and import it that way?

Alonski:
@ebryn So what do you suggest? Just use something like Math.random?
@cf19drofxots @snario Do collisions matter?

ebryn:
@Alonski i was thinking Date.now or performance.now (for finer grain timestamps) or Math.random

packages/cf.js/src/utils/uuid.ts
Alonski: Removed this file so resolving

PR #589, opened 2019-01-23
Move STATE_TRANSITION_COMMIT to Node

snario:


packages/node/src/node.ts
```diff
@@ -2,8 +2,10 @@ import {
   Context,
   InstructionExecutor,
   Opcode,
+  Protocol,
   ProtocolMessage
 } from "@counterfactual/machine";
+import { SetupParams } from "@counterfactual/machine/dist/src/protocol-types-tbd";
```
cf19drofxots: This should be exported/imported properly (like above it).

PR #591, opened 2019-01-23
[playground] Ptb/login error

patience-tema-baron:
Adds a login error message with a tooltip.

screenshot from 2019-01-23 14-07-18


patience-tema-baron:
Tests are now passing! Could I get a review @ebryn @joelalejandro @Alonski?

PR #593, opened 2019-01-23
[node] Syncs cf.js and Node message formats

cf19drofxots:


cf19drofxots:
Some npm & networking issues seem to be preventing the build step from passing. Will try again in a few hours.

PR #596, opened 2019-01-24
[cf.js][node] Spec for proposeState

cf19drofxots:
### Description

This adds the spec for making a state proposal call for an AppInstance in a channel.

### Related issues
`acceptState` and `rejectState` will similarly follow this.


cf19drofxots:

PR #597, opened 2019-01-24
[cf.js][node] Spec for acceptState

cf19drofxots:
### Description

This adds the spec for making a state acceptance call for an AppInstance who has an existing state proposal received from another party.


cf19drofxots:

PR #598, opened 2019-01-24
[cf.js][node] Spec for rejectState 

cf19drofxots:
### Description

This adds the spec for making a state rejection call for an AppInstance who has an existing state proposal received from another party.

### DIFF
--- a/packages/cf.js/API_REFERENCE.md
+++ b/packages/cf.js/API_REFERENCE.md
@@ -324,7 +324,15 @@ TODO
 
 ### Method: `rejectState`
 
-TODO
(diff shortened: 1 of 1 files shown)


packages/cf.js/API_REFERENCE.md
```diff
@@ -324,7 +324,15 @@ TODO
 
 ### Method: `rejectState`
 
-TODO
+Rejects a proposed state to an AppInstance and notifies the counterparty of
+said rejection.
+
+  Params:
+- `appInstanceId: string`
```
snario: Shouldn't you pass in like, the hash of the state that was proposed, so you can match it up

cf19drofxots:

PR #599, opened 2019-01-24
[nodeProvider] Update dependencies & rollup

cf19drofxots:
This updates the dist to bundle eventemitter with it and removes the unused cf.js package dependency.


ebryn:
I think we should just be bundling NodeProvider into CF.js?

cf19drofxots:
yeah we should be

PR #600, opened 2019-01-24
[node] Simplify event handler for INSTALL & de-couple store method

snario:


packages/node/src/methods/app-instance/take-action/operation.ts
```diff
@@ -31,19 +31,25 @@ async function makeApplyActionCall(
...
   } catch (e) {
-    const sanitizedError = e
-      .toString()
-      .replace("s: VM Exception while processing transaction: revert");
-    return Promise.reject(`${ERRORS.INVALID_ACTION}: ${sanitizedError}`);
+    if (
+      e.toString().indexOf("s: VM Exception while processing transaction:") !==
+      -1
+    ) {
+      return Promise.reject(`${ERRORS.INVALID_ACTION}: ${e.toString()}`);
```
cf19drofxots: You still wanna inform the user that the action is invalid, instead of throwing a not-as-clear message at them.

packages/node/src/store.ts
```diff
@@ -157,22 +157,15 @@ export class Store {
...
    * @param appInstanceInfo
    */
-  public async updateChannelWithAppInstanceInstallation(
-    stateChannel: StateChannel,
+  public async saveRealizedProposedAppInstance(
     appInstanceIdentityHash: string,
     appInstanceInfo: AppInstanceInfo
   ) {
     await this.storeService.set([
-      {
```
cf19drofxots: Can we change this behavior in another PR?

This needs to be looked at more closely in isolation.

PR #620, opened 2019-01-28
[node-provider][cf.js] Bundle up node provider with cf.js

cf19drofxots:
### Description

This bundles up @counterfactual/node-provider along with cf.js.

### DIFF
--- a/packages/apps/networks/3.json
+++ b/packages/apps/networks/3.json
@@ -11,8 +11,8 @@
   },
   {
     "contractName": "HighRollerApp",
-    "address": "0x6296F3ACf03b6D787BD1068B4DB8093c54d5d915",
-    "transactionHash": "0x99e53351e3161be34addad301a073380250155b2c60a183d17132677000acd65"
+    "address": "0x903217387B06a84F4dD0bEA565Ad8765Fc7cAA58",
+    "transactionHash": "0xc0ca508261cdb2401542a8cacf58e3430cbe0ec669887cee1fe98023ac7894a1"
   },
   {
     "contractName": "NimApp",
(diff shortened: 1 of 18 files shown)


cf19drofxots:
Marking as `DO NOT MERGE` until the refactoring is complete so as to not introduce regressions.

Alonski:
@joelalejandro @patience-tema-baron Can you take a look at the changes regarding using `cf.NodeProvider` instead of `NodeProvider` in the dApps to make sure everything looks good?

packages/dapp-high-roller/src/data/types.ts
```diff
@@ -335,4 +335,19 @@ export namespace cf {
     appInstances: { [appInstanceId: string]: AppInstance };
     nodeProvider: NodeProvider;
   };
+  export type NodeProvider = {
```
cf19drofxots: Can this package not have `nodeProvider` as a dependency so that type matching doesn't have to be exported like this?

Alonski: This file shouldn't exist. We couldn't get `@counterfactual/types` imported into the Stencil apps. Maybe now it is fixed and we should try again

Alonski:
@cf19drofxots This is also probably caused by type mismatches. Because I am mocking the types it doesn't realize that this has a constructor.

joelalejandro:
@cf19drofxots @Alonski I've pushed 7a8e8c0 to fix this. Basically what was needed was some Jest mocks with the proper signature.

Alonski:
@ebryn should I squash merge this?

PR #622, opened 2019-01-29
[machine] virtual app set state integration test

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR adds an integration test for `virtual-app-set-state-commitment`


packages/contracts/test/app-registry.spec.ts
```diff
@@ -50,7 +50,9 @@ describe("AppRegistry", () => {
     provider = waffle.createMockProvider();
     wallet = (await waffle.getWallets(provider))[0];
 
-    appRegistry = await waffle.deployContract(wallet, AppRegistry);
+    appRegistry = await waffle.deployContract(wallet, AppRegistry, [], {
+      gasLimit: 6000000
```
snario: Why is this needed?

packages/machine/test/integration/virtual-app-set-state-commitment.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+let appRegistry: Contract;
+
+let signingKeys: SigningKey[];
+let targetAppInstance: AppInstance;
+let c0: VirtualAppSetStateCommitment;
+let s0: Signature;
+
+expect.extend({ toBeEq });
+
+const rejectionOf = (promise: Promise<any>) =>
```
snario: `ethereum-waffle` offers a matcher: `await expect(<transaction).to.revertWith(...)`

packages/machine/test/integration/virtual-app-set-state-commitment.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    );
+  });
+  it("fails with nonce above expiry", async () => {
+    const c1 = new VirtualAppSetStateCommitment(
+      network,
+      targetAppInstance.identity,
+      65536,
+      targetAppInstance.timeout,
+      targetAppInstance.hashOfLatestState,
+      65537
```
snario: There are several magic numbers in here, can you make them more clear w.r.t. intent

packages/machine/test/integration/virtual-app-set-state-commitment.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    const contractAppState = await appRegistry.appStates(
+      targetAppInstance.identityHash
+    );
+
+    expect(contractAppState.appStateHash).toBe(
+      targetAppInstance.hashOfLatestState
+    );
+  });
+  it("fails with nonce above expiry", async () => {
+    const c1 = new VirtualAppSetStateCommitment(
```
snario: Better variable name than `c1`?

packages/contracts/test/app-registry.spec.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: the default gas limit is fixed at 4 million (it is not estimated) which is too low to deploy `AppRegistry`

snario: Worth noting this in a comment in the code imo

packages/machine/test/integration/virtual-app-set-state-commitment.spec.ts
snario: You can still import `solidity` from `ethereum-waffle` and use `chai.use(solidity)`

snario: Waffle is not a testing framework. Waffle's helper functions work well when used within `mocha` and has issues with `jest` simply because of the asynchronicity of it I believe. It might work with `jest` with `runInBand` on true.

In any case, you don't need to care about this to use the matchers.

PR #639, opened 2019-01-31
[playground] [playground-server] [dapp-high-roller] [dapp-tic-tac-toe] Adds support for "playground:matchmakeWith" flag

joelalejandro:
This PR introduces a new flag called `playground:matchmakeWith`, useful for dev/testing purposes.

In order to be used, this flag must be set in the Playground's `localStorage`, like this:

image

This flag will be passed to the dApp as part of the `playground:request:user` message response. If the flag is set, it'll be then passed as part of the request to the Matchmaking API:

```json
{ "data": { "attributes": { "matchmakeWith": "c" } } }
```

When the Playground Server parses this request, it'll return the specified user's data, instead of running the matchmaking algorithm.

if the flag is not set, matchmaking will work as usual.

### DIFF
--- a/packages/dapp-high-roller/src/components/app-wager/app-wager.tsx
+++ b/packages/dapp-high-roller/src/components/app-wager/app-wager.tsx
@@ -162,12 +162,20 @@ export class AppWager {
     }
 
     const { token } = this.account.user;
+    const { matchmakeWith } = this.account;
+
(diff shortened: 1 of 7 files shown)


cf19drofxots:
🙏 

now I don't have to manually modify the db query filter

PR #645, opened 2019-02-01
[cf.js] validate event types in cf.js

patience-tema-baron:
Throws an error if developers try to listen to events outside of the EventType type.


ebryn:
@patience-tema-baron could you add tests?

patience-tema-baron:
@ebryn Definitely. I'll be out tomorrow morning, but can add some in the afternoon.

PR #653, opened 2019-02-01
Implement Extended Public Key scheme for app-specific signingKeys

snario:
~Waiting on an ethers.js upgrade but this is roughly what it looks like.~

# Notes

## Introduction of extended public keys

We use an unreleased version of ethers.js to use upgrades to the `HDNode` object on `ethers`.

This gives us the ability to make the following changes:

- Node public addresses (`0xabd...`) are now extended public keys (`xpubabd...`)
- `multisigOwners` on the `StateChannel` object are now `userExtendedKeys`, storing the extended public key
- The keys on the multisig are the 0th children of the extended keys
- The signing keys for app with seqNo == k are the k-th children of the extended keys

## TODO

- [x] Update the way `PRIVATE_KEY` was used inside Playground
- [x] Wait for ethers.js to update today
- [ ] Make consistent across the specs too, probably updating message params `initiatingAddress` and `respondingAddress` to be xpub types


packages/node/src/events/install/controller.ts
```diff
@@ -67,14 +67,14 @@ function createAppInstanceFromAppInstanceInfo(
   return new AppInstance(
     channel.multisigAddress,
     // TODO: generate ephemeral app-specific keys
```
cf19drofxots: This TODO can be removed as of this PR.

packages/node/src/events/install/controller.ts
```diff
@@ -67,14 +67,14 @@ function createAppInstanceFromAppInstanceInfo(
...
-    channel.multisigOwners,
+    channel.getSigningKeysFor(channel.numInstalledApps - 1),
     proposedAppInstanceInfo.timeout.toNumber(),
     appInterface,
     terms,
     // TODO: pass correct value when virtual app support gets added
     false,
     // TODO: this should be thread-safe
-    channel.numInstalledApps + 1,
+    channel.numInstalledApps - 1,
```
cf19drofxots: Why -1?

If it's the first AppInstance being installed, 0 is already taken by the ETHFreeBalance, so shouldn't this be +1?

packages/node/src/node.ts
```diff
@@ -118,9 +119,13 @@ export class Node {
           );
         }
 
+        const signingKey = new SigningKey(
+          this.signer.derivePath("0").privateKey
```
cf19drofxots: Shouldn't this be the extended signing key for the AppInstance that this commitment is about?

packages/apps/package.json
```diff
@@ -26,7 +26,7 @@
     "@types/chai": "^4.1.7",
     "@types/mocha": "^5.2.5",
     "chai": "^4.2.0",
-    "ethers": "4.0.23",
+    "ethers": "https://github.com/ethers-io/ethers.js.git",
```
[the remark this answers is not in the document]
snario: Not on npm yet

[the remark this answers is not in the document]
cf19drofxots: Ah I thought he had a hook to publish to npm whenever the dist file gets updated

https://github.com/ethers-io/ethers.js/commits/master

packages/node/src/events/install/controller.ts
snario: This is an issue that will get fixed in `liam/virtualapp`. Basically this function is creating a "representation" of an `AppInstance`. However, the `channel` variable is the `StateChannel` _after_ the "real" `AppInstance` has already been installed. 

This helper function will be removed in my other PR because it is confusing and unnecessary.

packages/node/src/node.ts
snario: So, we actually sign every commitment using the 0th key (_except_ for `UpdateProtocol` commitments, in which case you're right this was a mistake on my part). However, the Update Protocol could also just make a call to the multisig instead of to the `AppRegistry` then we wouldn't need to change this code.

snario: Because every commitment except `SetStateCommitment` has `to` being `multisigAddress` and so we verify the signature based on `multisigOwners` which is the 0th key.

cf19drofxots: So the public signing keys used per AppInstance are unique, but are derived from a higher xkey that can be used to sign commitments, resulting in a signature whose recovered address is the Node's xpub address.

Right?

PR #661, opened 2019-02-04
[machine] create a channel for target virtual app

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
create a channel for target virtual app to live in

### DIFF
--- a/packages/machine/src/protocol/install-virtual-app.ts
+++ b/packages/machine/src/protocol/install-virtual-app.ts
@@ -16,7 +16,8 @@ import {
   ETHVirtualAppAgreementInstance,
   StateChannel
 } from "../models";
-import { Context, InstallVirtualAppParams, ProtocolMessage } from "../types";
+import { Context, InstallVirtualAppParams, ProtocolMessage, SolidityABIEncoderV2Struct } from "../types";
+import { virtualAppKey } from "@counterfactual/machine/src/virtual-app-key";
 
 // hardcoded assumption: all installed virtual apps can go through this many update operations
 const NONCE_EXPIRY = 65536;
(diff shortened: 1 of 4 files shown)


packages/machine/src/protocol/install-virtual-app.ts
```diff
@@ -1,11 +1,7 @@
 import { ETHVirtualAppAgreementCommitment } from "@counterfactual/machine/src/ethereum/eth-virtual-app-agreement-commitment";
 import { VirtualAppSetStateCommitment } from "@counterfactual/machine/src/ethereum/virtual-app-set-state-commitment";
-import {
-  AppInterface,
-  AssetType,
-  NetworkContext,
-  SolidityABIEncoderV2Struct
-} from "@counterfactual/types";
+import { virtualAppKey } from "@counterfactual/machine/src/virtual-app-key";
```
snario: Should export it from the module vs diving into the code

packages/machine/src/protocol/install-virtual-app.ts
```diff
@@ -142,6 +143,25 @@ function createTarget(
...
+  respondingAddress: string,
+  intermediaryAddress: string,
+  targetAppInstance: AppInstance
+) {
+  const key = virtualAppKey(
+    [initiatingAddress, respondingAddress],
+    intermediaryAddress
+  );
+  let sc = context.stateChannelsMap.get(key);
+  if (sc === undefined) {
```
snario: You can use the `.has` method instead of checking if `.get` returns `undefined`

packages/machine/src/virtual-app-key.ts
```diff
@@ -0,0 +1,11 @@
+import { keccak256 } from "ethers/utils/solidity";
+
+export function virtualAppKey(users: string[], intermediary: string) {
```
snario: Can you explain the usage / requirement of this?

packages/machine/src/protocol/install-virtual-app.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I think `.has` is not the best since it we need to do the `.get` in any case

packages/machine/src/protocol/install-virtual-app.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: it is exported - fixing the import path

packages/machine/src/virtual-app-key.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: as discussed, one place to put virtual app target app instances is a separate state channel. this function computes the key for that channel.

other possible approaches:
- have `virtualAppKey` be the `create2` address of a multisig with `users + [intermediary]` as owners. I do not like this approach because it means that this function changes if we upgrade the multisig code. furthermore, it is confusing because we explicitly never deploy a multisig currently. we already have a field on statechannel for the owner that can be set to the `create2` address in the future if needed
- change things so that channels are not forced to be keyed by a multisig address. I think this is the right approach, however it involves more work, for e.g., changing the store

PR #686, opened 2019-02-07
[node] Refactor node tests for better ganache/contracts handling support

cf19drofxots:
### Description

This refactors how the test environment gets setup to better support ganache / contract dependencies.

Since jest can't pass state between the `globalSetup` script and individual test suite, the workaround is to deploy the needed contracts in this script and write their addresses to a temp file that gets read by the `testEnvironment` script (which is sandboxed per test suite).

This will better support the upcoming PR for testing Node functionality when proper multisig contracts need to be deployed.

Caveat: As of Jest 24, `transform` can be run in `global{Setup, Teardown}` but not on the `testEnvironment` file, meaning the firebase service factory can't be used in the `testEnvironment` script (as it's TS, and the `testEnvironment` script is JS).


packages/node/package.json
```diff
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "build": "rollup -c",
-    "test": "jest --setupFiles dotenv-extended/config --runInBand --detectOpenHandles --bail",
```
cf19drofxots: This is intended until a follow up PR investigates proper closing of handles.

packages/node/package.json
```diff
@@ -52,6 +53,9 @@
         "statements": 90
       }
     },
+    "globalSetup": "<rootDir>/test/global-setup.jest.ts",
```
snario: Maybe extract this stuff into `jest.config.js` if its getting more complicated

packages/node/src/services.ts
```diff
@@ -46,6 +46,10 @@ export class FirebaseServiceFactory {
   createStoreService(storeServiceKey: string): IStoreService {
     return new FirebaseStoreService(this.app.database(), storeServiceKey);
   }
+
+  async close() {
```
snario: Maybe `closeConnection` or `disconnect` or `disconnectAndDeleteHistory` (does it delete history?)

packages/node/test/global-setup.jest.ts
```diff
@@ -0,0 +1,63 @@
...
+import dotenvExtended from "dotenv-extended";
+import { ContractFactory, Wallet } from "ethers";
+import { Web3Provider } from "ethers/providers";
+import fs from "fs";
+import ganache from "ganache-core";
+import mkdirp from "mkdirp";
+import os from "os";
+import path from "path";
+
+dotenvExtended.load();
```
snario: Isn't the `--setupFiles dotenv-extended/config` thing supposed do this?

packages/node/test/global-setup.jest.ts
```diff
@@ -0,0 +1,63 @@
...
+        balance: "120000000000000000",
+        secretKey: process.env.PRIVATE_KEY_B
+      },
+      {
+        balance: "120000000000000000",
+        secretKey: process.env.PRIVATE_KEY_C
+      }
+    ]
+  });
+  // @ts-ignore
```
snario: Reasoning behind this?

packages/node/test/global-setup.jest.ts
```diff
@@ -0,0 +1,63 @@
...
+  const proxyFactoryContract = await new ContractFactory(
+    ProxyFactory.abi,
+    ProxyFactory.bytecode,
+    wallet
+  ).deploy();
+  const tttContract = await new ContractFactory(
+    TicTacToeApp.interface,
+    TicTacToeApp.bytecode,
+    wallet
+  ).deploy();
```
snario: Perhaps this counterfactual-specific stuff should be in a separate file than the ganache stuff? Just to keep this file simple and easy to grok

packages/node/test/node-test-environment.jest.js
```diff
@@ -0,0 +1,31 @@
...
+    super(config);
+  }
+
+  async setup() {
+    await super.setup();
+    const addresses = readFileSync(path.join(DIR, "addresses"), "utf8");
+    if (!addresses) {
+      throw new Error("Contract addresses not found");
+    }
+    this.global.contractAddresses = JSON.parse(addresses);
```
snario: Maybe stay consistent with the `networkContext` variable name

packages/node/test/global-setup.jest.ts
cf19drofxots: It apparently doesn't work on setup files.

I assumed it did on my first try but the values ended up being `undefined`.

packages/node/test/global-setup.jest.ts
cf19drofxots: TS complains that a property being set on the `Global` object doesn't exist (`Property 'ganacheServer' does not exist on type 'Global'`) even though expected behavior for Jest is to be able to set properties like this.

So ignoring the false TS warning.

PR #687, opened 2019-02-07
[node] Multisig contract deployment

cf19drofxots:
### Description

This adds functionality to deploy a multisig wallet contract when opening a channel. It's a replacement 
for https://github.com/counterfactual/monorepo/pull/634 as that's somewhat stale, even though that PR can effectively be treated as a spike to get end-to-end functionality of this feature working.

~Also note the base branch for this PR, as that branch's commits enable easier testing for this feature.~


cf19drofxots:
@joelalejandro so the playground server can almost entirely use the multisig deployment feature of the Node, except the way it's written right now, there's no easy way to say "spin up server and instantiate this node inside it so that I can this node again"

at least it seems that way from the way the tests are written because when a call comes in to create a new channel, a new node gets created with a random private key, meaning it won't have a funded account to execute the contract deployment transaction (note: the node that gets created as part of the test setup is given the private key to a funded account, which is the Node that should be used internally in the playground-server)

```
      JSONAPI-TS:  Failed to create channel. Multisignature wallet cannot be deployed properly: Error: insufficient funds (version=4.0.24)
```

The insufficient funds is because the Node being used to deploy is the Node that gets created here: https://github.com/counterfactual/monorepo/blob/master/packages/playground-server/src/node.ts#L71

so we need to ensure this is in fact a singleton

cc @ebryn

PR #713, opened 2019-02-08
[playground] add netlify rewrite rules

patience-tema-baron:
Based on what I've read, this _should_ work, but we've got to get in on staging to verify.


patience-tema-baron:
Looks like the netlify part works, though several of the routes are still breaking due to playground being in a weird state if it loads from somewhere other than index.html. Let's hold off on merging until I work that stuff out.

patience-tema-baron:
Reloading should now work!

PR #714, opened 2019-02-08
[playground] [node] [playground-server] Refactor UI modals for ProposeInstall / RejectInstall

joelalejandro:
This PR applies the necessary upgrades to make the UI modals for ProposeInstall / RejectInstall show all relevant data, following the design spec.

Since having a "global" component is not supported by Stencil, the `node-listener` component is now a wrapper for the components that handle the `/` and `/dapp/:dappName` routes. The Node instance is wrapped in a static singleton class.

It also adds an `.off()` method to the Node, to prevent event re-subscription.

**Preview**

deepin-screen-recorder_select area_20190208183154


packages/playground-server/src/db.ts
```diff
@@ -125,8 +125,22 @@ export async function matchmakeUser(userToMatch: User): Promise<MatchedUser> {
   });
 }
 
-export async function getUsers(filters: {}): Promise<User[]> {
+export async function getUsers(
+  filters: {},
+  fields: string[] = []
+): Promise<User[]> {
   const db = getDatabase();
+  let returnFields = fields;
```
cf19drofxots: Why not just name the function parameter `returnFields`?

joelalejandro: Because it's not a good practice to reassign value parameters.

PR #719, opened 2019-02-09
[monorepo] Rename initiating and responding in proposal messages

snario:
The fact that `initiatingAddress` and `respondingAddress` were the same variable names for _proposals_ as they are for _actual protocol messages_ was confusing. This makes them explicitly different. Also changed some stale `Address` types to `string` since `xpub...` is not an address.

### DIFF
--- a/packages/node/src/events/propose-install-virtual/controller.ts
+++ b/packages/node/src/events/propose-install-virtual/controller.ts
@@ -21,7 +21,7 @@ export default async function proposeInstallVirtualEventController(
     requestHandler.store,
     nodeMsg.data.params,
     nodeMsg.data.appInstanceId,
-    nodeMsg.data.initiatingAddress,
+    nodeMsg.data.proposedByAddress,
     nodeMsg.from!
   );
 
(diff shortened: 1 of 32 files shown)


packages/node/src/events/propose-install-virtual/controller.ts
```diff
@@ -21,7 +21,7 @@ export default async function proposeInstallVirtualEventController(
     requestHandler.store,
     nodeMsg.data.params,
     nodeMsg.data.appInstanceId,
-    nodeMsg.data.initiatingAddress,
+    nodeMsg.data.proposedByAddress,
```
cf19drofxots: Let's stay consistent with the `Identifier` suffix.

packages/node/src/events/propose-install/operation.ts
```diff
@@ -1,18 +1,21 @@
-import { Address, Node } from "@counterfactual/types";
+import { Node } from "@counterfactual/types";
 
 import { ProposedAppInstanceInfo } from "../../models";
 import { Store } from "../../store";
 import { getChannelFromPeerAddress } from "../../utils";
 
 export async function setAppInstanceIDForProposeInstall(
-  selfAddress: Address,
+  selfAddress: string,
```
cf19drofxots: Change variable naming I suppose since it's not an address.

packages/node/src/events/propose-install-virtual/controller.ts
snario: Oh actually this is a bug. Fixing.

packages/node/src/events/propose-install/operation.ts
snario: FIxing.

PR #607, opened 2019-01-25
[node] Install Virtual App Integration Node<>Machine

snario:

PR #720, opened 2019-02-09
[node] Removes uuid appInstanceId distinction

snario:


snario:
Closing this PR because the tests fail but it is because the implementation on `master` is incorrect at the moment. Instead, going to rebase #607 on top of this PR and then merge in that PR instead of fixing the tests on this PR (which would be a worthless effort).

PR #727, opened 2019-02-11
[playground] Use non-zero gas fee for transaction

cf19drofxots:
### Description

This estimates how much gas to use for the deposit transaction.

### DIFF
--- a/packages/playground/src/components/account/account-deposit/account-deposit.tsx
+++ b/packages/playground/src/components/account/account-deposit/account-deposit.tsx
@@ -16,6 +16,7 @@ export class AccountDeposit {
   @Prop() updateAccount: (e) => void = e => {};
   @Prop() history: RouterHistory = {} as RouterHistory;
   @Prop() signer: Signer = {} as Signer;
+  @Prop() provider: Web3Provider = {} as Web3Provider;
 
   @State() error: string = "";
   @State() amountDeposited;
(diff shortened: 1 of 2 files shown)


packages/typescript-typings/types/ethers/index.d.ts
```diff
@@ -13,6 +13,7 @@ declare class Web3Provider {
   getBalance(address: string): Promise<BigNumber>;
   on(event: string, callback: (...args) => void): void;
   removeAllListeners(event: string): void;
+  estimateGas(tx): Promise<any>;
```
snario: I'd just remove this entire folder altogether, its outdated

packages/playground/src/components/account/account-deposit/account-deposit.tsx
```diff
@@ -30,9 +31,14 @@ export class AccountDeposit {
...
-      await this.signer.sendTransaction({
+      const tx = {
         to: this.user.multisigAddress,
         value: this.amountDeposited
+      };
+      const gasEstimate = await this.provider.estimateGas(tx);
+      await this.signer.sendTransaction({
+        ...tx,
+        gasPrice: gasEstimate
       });
```
snario: Probably will look cleaner to just inline all of these vars

packages/typescript-typings/types/ethers/index.d.ts
joelalejandro: @snario It's not relevant at this moment, and also until Ethers support ES modules (ethers-io/ethers.js#114) we can't import it properly into Stencil.

cf19drofxots: Well apparently it's still being used...

cc @joelalejandro

packages/playground/src/components/account/account-deposit/account-deposit.tsx
joelalejandro: @snario Doesn't seem like a blocker.

PR #730, opened 2019-02-11
[playground] Fix provider syntax

cf19drofxots:


snario:
Pretty sure `packages/typescript-typings/types/ethers/index.d.ts` can be deleted as it was added before `ethers` had typescript support.

joelalejandro:
@snario You're half-right. It does support TypeScript, but we cannot import `ethers` into Stencil because of the issue with ES Modules I've mentioned earlier.

PR #733, opened 2019-02-11
[machine, node] UninstallVirtualApp implementation

snario:


packages/machine/src/models/state-channel.ts
```diff
@@ -114,7 +114,15 @@ export class StateChannel {
     > = new Map<AssetType, string>([]),
     private readonly monotonicNumInstalledApps: number = 0,
     public readonly rootNonceValue: number = 0
-  ) {}
+  ) {
+    userNeuteredExtendedKeys.forEach(xpub => {
+      if (xpub.slice(0, 4) !== "xpub") {
+        throw Error(
+          `StateChannel constructor given invalid extended keys: ${userNeuteredExtendedKeys}`
```
cf19drofxots: Why not point out the bad key that was provided instead of all of them?

packages/machine/src/models/state-channel.ts
```diff
@@ -388,14 +416,31 @@ export class StateChannel {
...
-      this.multisigOwners,
+      this.userNeuteredExtendedKeys,
       this.appInstances,
       ethVirtualAppAgreementInstances,
       this.freeBalanceAppIndexes,
       this.monotonicNumInstalledApps
     ).incrementFreeBalance(AssetType.ETH, increments);
   }
 
+  public removeVirtualApp(targetIdentityHash: string) {
```
cf19drofxots: Add return type to fn signature

packages/node/src/events/protocol-message/controller.ts
```diff
@@ -25,4 +30,18 @@ export default async function protocolMessageEventController(
...
+    const uninstallMsg: UninstallMessage = {
+      from: requestHandler.publicIdentifier,
+      type: NODE_EVENTS.UNINSTALL_VIRTUAL,
+      data: {
+        appInstanceId: (nodeMsg.data.params as UninstallVirtualAppParams)
+          .targetAppIdentityHash
+      }
+    };
+
+    requestHandler.outgoing.emit(uninstallMsg.type, uninstallMsg);
```
cf19drofxots: So this event is being sent out twice?

It's also being sent out outside of being handled here, right?

packages/node/src/methods/app-instance/uninstall/operation.ts
```diff
@@ -9,16 +9,18 @@ export async function uninstallAppInstanceFromChannel(
...
 
   const stateChannel = await store.getChannelFromAppInstanceID(appInstanceId);
 
   const appInstance = stateChannel.getAppInstance(appInstanceId);
 
+  const currentChannels = new Map(Object.entries(await store.getAllChannels()));
+
   const stateChannelsMap = await instructionExecutor.runUninstallProtocol(
-    new Map(Object.entries(await store.getAllChannels())),
+    currentChannels,
```
cf19drofxots: Why not inline? `currentChannels` doesn't seem to be being used anywhere else

packages/node/src/store.ts
```diff
@@ -206,6 +206,10 @@ export class Store {
   public async addVirtualAppInstanceProposal(
     proposedAppInstance: ProposedAppInstanceInfo
   ) {
+    const sortedXpubs = [
```
cf19drofxots: Why not use `sortAddresses` from xkeys.ts?

packages/node/test/integration/uninstall-virtual.spec.ts
```diff
@@ -25,7 +25,7 @@ import {
 } from "./utils";
 
 describe("Node method follows spec - uninstall", () => {
-  jest.setTimeout(30000);
+  jest.setTimeout(80000);
```
cf19drofxots: wtf...does it need to be that high?

packages/node/src/events/protocol-message/controller.ts
cf19drofxots: Why is this being sent out to begin with?

packages/machine/src/protocol/uninstall-virtual-app.ts
```diff
@@ -5,87 +5,273 @@ import {
...
+        toAddress: intermediaryAddress,
+        signature: context.signatures[0]
+      };
+    },
+
     // send M1, wait for M4
     Opcode.IO_SEND_AND_WAIT,
 
-    uninstallLeftAgreement,
+    // TODO: Add signature verification
```
cf19drofxots: Adding signature verification is not a part of this PR?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I'm supposed to add it in a separate PR

packages/node/src/events/protocol-message/controller.ts
snario: This is the `UNINSTALL_VIRTUAL` event that the UI needs to listen to in order to confirm the app has been uninstalled.

snario: Also, just sent out once.

packages/node/src/store.ts
snario: Because these are the sorted xpubs not the sorted addresses

packages/node/test/integration/uninstall-virtual.spec.ts
snario: It does not have to be this high, will revert

PR #738, opened 2019-02-12
[machine, node] make virtual app sequence numbers unique

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/machine/src/models/state-channel.ts
```diff
@@ -304,7 +304,8 @@ export class StateChannel {
       this.userNeuteredExtendedKeys,
       appInstances,
       this.ethVirtualAppAgreementInstances,
-      this.freeBalanceAppIndexes
+      this.freeBalanceAppIndexes,
+      this.monotonicNumInstalledApps + 1
```
snario: Don't think we need this + 1

packages/machine/src/protocol/install-virtual-app.ts
```diff
@@ -133,13 +137,31 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
...
+
+  const sc =
+    context.stateChannelsMap.get(key) ||
+    StateChannel.createEmptyChannel(key, [initiatingXpub, respondingXpub]);
+
+  const appSeqNo = sc.numInstalledApps;
+
+  const signingKeys = [xkeyKthAddress(intermediaryXpub, appSeqNo)].concat(
+    xkeysToSortedKthAddresses([initiatingXpub, respondingXpub], appSeqNo)
+  );
```
snario: Why isn't the whole list sorted?

packages/machine/src/protocol/install-virtual-app.ts
```diff
@@ -150,40 +172,25 @@ function createTarget(
...
-  const key = virtualChannelKey(
-    [initiatingXpub, respondingXpub],
-    intermediaryXpub
-  );
 
-  const sc = (
-    context.stateChannelsMap.get(key) ||
-    StateChannel.createEmptyChannel(key, [initiatingXpub, respondingXpub])
-  ).addVirtualAppInstance(targetAppInstance);
+  const newSc = sc.addVirtualAppInstance(target);
```
snario: Can you add a check inside `addVirtualAppInstance` to verify the correct `appSeqNo` was set?

packages/machine/src/protocol/install-virtual-app.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/mixins/MixinVirtualAppSetState.sol#L99

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/mixins/MixinVirtualAppSetState.sol#L19

packages/machine/src/models/state-channel.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why not? we need the number to increment for the next time someone calls this function

snario: Ah, my bad, yeah

packages/machine/src/protocol/install-virtual-app.ts
snario: Got it, can you reference the specs in a comment in the code

PR #751, opened 2019-02-13
[playground] [playground-server] Enables asynchronous registration flow (no need to wait for the multisig)

joelalejandro:
This PR introduces the following changes:

- The server no longer `await`s for the multisig to be created. Instead, it adds a `.then()` callback to store it to the database once it's available.
- Deposit is now queued if no multisig address is available. If it happens to be there, albeit unlikely, it works as before, executing the transaction immediately.
- The Playground polls every 5 seconds to the `/api/users` endpoint to figure out if the `multisig` is ready _(defining "ready" as "there is an address stored in the multisig_address column")_. While it's polling and no new data arrives, a spinner with the following message is shown:

>```
>We're configuring your state channel with the Playground.
>This can take 15-90 seconds, depending on network speed.
>```

- As soon as the multisig is detected, a modal is shown, with a "Proceed" button, and the following text:

>```
>To complete your registration, we'll ask you to confirm
>the deposit in the next step.
>```

- Confirming the deposit will trigger a transaction, monitored by the UI with a spinner with the following message:

>```
>We're waiting for the network to confirm your latest deposit.
>```

This PR also includes a complete refactor of the account creation / login / deposit logic, that was previously coupled to the UI `<header-account>` component. All of the related code has been moved to injectable functions as part of the `AccountState` tunnel.

**Preview**
deepin-screen-recorder_google-chrome_20190213182824

### DIFF
--- a/packages/machine/package.json
+++ b/packages/machine/package.json
@@ -20,7 +20,7 @@
     "@counterfactual/typescript-typings": "0.0.1",
     "@types/chai": "^4.1.7",
     "@types/dotenv-safe": "^5.0.3",
-    "@types/jest": "^24.0.1",
+    "@types/jest": "^24.0.4",
     "@types/node": "^11.9.0",
     "chai": "^4.2.0",
     "dotenv-safe": "^6.1.0",
--- a/packages/machine/src/models/state-channel.ts
+++ b/packages/machine/src/models/state-channel.ts
@@ -315,6 +315,13 @@ export class StateChannel {
   }
 
   public addVirtualAppInstance(appInstance: AppInstance) {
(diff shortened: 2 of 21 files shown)


joelalejandro:
@snario @cf19drofxots Something I've noticed while testing this PR is that `provider.estimateGas(tx)` seems to be doing nothing, and therefore the Metamask transaction gas cost defaults to zero. Maybe I'm misusing it somehow? https://github.com/counterfactual/monorepo/pull/751/files#diff-622ca62f685a0cd63d9e8fcc0447a630R41

cf19drofxots:
No you're not misusing it - it's implemented incorrectly

PR #759, opened 2019-02-14
[node] [playground] [playground-server] Update channel creation API 

cf19drofxots:
### Description

This updates the spec and implementation of the channel creation API to allow async channel creation.

It also includes a refactor for the Playground Server / Client to support the new API, by directly waiting on events to trigger instead of polling the API for changes.


packages/playground-server/src/db.ts
```diff
@@ -328,6 +329,31 @@ export async function bindMultisigToUser(
...
+  const db = getDatabase();
+
+  const query = db("users")
+    .where({ id: user.id })
+    .update("transaction_hash", transactionHash);
+
+  try {
+    await query;
+
+    Log.debug("Executed createUser query", {
```
cf19drofxots: Minor correction: Not creating a user here, just updating `transaction_hash` property

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -18,62 +19,80 @@ const CREATE_PROXY_AND_SETUP_GAS = 6e6;
...
 export default async function createChannelController(
   requestHandler: RequestHandler,
   params: Node.CreateChannelParams
-): Promise<Node.CreateChannelResult> {
-  const multisigAddress = await deployMinimumViableMultisigAndGetAddress(
+): Promise<Node.CreateChannelTransactionResult> {
+  const {
+    proxyFactory,
+    transactionHash
+  } = await deployMinimumViableMultisigAndGetTransactionHash(
```
snario: A slightly nicer API would be `const tx: Transaction = await sendMultisigCreationTxToChain(...)` and then `txHash = tx.hash`.

packages/playground-server/src/resources/user/processor.ts
```diff
@@ -69,12 +70,16 @@ export default class UserProcessor extends OperationProcessor {
...
-      async (result: Node.CreateChannelResult) => {
+    const { transactionHash } = await NodeWrapper.createStateChannelFor(
+      nodeAddress as string,
+      async (result: NodeTypes.CreateChannelResult) => {
         await bindMultisigToUser(newUser, result.multisigAddress);
       }
     );
 
+    newUser.attributes.transactionHash = transactionHash;
+    await bindTransactionHashToUser(newUser, transactionHash);
```
cf19drofxots: Is it a bad idea to set the `transactionHash` attribute on the front-end User instance _inside_ the `bindTransactionHashToUser` function even though this function is only concerned with setting the property on the backend DB?

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -234,28 +233,11 @@ export class AppRoot {
...
-      const user = await PlaygroundAPIClient.getUser(token as string);
-
-      if (!user.multisigAddress) {
-        this.waitForMultisig();
-        return;
-      }
+    const { provider, user } = this.accountState;
 
-      await this.setMultisig(user.multisigAddress);
+    provider.once(user.transactionHash, async () => {
```
cf19drofxots: Will this not race between doing a deposit and setting the multisig address on the user?

How do we ensure the multisig address is in fact set _before_ the deposit call is made?

packages/playground-server/src/db.ts
joelalejandro: Oh, snap :P I'll fix that.

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -18,62 +19,80 @@ const CREATE_PROXY_AND_SETUP_GAS = 6e6;
...
-  );
+      owners: params.owners
+    } as Node.CreateChannelResult;
 
-  await requestHandler.store.saveStateChannel(
-    stateChannelsMap.get(multisigAddress)!
-  );
+    const [respondingXpub] = params.owners.filter(
+      owner => owner !== requestHandler.publicIdentifier
+    );
```
snario: Might be nicer if the params are `initiatingXpub` and `respondingXpub` or, more generally speaking, clearly specifying who "I am" and who "I want to open a channel with is" versus passing in an array of the owners. Here's Raiden's API: https://raiden-network.readthedocs.io/en/v0.9.0/api_walkthrough.html#opening-a-channel. They use `partner_address` and get "my address" from the rest of the message.

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -18,62 +19,80 @@ const CREATE_PROXY_AND_SETUP_GAS = 6e6;
...
+async function deployMinimumViableMultisigAndGetTransactionHash(
   ownersPublicIdentifiers: string[],
   signer: Signer,
   multisigMasterCopyAddress: Address,
   proxyFactoryAddress: Address
-): Promise<Address> {
+): Promise<{
+  proxyFactory: Contract;
+  transactionHash: string;
+}> {
```
snario: This should be defined as a type in the file

packages/playground-server/src/resources/user/processor.ts
joelalejandro: It's a matter of separation of concerns. `bindTransactionHashToUser` (as well as `bindMultisigToUser`) should be better named. Something like `saveTransactionHash`, to make it clear that the function doesn't deal with how the API responds.

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -86,23 +105,28 @@ async function deployMinimumViableMultisigAndGetAddress(
...
-        { gasLimit: CREATE_PROXY_AND_SETUP_GAS }
+  try {
+    // TODO: implement retry around this with exponential backoff on the
+    // gas limit to increase probability of proxy getting created
+    const transaction: TransactionResponse = await proxyFactory.functions.createProxy(
+      multisigMasterCopyAddress,
+      new Interface(MinimumViableMultisig.abi).functions.setup.encode([
+        multisigOwnerAddresses
+      ]),
+      { gasLimit: CREATE_PROXY_AND_SETUP_GAS }
```
snario: I'm not sure where this variable is defined but we should probably create a "gas estimates" file in the codebase.

packages/node/test/integration/cant-take-action.spec.ts
```diff
@@ -112,43 +112,46 @@ describe("Node method follows spec - fails with improper action taken", () => {
         }
       };
 
-      const multisigAddress = await getNewMultisig(nodeA, [
+      nodeA.on(
+        NODE_EVENTS.CREATE_CHANNEL,
+        async (data: NodeTypes.CreateChannelResult) => {
+          const tttAppInstanceProposalReq = makeTTTAppInstanceProposalReq(
+            nodeB.publicIdentifier,
+            // @ts-ignore
```
snario: I don't see why we need to ignore this versus type it correctly

packages/node/test/integration/cross-node-multisig-creation.spec.ts
```diff
@@ -1,11 +1,22 @@
...
 
-import { IMessagingService, IStoreService, Node, NodeConfig } from "../../src";
+import {
+  IMessagingService,
+  IStoreService,
+  Node,
+  NODE_EVENTS,
+  NodeConfig
+} from "../../src";
 import { MNEMONIC_PATH } from "../../src/signer";
```
snario: We're doing two kinds of path resolution here. `"../../src/signer"` and `"../../src"`; need to decide on one or the other I think

packages/node/test/integration/cross-node-multisig-creation.spec.ts
```diff
@@ -65,14 +76,36 @@ describe("Node can create multisig, other owners get notified", () => {
...
+async function confirmChannelCreation(
+  nodeA: Node,
+  nodeB: Node,
+  ownersPublicIdentifiers: string[],
+  data: NodeTypes.CreateChannelResult
+) {
+  const openChannelsNodeA = await getChannelAddresses(nodeA);
+  const openChannelsNodeB = await getChannelAddresses(nodeB);
+  expect(openChannelsNodeA[0]).toEqual(data.multisigAddress);
+  expect(openChannelsNodeB[0]).toEqual(data.multisigAddress);
```
snario: Replace `toEqual` with `toBe` for string comparisons. `toEqual` is intended for object equality

packages/node/test/integration/get-app-instance.spec.ts
```diff
@@ -72,62 +78,65 @@ describe("Node method follows spec - getAppInstanceDetails", () => {
...
+
+        const installAppInstanceRequestId = generateUUID();
+        let installedAppInstance: AppInstanceInfo;
+
+        nodeA.on(NodeTypes.MethodName.INSTALL, async res => {
+          const installResult: NodeTypes.InstallResult = res.result;
+
+          installedAppInstance = installResult.appInstance;
+
+          // now we check to validate for correct AppInstance retrieval
```
snario: Unnecessary comment--- should be obvious from code

packages/playground/src/components/app-root/app-root.tsx
joelalejandro: This doesn't race because we don't actually get the multisig address until the user clicks the "Proceed" button on the modal.

packages/node/test/integration/reject-install.spec.ts
```diff
@@ -86,53 +86,67 @@ describe("Node method follows spec - rejectInstall", () => {
...
+                // Node A should have a proposal in place before Node B rejects it
+                expect((await getProposedAppInstances(nodeA)).length).toEqual(
+                  1
+                );
+
+                await nodeB.call(rejectReq.type, rejectReq);
+
+                expect((await getProposedAppInstances(nodeB)).length).toEqual(
+                  0
+                );
```
snario: The code above here can be cleaner... e.g.,
```js
const proposalsA = await getProposedAppInstances(nodeA);
const proposalsB = await getProposedAppInstances(nodeB);
expect(proposalsA.length).toBe(0);
expect(proposalsB.length).toBe(0);
```

I'm just particularly concerned with the code cleanliness in these files because they are pretty critical tests and if they aren't approachable to outsiders / other team members, no one will bother reading them. Again, not urgent to fix in this PR particularly but I want to signal my perspective on this.

packages/playground-server/src/db.ts
```diff
@@ -7,16 +7,16 @@ import { v4 as generateUuid } from "uuid";
 import Errors from "./errors";
 import User, { MatchedUser } from "./resources/user/resource";
 
-const DATABASE_URL = (process.env.DATABASE_URL as string) || (process.env.DB_CONNECTION_STRING as string);
+const DATABASE_URL =
+  (process.env.DATABASE_URL as string) ||
+  (process.env.DB_CONNECTION_STRING as string);
```
snario: Are these env variables documented anywhere? e.g., in the README
Also kind of confused why there are two for the same thing?

packages/playground-server/src/db.ts
```diff
@@ -276,7 +276,8 @@ export async function createUser(user: User): Promise<User> {
     email: user.attributes.email,
     eth_address: user.attributes.ethAddress,
     multisig_address: user.attributes.multisigAddress,
-    node_address: user.attributes.nodeAddress
+    node_address: user.attributes.nodeAddress,
+    transaction_hash: ""
```
snario: Why is `transaction_hash` on the user model? Also its a pretty generic token. Maybe specify as `multisig-deployment_transaction_hash`? In either case, seems odd to be on the user model.

packages/playground-server/src/node.ts
```diff
@@ -124,6 +102,7 @@ export default class NodeWrapper {
       {
         STORE_KEY_PREFIX: "store"
       },
+      // @ts-ignore
```
snario: Any better alternative than ignoring?

packages/playground/src/components/app-root/app-root.tsx
cf19drofxots: But the following scenario is possible, right?
1. transaction gets mined so this callback is called
2. modal pops up and user clicks to proceed
3. since the transaction is mined, it's _assumed_ that the multisig has in fact been created and it's _assumed_ that the address is set on the User property, but due to some network hiccup / latency issues, it's actually not set
4. Doing a deposit then obviously fails because `multisigAddress` is undefined

This might be an edge case but it's not extremely unlikely - it could definitely be worthwhile to implement some kind of resiliency around this

packages/node/test/integration/cant-take-action.spec.ts
cf19drofxots: This can be fixed by defining a custom namespace named `NodeJs` with an interface named `Global` that has the properties we're setting/getting (i.e. `networkContext`, etc) since these properties don't exist on the pre-existing `NodeJS.Global` type.

packages/node/src/methods/state-channel/create/controller.ts
cf19drofxots: We don't only care about the transaction itself, but also the contract that deploys it so we can listen for the creation and return the multisig address.

packages/playground-server/src/db.ts
joelalejandro: @snario I think it's for @ebryn's experiment of deploying to Heroku. We will probably later unify and have a single variable for this. There's a mention in the `README`, but definitely needs improvement.

packages/playground-server/src/db.ts
joelalejandro: @snario It probably doesn't need to be here, as it is quickly discarded. But I'm thinking - what if the transaction gets interrupted for some reason? We could use it to resume and finish the account creation flow.

packages/playground-server/src/node.ts
joelalejandro: Yeah, removing that, lol. It's not even necessary anymore.

packages/node/src/methods/state-channel/create/controller.ts
cf19drofxots: This would be nice (though it would limit the public facing API to two parties which would be awkward in terms of API versioning - can discuss more later).

packages/node/src/methods/state-channel/create/controller.ts
cf19drofxots: Will address as a follow up since the scope of that is very clear.

packages/node/test/integration/cant-take-action.spec.ts
joelalejandro: @cf19drofxots You can work around that using string literals like `global["networkContext"]`. That doesn't require ts-ignoring.

cf19drofxots: Yeah doing that right now.

cf19drofxots: I wanted to specify an interface so it would error if the string literal was typo'ed but it's giving me weird behavior that I'm not gonna spend more time trying to resolve.

packages/node/test/integration/reject-install.spec.ts
cf19drofxots: Agreed!

packages/playground-server/src/resources/user/processor.ts
cf19drofxots: Not urgent, just a passing thought.

packages/playground/src/components/app-root/app-root.tsx
joelalejandro: The scenario *is* possible, since we're listening events for the same action in two different places. If it's OK with you guys, I'd like to think of resiliency as an iteration (since it might be an edge case) so we can get this to `master` in a working state.

cf19drofxots: That's fine as long as we don't forget it / put it off for too much later, as reactively dealing with state-divergence issues is usually really annoying in distributed systems and stronger guarantees (i.e. safer assumptions) are heavily preferred if their cost is relatively negligible.

joelalejandro: As discussed, resiliency concerns will be addressed _before_ merging this.

joelalejandro:
@snario Do not merge yet, there's a race condition currently happening as described in this thread: https://github.com/counterfactual/monorepo/pull/759#discussion_r258144917

snario:
Can we rebase this on `master` @joelalejandro

joelalejandro:
@snario @cf19drofxots This PR still needs work on the Node layer. The CREATE_CHANNEL event is never being triggered on the Playground Server, therefore the multisig is never stored in the database.

PR #762, opened 2019-02-14
[playground] Implement metamask account checking UI

snario:
<need to add details>

# No Web3
screen shot 2019-02-14 at 16 25 19

### DIFF
--- a/packages/playground/src/components/webthree-connector/webthree-connector.tsx
+++ b/packages/playground/src/components/webthree-connector/webthree-connector.tsx
@@ -14,23 +14,23 @@ export class Web3Connector {
   @Prop() networkState: NetworkState = {};
 
   getCurrentAddress() {
-    return (window as any).web3.eth.accounts[0];
+    return window["web3"].eth.accounts[0];
   }
 
   getCurrentNetwork() {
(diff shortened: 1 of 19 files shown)


packages/playground/src/components/webthree-connector/webthree-connector.tsx
```diff
@@ -11,23 +13,77 @@ export class Web3Connector {
   @Prop() accountState: AccountState = {} as AccountState;
   @Prop() networkState: NetworkState = {};
 
-  async componentDidLoad() {
+  getCurrentAddress() {
+    return (window as any).web3.eth.accounts[0];
```
joelalejandro: @snario You can avoid the `as any` cast using index references

```
return window["web3"].eth.accounts[0];
```

Also applies to functions below.

packages/playground/package.json
```diff
@@ -29,7 +29,7 @@
   "homepage": "https://counterfactual.com",
   "devDependencies": {
     "@counterfactual/typescript-typings": "^0.0.1",
-    "@types/jest": "24.0.4",
+    "@types/jest": "23.0.4",
```
joelalejandro: @snario This change shouldn't be necessary?

PR #765, opened 2019-02-15
[playground] Implements account withdrawal

joelalejandro:


packages/playground/src/components/account/account-deposit/account-deposit.tsx
```diff
@@ -32,8 +32,8 @@ export class AccountDeposit {
     );
   }
 
-  async formSubmissionHandler(e) {
-    this.amountDeposited = ethers.utils.parseEther(e.target.value);
+  async formSubmitionHandler(e) {
```
cf19drofxots: `submission` is actually the correct spelling

joelalejandro: lol, someone corrected this in master :P

joelalejandro: I'll fix it in another PR so I don't trigger another CI job here.

PR #767, opened 2019-02-15
Multiple improvements to UI

snario:
- Don't popup metamask automatically at start (connect to metamask button)
- Various css changes


packages/playground/src/components/layout/layout-header/header-account/header-account.tsx
```diff
@@ -60,6 +61,19 @@ export class HeaderAccount {
     }
   }
 
+  onConnectMetamask() {
+    this.metamaskConfirmationUIOpen = true;
+
+    window["ethereum"]
+      .enable()
+      .then(() => {
```
joelalejandro: @snario Is it possible to `async/await` this? 

```ts
try {
  await window["ethereum"].enable();
} catch {} finally {
  this.metamaskConfirmationUIOpen = false;
}
```

snario: I wanted to make it asynchronous specifically so the line just above this which sets a local component variable executed and triggered the render on the component. Is there a "force render" method I could call?

snario: Nevermind, fixed.

PR #768, opened 2019-02-15
[playground] Adds boolean for first check of network state

snario:


packages/playground/src/components/app-home/app-home/app-home.tsx
```diff
@@ -28,28 +30,56 @@ export class AppHome {
   }
 
   render() {
+    let content;
+
+    if (!this.hasDetectedNetwork) {
+      content = (
+        <div class="loading">
```
joelalejandro: @snario Why not use `<widget-spinner />`? We already have a component for this.

snario: I wanted the specific effect of pulsating circles instead of a spinning circle. Maybe I should make that a flavour / mode of that widget?

joelalejandro: Oh, cool. :+1: We can add that as a setting later to the component. Didn't notice it was a different shape.

PR #770, opened 2019-02-16
[node] Change messaging inbox structure to support multiple peers

snario:


snario:
Waiting for message queueing PR.

PR #772, opened 2019-02-16
[node] Add gasPrice and gasLimit to tx on deposit and withdraw

snario:
Note, some of this is already included inside #765


packages/node/src/methods/state-channel/withdraw/controller.ts
```diff
@@ -25,14 +26,26 @@ export default async function withdrawController(
...
+    to,
+    value,
+    data
+  } = await requestHandler.store.getWithdrawalCommitment(multisigAddress);
+
+  const tx = {
+    to,
+    value,
+    data,
+    gasLimit: 300000, // TODO: Estimate correct value,
```
cf19drofxots: Remove TODO as estimate works

PR #773, opened 2019-02-16
[contracts] Upgrade solidity and solc to 0.5.3

snario:
Part 1 of a 3 part series to use CREATE2.


snario:
I'm downgrading this to 0.5.3 because of https://github.com/ethereum/solidity/issues/5997.

PR #774, opened 2019-02-17
Fix properly setting the multisig on the user attributes

cf19drofxots:


packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -242,11 +243,25 @@ export class AppRoot {
...
       }
 
-      this.updateAccount({ ...this.accountState, user });
+      await this.setMultisig(user.multisigAddress);
       await this.requestToDepositInitialFunding();
     }, 5000);
   }
 
+  async setMultisig(multisigAddress: string) {
+    setTimeout(async () => {
```
snario: Why timeout

PR #781, opened 2019-02-18
[node] Fix installing multiple virtual app instances

cf19drofxots:
### Description

This fixes the issue with being able to install multiple virtual AppInstances, with each having unique IDs.

It also adds a test for this case.

### DIFF
--- a/packages/machine/src/index.ts
+++ b/packages/machine/src/index.ts
@@ -23,6 +23,7 @@ import {
   xkeysToSortedKthSigningKeys
 } from "./xkeys";
 
+export * from "./virtual-app-key";
 export {
   appIdentityToHash,
   AppInstance,
--- a/packages/machine/src/index.ts
+++ b/packages/machine/src/index.ts
@@ -23,7 +23,7 @@ import {
   xkeysToSortedKthSigningKeys
 } from "./xkeys";
 
-export * from "./virtual-app-key";
+export { virtualChannelKey } from "./virtual-app-key";
 export {
   appIdentityToHash,
   AppInstance,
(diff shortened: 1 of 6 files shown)


packages/machine/src/index.ts
```diff
@@ -23,6 +23,7 @@ import {
   xkeysToSortedKthSigningKeys
 } from "./xkeys";
 
+export * from "./virtual-app-key";
```
snario: Can you use explicit naming

packages/node/src/events/propose-install-virtual/operation.ts
```diff
@@ -1,8 +1,8 @@
 import { Node } from "@counterfactual/types";
 
+import { getOrCreateVirtualChannel } from "../../methods/app-instance/propose-install-virtual/operation";
```
snario: Perhaps lift to util file

packages/node/test/integration/install-virtual.spec.ts
```diff
@@ -23,7 +23,7 @@ import {
 } from "./utils";
 
 describe("Node method follows spec - proposeInstallVirtual", () => {
-  jest.setTimeout(30000);
+  jest.setTimeout(60000);
```
snario: lower?

packages/node/src/events/propose-install-virtual/operation.ts
cf19drofxots: If other contexts end up needing this we can do that, but this is relatively specific to the context of virtual channels so might be better to leave it here (as opposed to a more general util location)

packages/node/test/integration/install-virtual.spec.ts
cf19drofxots: I can experiment with lower timeouts, just doubled it because the test is twice as long now.

PR #792, opened 2019-02-19
Bump @counterfactual/node dependency

snario:


joelalejandro:
Let's hold off on this for a bit. I've returned the Playground Server to yarn workspaces so we can think this through more in depth.

PR #800, opened 2019-02-20
[machine] [discussion] runtime validation of message.params

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
I don't think any other code is doing runtime validation of this so an invalid message will be caught as late as possible (e.g. when trying to do `new HDNode(undefined)`; this PR proposes one early place to do runtime validation; what do people think?

### DIFF
--- a/packages/machine/src/protocol/withdraw-eth.ts
+++ b/packages/machine/src/protocol/withdraw-eth.ts
@@ -13,7 +13,12 @@ import {
   WithdrawETHCommitment
 } from "../ethereum";
 import { AppInstance, StateChannel } from "../models";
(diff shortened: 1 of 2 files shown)


snario:
Seems like this could be automated for each type versus hand-crafting this code

snario:
Updates here @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll ?

cf19drofxots:
Is this PR still up for consideration or can we close it? @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll @snario

cf19drofxots:
Assuming this is stale so closing. Re-open if necessary.

PR #804, opened 2019-02-20
[node] Introduce queueing for method calls

cf19drofxots:
### Description

This introduces queueing internal to the Node to provide safer execution guarantees for calls being made against the Node / updating the state of a Node.

### Related issues
- https://github.com/counterfactual/monorepo/pull/765


packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -4,68 +4,83 @@ import { xkeysToSortedKthAddresses } from "@counterfactual/machine";
...
-  const multisigAddress = await deployMinimumViableMultisigAndGetAddress(
-    params.owners,
-    requestHandler.wallet,
-    requestHandler.networkContext.MinimumViableMultisig,
-    requestHandler.networkContext.ProxyFactory
-  );
+class ChannelCreator extends NodeController {
+  static async enqueueByShard(requestHandler: RequestHandler): Promise<Queue> {
+    return await requestHandler.getShardedQueue(
+      QUEUE_SHARD_KEYS.CHANNEL_CREATION
```
snario: Maybe just use a string `"ChannelCreator"`

packages/node/src/types.ts
```diff
@@ -1,5 +1,41 @@
 import { ProtocolMessage } from "@counterfactual/machine";
 import { Address, Node } from "@counterfactual/types";
+import Queue from "promise-queue";
+
+import { RequestHandler } from "./request-handler";
+
+export class NodeController {
```
snario: Can we just put this in its own file?

packages/node/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "declaration": true,
-    "declarationDir": "dist/"
+    "declarationDir": "dist/",
+    "experimentalDecorators": true
```
snario: Isn't this already true in the root level config

PR #822, opened 2019-02-24
[playground][playground-server] Spin up local Firebase server

cf19drofxots:
### Description

This spins up a local Firebase server to make iterating/testing in isolation easier.


packages/playground/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.0.1",
   "scripts": {
     "build": "stencil build --debug && cp src/_redirects www/_redirects",
-    "start": "stencil build --dev --watch --serve",
+    "serve": "stencil build --dev --watch --serve",
```
cf19drofxots: @ebryn we'll have to update the CD script to use `serve` instead of `start` for the playground

snario: We can't keep it as `start`?

PR #841, opened 2019-02-26
Use same version of jest everywhere possible

snario:


snario:
Added #845 to cover discrepancy that "where possible" implies

PR #842, opened 2019-02-26
Apply 'types' field in tsconfigs

snario:


packages/cf.js/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "sourceRoot": ".",
     "declarationDir": "dist/",
-    "outDir": "dist/"
+    "outDir": "dist/",
+    "types": ["jest"]
```
cf19drofxots: Why is this needed?

snario: So we don't need to import `expect`, `beforeAll`, etc and also so we don't need to rely on a global in typescript-typings

cf19drofxots: We import those? Where?

snario: It is kind of subtle at the moment. `@types/jest` happens to get hoisted and in the root tsconfig we have a typeRoots config option that fetches it. However if we explicitly declare this on a per-package level it is safer and does not assume hoisting. Super tiny detail

PR #855, opened 2019-02-26
[dapp-high-roller] [dapp-tic-tac-toe] Implements bet range of 0.0 to 0.01 ETH across all games

joelalejandro:
This PR implements range validations for the "Bet" fields on both High Roller and Tic-Tac-Toe.

It also solves a routing error with a workaround we had to a previously unsolved issue with the Stencil Router.

**Screenshots**
image

image

image

image


packages/dapp-high-roller/src/components/app-root/app-root.tsx
```diff
@@ -214,7 +217,7 @@ export class AppRoot {
       pathname: "/game",
       state: {
         isProposing,
-        betAmount: "0.1"
+        betAmount: ethers.utils.formatEther(this.state.appInstance.myDeposit)
```
cf19drofxots: Do we do any error handling before we `formatEther`?

What happens if someone enters "some amount"?

joelalejandro: Since the input type is `number`, this happens:

image

joelalejandro: Also, note that this comes from the appInstance - it can't get there without blowing up somewhere before in the stack :P

packages/dapp-high-roller/src/components/app-wager/app-wager.tsx
```diff
@@ -249,11 +260,12 @@ export class AppWager {
...
               type="number"
-              placeholder="3 ETH"
+              placeholder="0.01"
               value={this.betAmount}
               onInput={e => this.handleChange(e, "betAmount")}
-              readonly={true}
-              step="0.01"
+              min={0}
+              max={0.01}
+              step={0.001}
```
cf19drofxots: So the value can't be something like `0.00001` if step is `0.001`?

joelalejandro: It can, but you have to type it manually.

PR #858, opened 2019-02-26
[machine] Fix bigNumber issue

cf19drofxots:
### Description

This resolves an overflow issue when installing virtual AppInstances.


packages/machine/src/models/eth-virtual-app-agreement-instance.ts
```diff
@@ -24,7 +24,7 @@ export class ETHVirtualAppAgreementInstance {
     public expiry: number,
     // todo(xuanji): The following field is a js `number`, which is
```
joelalejandro: This comment can be removed now, I guess :D

PR #862, opened 2019-02-27
[playground] several improvements to state management and UI

snario:


packages/node/src/methods/state-channel/withdraw/controller.ts
```diff
@@ -1,8 +1,10 @@
+import { xkeyKthAddress } from "@counterfactual/machine";
```
ebryn: `Kth`?

snario: The `k`-th child of the Node's [extended key](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys)

packages/playground/src/components/account/account-exchange/account-exchange.tsx
```diff
@@ -47,18 +52,40 @@ export class AccountExchange {
 
   async onDepositClicked(e) {
     const amount = e.target.value;
-    await this.deposit(amount, this.user.multisigAddress);
+    try {
+      await this.deposit(ethers.utils.parseEther(amount));
+    } catch (e) {
+      if (e.toString().includes("Cannot deposit while another deposit")) {
+        window.alert(HUB_IS_DEPOSITING_ALERT);
```
joelalejandro: @snario We have a `<widget-dialog>` component to take care of this sort of messages. We should render that instead of using `window.alert()`.

You can define a `@State() dialog: JSX.Element` member in the component class and have it set to something like:

```jsx
const content = <label>{HUB_IS_DEPOSITING_ALERT}</label>;

this.dialog = <widget-dialog
  visible={true}
  dialogTitle="Sorry, can't deposit right now :("
  content={content}
  primaryButtonText="OK"
  onPrimaryButtonClicked={() => this.dialog = undefined}
/>;
```

Then, you should add `{this.dialog}` to the `render()` JSX.

packages/playground/src/components/account/account-exchange/account-exchange.tsx
```diff
@@ -47,18 +52,40 @@ export class AccountExchange {
...
   }
 
   async onWithdrawClicked(e) {
     const amount = e.target.value;
-    await this.withdraw(amount);
+    try {
+      await this.withdraw(ethers.utils.parseEther(amount));
+    } catch (e) {
+      if (e.toString().includes("Cannot withdraw while another deposit")) {
+        window.alert(HUB_IS_DEPOSITING_ALERT);
```
joelalejandro: @snario Same comment as before, we should use `<widget-dialog>`.

packages/playground/src/components/account/account-register/account-register.tsx
```diff
@@ -78,16 +79,21 @@ export class AccountRegister {
...
-    );
+    const payload = buildRegistrationSignaturePayload(data);
+
+    this.metamaskConfirmationUIOpen = true;
+
+    try {
+      const signature = await this.signer.signMessage(payload);
+      await this.register(signature);
+    } catch (e) {
+      await this.register("", e);
```
joelalejandro: @snario What's the purpose of this? Seems a bit counter-intuitive.

packages/playground/src/components/account/account-register/account-register.tsx
```diff
@@ -198,14 +203,19 @@ export class AccountRegister {
             error={this.errors.email}
             onChange={e => this.change("email", e)}
           />
-          <form-input
+          {/* <form-input
```
joelalejandro: @snario Why are we hiding the current address? Seems to be useful information. We might render it like something else instead of a text input, i.e. a label.

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -44,7 +44,7 @@ export class AppRoot {
     this.setup();
   }
 
-  async updateAccount(newProps: AccountState) {
+  async updateAccount(newProps: Partial<AccountState>) {
```
joelalejandro: @snario TIL `Partial<T>`, this is awesome!

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -182,7 +182,7 @@ export class AppRoot {
 
     if (user.ethAddress) {
       provider.removeAllListeners(user.ethAddress);
```
joelalejandro: @snario If you're commenting the `provider.on()` line, this one doesn't make too much sense on its own.

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -201,7 +201,7 @@ export class AppRoot {
     const { user } = this.accountState;
     const { signer } = this.walletState;
 
-    const signature = await signer.signMessage(
+    const signature = await signer!.signMessage(
```
joelalejandro: @snario Let's avoid using `!`. We can do above:

```ts
const signer = this.walletState.signer as Signer;
```

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -219,100 +219,121 @@ export class AppRoot {
...
 
-    return {
-      balance: multisigBalance,
-      accountBalance: walletBalance
+    const vals = {
+      ethFreeBalanceWei: ethers.utils.bigNumberify(balance),
+      ethMultisigBalance: await provider!.getBalance(multisigAddress)
     };
+
+    this.accountState.updateAccount!(vals);
```
joelalejandro: @snario While working on `app-root`, you shouldn't call the methods on the `accountState`, since the component owns the implementations of the `update*` functions. Call directly `this.updateAccount()` to get rid of the `!`.

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -219,100 +219,121 @@ export class AppRoot {
...
-      accountBalance: (accountBalance as number) - valueInWei,
-      unconfirmedBalance: (this.accountState.balance as number) + valueInWei
+    node.once(Node.EventName.DEPOSIT_STARTED, args => {
+      this.updateAccount({
+        ethPendingDepositTxHash: args.txHash,
+        ethPendingDepositAmountWei: valueInWei
+      });
     });
 
+    const resetPendingDepositState = () =>
```
joelalejandro: @snario This shouldn't be an inline function, since it'll be re-declared every time we deposit. Can we expose it as a class function, instead?

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -219,100 +219,121 @@ export class AppRoot {
...
-      accountBalance: (accountBalance as number) + valueInWei,
-      unconfirmedBalance: (this.accountState.balance as number) - valueInWei
+    node.once(Node.EventName.WITHDRAWAL_STARTED, args => {
+      this.updateAccount({
+        ethPendingWithdrawalTxHash: args.txHash,
+        ethPendingWithdrawalAmountWei: valueInWei
+      });
     });
 
+    const resetPendingWithdrawalState = () =>
```
joelalejandro: @snario Same as `resetPendingDepositState`, shouldn't be an inline function.

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -321,59 +342,52 @@ export class AppRoot {
...
 
-  async confirmDepositInitialFunding(
-    pendingAccountFunding,
-    multisigAddress: string
-  ) {
+  async confirmDepositInitialFunding() {
     this.modal = {};
 
-    await this.deposit(pendingAccountFunding, multisigAddress);
+    await this.deposit(this.accountState.precommitedDepositAmountWei!);
```
joelalejandro: @snario Let's not use `!` -- instead, let's cast to the appropriate type:

```ts
await this.deposit(this.accountState.precommitedDepositAmoutnWei as BigNumber);
```

packages/playground/src/components/layout/layout-header/header-account/header-account.scss
```diff
@@ -6,6 +6,10 @@
     .btn {
       font-size: 15px;
     }
+
+    .btn-error {
+      background: #d02020;
```
joelalejandro: @snario We have a `$c-red` variable, if I recall correctly, we should use that.

packages/playground/src/components/layout/layout-header/header-account/header-account.tsx
```diff
@@ -19,14 +19,17 @@ import { UserSession } from "../../../../types";
 })
 export class HeaderAccount {
   @Element() el!: HTMLStencilElement;
-  @Prop() balance: number = 0;
+  @Prop() ethFreeBalanceWei: BigNumber = ethers.constants.Zero;
   @Prop() network: string = "";
+  @Prop() error: { primary: string; secondary: string } = {
+    primary: "",
+    secondary: "'"
```
joelalejandro: @snario What's the purpose of the single quote here?

packages/types/src/node-protocol.ts
```diff
@@ -87,9 +90,11 @@ export namespace Node {
 
   export type WithdrawParams = {
     multisigAddress: string;
+    recipient?: string;
     amount: BigNumber;
   };
   export type WithdrawResult = {
+    recipient: string;
```
joelalejandro: @snario Why is it required to have a `recipient` here and not on the `WithdrawParams` type?

packages/playground/src/components/account/account-exchange/account-exchange.tsx
snario: This was actually meant to be a temporary stub and my plan was to listen for the deposit events from the playground-server and disable the entire /exchange UI and replace with an overlay that would say "Please wait for the hub to collateralize your channel." and some secondary text "Curious why you can't deposit? See <this issue> on the Counterfactual GitHub if you're curious to understand how asynchronous deposits will be added in the future"

packages/playground/src/components/account/account-register/account-register.tsx
snario: The way `register` was written is that is gets called with two arguments; an error or a signed message and it handles it differently depending. Pretty easy to refactor. I'll refactor.

packages/playground/src/components/account/account-register/account-register.tsx
snario: Yeah I think I'll render it differently, but it doesn't make sense to be a form element

packages/playground/src/components/app-root/app-root.tsx
snario: I think I uncommented that line

packages/playground/src/components/app-root/app-root.tsx
snario: Why is this better?

packages/playground/src/components/app-root/app-root.tsx
snario: ah ok

packages/types/src/node-protocol.ts
snario: In the params it can go to a default value but in the result it gives you the one it used (either the param or the default)

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -321,59 +344,53 @@ export class AppRoot {
     const { user } = this.accountState;
     const { provider } = this.walletState;
 
-    provider.once(user.transactionHash, async () => {
+    provider!.once(user.transactionHash, async () => {
```
joelalejandro: @snario ¬¬ this again?

snario: My bad

packages/playground/src/components/webthree-connector/webthree-connector.tsx
```diff
@@ -75,16 +85,13 @@ export class Web3Connector {
...
-      accountBalance: 0
+      }
     });
 
-    const provider = new Web3Provider(this.getProvider());
     this.walletState.updateWalletConnection!({
       ...(await this.getCurrentWalletState()),
-      provider,
-      signer: provider.getSigner(),
+      provider: this.getProvider(),
```
joelalejandro: @snario This doesn't seem necessary.

snario: It is necessary, this is the only place it gets set

PR #863, opened 2019-02-27
[machine,node] commitment store cleanup

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- rename `STATE_TRANSITION_COMMIT ` -> `WRITE_COMMITMENT`, since the state commitment change actually happens outside of any middlewares
- add `context.finalCommitment`, for `WRITE_COMMITMENT` to read
- use `context.finalCommitment` with withdraw and setup
- add actual signing to `instruction-executor.spec.ts` to avoid errors
- export more params via `machine` instead of `dist`

### DIFF
--- a/packages/node/src/node.ts
+++ b/packages/node/src/node.ts
@@ -4,16 +4,14 @@ import {
   Opcode,
   Protocol,
   ProtocolMessage,
-  SetupParams
+  SetupParams,
+  UpdateParams,
+  WithdrawParams,
+  Transaction
 } from "@counterfactual/machine";
 import {
-  UpdateParams,
-  WithdrawParams
-} from "@counterfactual/machine/dist/src/types";
-import {
   NetworkContext,
-  Node as NodeTypes,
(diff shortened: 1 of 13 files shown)


packages/node/src/node.ts
```diff
@@ -216,34 +215,41 @@ export class Node {
...
+            context.finalCommitment!
           );
         } else {
+          if (!context.commitments[0]) {
+            throw new Error(
+              `State transition without commitment: ${JSON.stringify(message)}`
+            );
+          }
+          const transaction = context.commitments[0].transaction([
+            context.signatures[0] // TODO: add counterparty signature
```
snario: Can we address this before merging @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: OK - this should be resolved by switching all the protocols that are in this case

PR #873, opened 2019-02-28
[playground] [dapp-high-roller] [dapp-tic-tac-toe] Implements dApp resumability

joelalejandro:
This PR implements resumability by sending the dApps' URL to the Playground so when the user refreshes, the dApp's iframe can be loaded with the last known URL. 

:construction: Pending to test if the game state resumes where it left (specially in High Roller). 

**Some notes:**

- `stencil-router` has no support for listening to route changes, so I'm forced to use `hash` routing on the High Roller dApp and listen to `popstate` events. Browser-mode routing isn't event-driven, and polling is not a preferred solution.

- The `app-root` component now delegates control to the Router after it's finished processing the user data sent via the `playground:response:user` postMessage() reply.

deepin-screen-recorder_vivaldi-stable_20190228135424

### DIFF
--- a/packages/dapp-high-roller/src/components/app-root/app-root.scss
+++ b/packages/dapp-high-roller/src/components/app-root/app-root.scss
@@ -1,0 +1,13 @@
+.App {
+  position: relative;
+  max-width: 20rem;
+  height: 100vh;
+  margin: 0 auto;
+}
+
+.message {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  text-align: center;
+}
--- a/packages/dapp-high-roller/src/components/app-root/app-root.tsx
+++ b/packages/dapp-high-roller/src/components/app-root/app-root.tsx
@@ -134,6 +134,7 @@ export class AppRoot {
         multisigAddress: "0x9499ac5A66c36447e535d252c049304D80961CED"
       };
       this.updateAccount(mockAccount);
+      this.userDataReceived = true;
     }
   }
 
--- a/packages/dapp-high-roller/src/components/app-root/app-root.tsx
+++ b/packages/dapp-high-roller/src/components/app-root/app-root.tsx
@@ -1,6 +1,6 @@
 declare var ga: any;
 
-import { Component, Prop } from "@stencil/core";
+import { Component, Prop, State } from "@stencil/core";
 import { RouterHistory } from "@stencil/router";
 
 import CounterfactualTunnel from "../../data/counterfactual";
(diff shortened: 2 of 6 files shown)


packages/dapp-high-roller/src/components/app-root/app-root.scss
```diff
@@ -0,0 +1,13 @@
+.App {
```
snario: I think this is the first time we use a capital letter in css class names? Super minor but maybe we should keep this consistent

packages/dapp-high-roller/src/components/app-root/app-root.tsx
```diff
@@ -284,6 +296,8 @@ export class AppRoot {
           </CounterfactualTunnel.Provider>
         </main>
       </div>
+    ) : (
+      <h1 class="App message">connecting....</h1>
```
snario: Might be an opportunity for a better message here. "Loading state from state channels wallet" or "Connecting to wallet..."? Also separately we could re-use the loading animation with the 3 pulsing dots

packages/playground/src/components/dapp-container/dapp-container.tsx
```diff
@@ -106,7 +108,7 @@ export class DappContainer {
   }
 
   private async handlePlaygroundMessage(event: MessageEvent): Promise<void> {
-    if (!this.frameWindow) {
+    if (!this.frameWindow || typeof event.data !== "string") {
```
snario: What does `typeof event.data !== "string"` represent? Maybe assign it to a variable with a descriptive name

packages/dapp-high-roller/src/components/app-root/app-root.scss
joelalejandro: Actually it's like that on TTT as well. Just copy pasted enough code to make the connecting message available. I'll fix it.

packages/dapp-high-roller/src/components/app-root/app-root.tsx
joelalejandro: This is the same message TTT has. We can probably iterate it.

packages/playground/src/components/dapp-container/dapp-container.tsx
joelalejandro: event.data is the postMessage() payload sent from the dApp to the playground. Since the `message` event is also shared with browser extensions and all sort of things that *could* be transferring objects, I need to type check it.

PR #877, opened 2019-02-28
[machine,node] write commitments for install, update, uninstall

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/node/src/store.ts
```diff
@@ -231,6 +231,7 @@ export class Store {
             ["string", "string", "string"],
             [
               proposedAppInstance.intermediaries![0],
+              // todo(xuanji): doesn't sorting destroy this?
```
snario: Pls solve before merging

packages/node/src/node.ts
```diff
@@ -241,25 +241,34 @@ export class Node {
...
-            protocol,
-            transaction
+          verifyFinalCommitment(context.finalCommitment);
+          await this.requestHandler.store.setSetupCommitmentForMultisig(
+            params.multisigAddress,
+            context.finalCommitment!
           );
+        } else {
+          /// install-virtual-app, uninstall-virtual-app
+          console.log("skipping commitment write - not implemented yet");
```
snario: Pls make sure there is a story to fix this in the next day or two

packages/node/src/node.ts
```diff
@@ -241,25 +241,34 @@ export class Node {
...
+              "WRITE_COMMITMENT: params did not contain multisigAddress"
             );
           }
-          const transaction = context.commitments[0].transaction([
-            context.signatures[0] // TODO: add counterparty signature
-          ]);
-          if (!context.appIdentityHash) {
-            throw new Error(
-              `appIdentityHash required to store commitment. protocol=${protocol}`
+          if (!getAddress(params.multisigAddress)) {
```
snario: Why is this here? Surely it would have been caught well ahead of time in the instruction executor process

packages/node/src/node.ts
```diff
@@ -241,25 +241,34 @@ export class Node {
...
+        } else if (
+          protocol === Protocol.Install ||
+          protocol === Protocol.Uninstall ||
+          protocol === Protocol.Update
+        ) {
+          /// install, update, uninstall,
+          /// All other protocols are stored under the key
+          /// appIdentityHash/protocolName. This means that an app only has one
+          /// update commitment.
+          const params = message.params as any;
```
snario: Don't cast to `any`

packages/machine/src/protocol/update.ts
```diff
@@ -85,14 +91,20 @@ export const UPDATE_PROTOCOL: ProtocolExecutionFlow = {
     // Sign the same state update yourself
     Opcode.OP_SIGN,
 
+    (message: ProtocolMessage, context: Context) => {
+      context.finalCommitment = context.commitments[0].transaction([
+        context.signatures[0],
+        message.signature!
+      ]);
```
snario: For code clarity could we name these values?

e.g.,

```ts
    function generateSignedCommitment(message: ProtocolMessage, context: Context) => {
        const setStateCommitment = context.commitments[0] as SetStateCommitment;
        const mySignature = context.signatures[0];
        const counterpartySignature = message.signature as Signature;
	      context.finalCommitment = setStateCommitment.transaction([
            mySignature, 
            counterpartySignature
        ]);
    }

packages/node/src/node.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Typescript doesn't let me access `params.multisigAddress` without the cast

packages/machine/src/protocol/update.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: good idea

packages/node/src/node.ts
snario: Cast it to what it is, not `any`. In this case it would be `SetupParams | InstallParams | UninstallParams`

PR #890, opened 2019-03-01
[machine] replace Context with generator-based communication

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Currently, the protocol definitions are organized as an array of async functions and opcodes. The `Context` object is used for communication between the async functions and the opcodes, as well as to persist values between the async functions.

This PR replaces the array with one async generator function. To call a middleware, the generator function yields, suspending execution and returning to the caller (i.e., `InstructionExecutor`). The yielded value is interpreted as an array `[opcode, ..args]` where `args` are passed by the `InstructionExecutor` to the middleware responsible for `opcode`.

```
yield [Opcode.OP_SIGN, installRefundCommitment]
```

The values returned by the middleware is passed into the generator function as the return value of `next`.

```
const s1 = yield [Opcode.OP_SIGN, installRefundCommitment]
```


ebryn:
I like this. I think we should perhaps wait to merge it until after our release, to prevent any further instability?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
SGTM

PR #891, opened 2019-03-01
[machine] fold some validations into the same function

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Ah

PR #894, opened 2019-03-01
[on hold] [wip] [playground-server] Implements available funds criteria into the matchmaking process

joelalejandro:
This PR modifies the way matchmaking works to take into consideration if a matched user has enough funds to join the game.

The API queries the Node to get the available funds in the State Channel. 

:construction: Pending validation when an Internet connection is stable.


ebryn:
@joelalejandro This was intentionally left out of scope. Let's hold off on this until after launch.

PR #927, opened 2019-03-03
Implementation of eth_call methods inside the machine (fixes most bugs on demo too)

snario:
# Hook up `TAKE_ACTION` method on `Node` to `machine`
The `TAKE_ACTION` method was previously just setting state in the DB and not actually running the protocol at all. This fixes that.

# Add state validation inside the `machine`

## Implement 
The `TAKE_ACTION` method on the `Node` did not use the machine at all previously. Now it does. 

## Add `JsonRpcProvider` to `machine`
Additionally, the `machine` now accepts `JsonRpcProvider` which is being used to make `eth_call`s to the on-chain contracts to compute state transitions. This is used in the `TakeActionProtocol` and `Uninstall[Virtual]Protocol`.

## Call `resolve` inside the `Uninstall[Virtual]Protocol`
We now call `resolve` inside these protocols so that the resolution can be verified by all parties involved. **It is very important to note** that the `resolve` function must return a `Transfer.Transaction` object where the `to` field is populated with the 0th child address of the extended key of the Node. For example, if Node A is playing Tic Tac Toe and wins, the resolution must be to the 0th child key address of Node A.


packages/machine/src/protocol/take-action.ts
```diff
@@ -0,0 +1,129 @@
...
+  addSignedCommitmentToOutboxForSeq1
+} from "./utils/signature-forwarder";
+import { validateSignature } from "./utils/signature-validator";
+
+type TakeActionProtocolMessage = ProtocolMessage & { params: TakeActionParams };
+
+/**
+ * @description This exchange is described at the following URL:
+ *
+ * TODO:
```
snario: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll

packages/dapp-high-roller/package.json
```diff
@@ -26,7 +26,7 @@
   "homepage": "https://counterfactual.com",
   "devDependencies": {
     "@stencil/sass": "^0.1.1",
-    "@types/jest": "24.0.9",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: ?

packages/dapp-high-roller/src/data/types.ts
```diff
@@ -2,15 +2,36 @@ import { BigNumber, BigNumberish } from "ethers/utils";
...
 export type Bytes32 = string;
 
-export type SolidityABIEncoderV2Struct = any;
-export type AppAction = any;
+// The application-specific state of an app instance, to be interpreted by the
+// app developer. We just treat it as an opaque blob; however since we pass this
+// around in protocol messages and include this in transaction data in disputes,
+// we impose some restrictions on the type; they must be serializable both as
+// JSON and as solidity structs.
+export type SolidityABIEncoderV2Struct = {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: can these be imported?

packages/node/test/integration/propose-install.spec.ts
```diff
@@ -29,7 +29,7 @@ describe("Node method follows spec - proposeInstall", () => {
   let nodeB: Node;
   let storeServiceB: IStoreService;
   let nodeConfig: NodeConfig;
-  let provider: BaseProvider;
+  let provider: JsonRpcProvider;
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: can you split out all the `BaseProvider` -> `JsonRpcProvider` changes

packages/dapp-high-roller/src/data/types.ts
snario: fixed

packages/dapp-high-roller/package.json
snario: Running `yarn test` does this automatically. I think its harmless.

packages/node/test/integration/propose-install.spec.ts
snario: sure

packages/cf.js/src/provider.ts
```diff
@@ -392,16 +392,14 @@ export class Provider {
...
-      oldState
+      newState
     } = nodeEvent.data as Node.UpdateStateEventData;
     const appInstance = await this.getOrCreateAppInstance(appInstanceId);
     const event = {
       type: EventType.UPDATE_STATE,
       data: {
         appInstance,
         newState,
-        oldState,
```
ebryn: why are you removing `oldState`?

packages/dapp-high-roller/package.json
joelalejandro: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll It's valid. We were using a wrong `types` package for Jest. (Using Jest 23 with typings for Jest 24).

packages/cf.js/src/provider.ts
snario: The caller should already have the old state, seems redundant to add in the result

PR #932, opened 2019-03-04
[playground] Updates canUseApps condition to also consider counterparty's FreeBalance state

joelalejandro:
This PR removes the multisig balance check from the `canUseApps` condition to use the counterparty's FreeBalance as part of the evaluation.

Also, it restricts the minimum viable deposit to `0.01 ETH`.

### DIFF
--- a/packages/playground/src/components/account/account-eth-form/account-eth-form.tsx
+++ b/packages/playground/src/components/account/account-eth-form/account-eth-form.tsx
@@ -25,6 +25,15 @@ export class AccountEthForm {
   handleSubmit(event) {
     event.preventDefault();
 
+    const value = Number(this.value);
+
+    if (value < this.min || value > this.max) {
+      this.error = `Please enter a value between ${this.min} and ${
+        this.max
+      } ETH.`;
+      return;
+    }
+
     this.submit.emit(event);
   }
 
--- a/packages/playground/src/components/account/account-eth-form/account-eth-form.tsx
+++ b/packages/playground/src/components/account/account-eth-form/account-eth-form.tsx
@@ -11,6 +11,7 @@ export class AccountEthForm {
   @Prop() header: string = "";
   @Prop() button: string = "";
   @Prop() disabled: boolean = false;
+  @Prop() min: number = 0.01;
   @Prop() max: number = 1;
   @Prop() available: BigNumber = { _hex: "0x00" } as BigNumber;
   @Prop({ mutable: true }) value: string | number = "";
@@ -37,7 +38,7 @@ export class AccountEthForm {
             value={this.value}
             error={this.error}
             disabled={this.disabled}
-            min={0}
+            min={this.min}
             max={Math.min(
               parseInt(ethers.utils.formatEther(this.available), 10),
               this.max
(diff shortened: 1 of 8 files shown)


packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -255,26 +257,50 @@ export class AppRoot {
...
+        ethers.utils.HDNode.fromExtendedKey(xkey).derivePath(String(k))
+          .publicKey
+      );
+
+    const balances = [
+      ethers.utils.bigNumberify(state.aliceBalance),
+      ethers.utils.bigNumberify(state.bobBalance)
+    ];
+
+    const [myBalance, counterpartyBalance] = [
```
cf19drofxots: Both of these seem to be getting the same balance?

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -255,26 +257,50 @@ export class AppRoot {
...
+      balances[
+        [state.alice, state.bob].findIndex(
+          address => address === getAddress(nodeAddress, 0)
+        )
+      ],
+      balances[
+        [state.alice, state.bob].findIndex(
+          address => address !== getAddress(nodeAddress, 0)
```
cf19drofxots: This `nodeAddress`  has to be the other party's public identifier.

joelalejandro: Is it necessary? I'm basically saying: `OK, you're *not* my address derived from my xpub, so you _have_ to be the counterparty's address`

packages/playground/src/components/app-root/app-root.tsx
joelalejandro: You mean `myBalance = balances[0]` and `counterpartyBalance = balances[0]`?

packages/playground/src/components/app-root/app-root.tsx
cf19drofxots: Err didn't catch the `!==`

Yeah I suppose this is fine.

packages/playground/src/components/app-root/app-root.tsx
cf19drofxots: Stale comment based on misunderstanding in the other comment

packages/playground/src/components/account/account-eth-form/account-eth-form.tsx
```diff
@@ -11,6 +11,7 @@ export class AccountEthForm {
   @Prop() header: string = "";
   @Prop() button: string = "";
   @Prop() disabled: boolean = false;
+  @Prop() min: number = 0.01;
```
cf19drofxots: Is this all that's required to restrict the min amount? I seem to be able to deposit less than this still... :/

joelalejandro: Nope, it isn't enough. Forgot something.

joelalejandro: This change only covered `/deposit`, not `/exchange`.

joelalejandro: Fixed!

PR #935, opened 2019-03-04
[playground] [node] Remove calls to provider.estimateGas()

joelalejandro:
This will enable full support in the Kovan testnet.

### DIFF
--- a/packages/node/src/methods/state-channel/deposit/operation.ts
+++ b/packages/node/src/methods/state-channel/deposit/operation.ts
@@ -92,13 +92,9 @@ export async function makeDeposit(
 
   const to = params.multisigAddress;
   const value = bigNumberify(params.amount);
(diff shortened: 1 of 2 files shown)


packages/node/src/methods/state-channel/deposit/operation.ts
```diff
@@ -92,13 +92,9 @@ export async function makeDeposit(
...
   const value = bigNumberify(params.amount);
+  const gasLimit = 30000;
 
-  const tx: TransactionRequest = {
-    to,
-    value,
-    gasPrice: await provider.getGasPrice(),
-    gasLimit: await provider.estimateGas({ to, value })
-  };
+  const tx: TransactionRequest = { to, value, gasLimit };
```
cf19drofxots: Oh we wanna keep the `gasPrice`

PR #944, opened 2019-03-05
[node, machine] Use BaseProvider only inside the Node

snario:


ebryn:
@snario why?

snario:
> @snario why?

The whole codebase uses `JsonRpcProvider` so that was the motivation. However I just realized `BaseProvider` can be the type inside the Node and Machine but then we can use `JsonRpcProvider` when we're passing it in (in those specific situations)

PR #945, opened 2019-03-05
[types] Remove AppAction type

snario:


ebryn:
It might be useful to better document `SolidityABIEncoderV2Struct` conceptually for others

ebryn:
Like, this is essentially `any` but for data passing to Solidity, right?

snario:
> Like, this is essentially `any` but for data passing to Solidity, right?

Yes

snario:
> It might be useful to better document `SolidityABIEncoderV2Struct` conceptually for others

Any suggestions as to where to put this

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
@ebryn I wrote the description which currently lives at https://github.com/counterfactual/monorepo/pull/945/files#diff-352353220a3ea3fe60b14e8e9d896d7eR8 ; the important things are that it can be serialized into JSON, and can be `abi.encode`-ed

ebryn:
> > It might be useful to better document `SolidityABIEncoderV2Struct` conceptually for others
> 
> Any suggestions as to where to put this

I think @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll's link above covers what I had in mind

PR #952, opened 2019-03-05
[machine,node] Correctly add limit field for virtual apps

snario:
The limit was set to Zero. It needs to be the sum of the deposits.

### DIFF
--- a/packages/machine/src/protocol/install-virtual-app.ts
+++ b/packages/machine/src/protocol/install-virtual-app.ts
@@ -5,7 +5,7 @@ import {
   SolidityABIEncoderV2Struct
 } from "@counterfactual/types";
 import { AddressZero, Zero } from "ethers/constants";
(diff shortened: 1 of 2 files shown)


packages/machine/src/protocol/install-virtual-app.ts
```diff
@@ -270,8 +272,10 @@ function createAndAddTarget(
     appInterface,
     {
       assetType: AssetType.ETH,
-      limit: Zero, // limit field is ignored, since limits are enforced by virtual app agreement
-      token: AddressZero
+      limit: bigNumberify(initiatingBalanceDecrement).add(
+        bigNumberify(respondingBalanceDecrement)
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: add a comment: the limit field is used for resolve, but not respected by `transfer`, instead the virtual app agreement enforces its own limit

PR #954, opened 2019-03-05
[playground-server] Restricts matchmaking to user whitelist

joelalejandro:
This PR modifies the `matchmakeUser` DB call to consider a MATCHMAKE_WHITELIST env-var, currently set to contain a list of bot users.


joelalejandro:
Misinterpreted requirement.

PR #959, opened 2019-03-05
[playground] Adds "Log out" feature

joelalejandro:
deepin-screen-recorder_vivaldi-stable_20190305132923


snario:
Can we hide the button inside the account page vs keeping it front and center like that?

joelalejandro:
@snario Seems a bit uncomfortable to use inside the Account page. But I guess we can do that.

snario:
I just think it shouldn't be "always present"

ebryn:
We can fix in subsequent PRs

PR #963, opened 2019-03-05
[playground][playground-server][apps][node][dapp-tic-tac-toe] Support Playground functionality on Kovan

cf19drofxots:
### Description

This adds support for the Kovan network to be selected when using the Playground.

Note that High Roller hasn't been fully updated to reflect this change yet.

Notable updates:

To use Kovan, simply set `ETHEREUM_NETWORK=kovan` as an environment variable before running `yarn run:playground`, and be sure to have your browser wallet set to use the Kovan testnet.


snario:
We should try to get users to only use one network though, not one or the other. I prefer Kovan.

PR #964, opened 2019-03-06
Miscellaneous cleanups

snario:
- Add env var that is defaulted in the code
- Fix UI error preventing showing of balances
- Show alert for weird edge case

### DIFF
--- a/packages/playground/src/components/layout/layout-header/header-account/header-account.tsx
+++ b/packages/playground/src/components/layout/layout-header/header-account/header-account.tsx
@@ -75,6 +75,9 @@ export class HeaderAccount {
       await window["ethereum"].enable();
     } catch {
       console.error("Was not able to call `window.ethereum.enable()`");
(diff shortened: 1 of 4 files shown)


packages/playground/src/components/layout/layout-header/header-account/header-account.tsx
```diff
@@ -75,6 +75,9 @@ export class HeaderAccount {
       await window["ethereum"].enable();
     } catch {
       console.error("Was not able to call `window.ethereum.enable()`");
+      window.alert(
+        "Your browser does not support enabling your wallet programatically. Please unlock your Web3 wallet and try again."
```
ebryn: `window.alert`?

snario: Yeah it's a massive edge case so don't want to spend too much time on this with proper UI etc

PR #980, opened 2019-03-07
WIP Get high roller to work

snario:


packages/machine/src/models/app-instance.ts
```diff
@@ -106,7 +106,7 @@ export class AppInstance {
     const latestState = json.latestState;
     for (const key in latestState) {
       // @ts-ignore
-      if (latestState[key]["_hex"]) {
+      if (latestState[key] && latestState[key]["_hex"]) {
```
snario: This was added because `playerName` was `undefined` and is also not part of the app state. It is UI state but for some reason it was added into the same object that gets passed to the install protocol's `initialState` parameter.

packages/playground/src/components/dapp-container/dapp-container.tsx
```diff
@@ -93,7 +93,9 @@ export class DappContainer {
     window.addEventListener("message", this.handlePlaygroundMessage.bind(this));
 
     // Callback for passing an app instance, if available.
-    iframe.addEventListener("load", this.sendAppInstance.bind(this));
+    setTimeout(() => {
+      this.sendAppInstance();
+    }, 3000);
```
snario: We tried to just set a timeout in case it was an async bug.

packages/dapp-high-roller/src/data/game-types.ts
```diff
@@ -13,7 +13,6 @@ export type HighRollerAppState = {
   commitHash: string;
   playerFirstNumber: number;
   playerSecondNumber: number;
-  playerNames?: string[];
```
snario: This should not be inside this object.

cf19drofxots:
This looks stale - closing. Re-open if necessary.

PR #994, opened 2019-03-08
[playground] White wallet svg

snario:
Screen Shot 2019-03-09 at 00 49 15


joelalejandro:
@snario does it look black in mobile?

snario:
No

joelalejandro:
Then we need a different approach. I think you can mutate the vector's color with CSS and stroke/fill.

snario:
See #1002

PR #1000, opened 2019-03-11
[playground][playground-server][apps][node][dapp-tic-tac-toe] Support Playground functionality on Kovan (2)

cf19drofxots:
Description
This adds support for the Kovan network to be selected when using the Playground **and assumes Kovan is the only network being used**

Note that High Roller hasn't been fully updated to reflect this change yet.

Notable updates:

To use Kovan, simply set ETHEREUM_NETWORK=kovan as an environment variable before running yarn run:playground, and be sure to have your browser wallet set to use the Kovan testnet.


packages/playground/src/components/webthree-connector/webthree-connector.tsx
```diff
@@ -5,7 +5,7 @@ import { WalletState } from "../../data/wallet";
 
 const { Web3Provider } = ethers.providers;
 
-const permittedNetworkIds = ["3", "42"];
+const permittedNetworkIds = ["42"];
```
snario: Instead of `"42"` can you use a constant variable assigned to this

PR #1005, opened 2019-03-12
[monorepo] set setstate timeouts to 30 days

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
nonce registry doesn't have a default timeout, I believe we currently only use 0

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
I don't think ethers has one, couldn't find it in https://docs.ethers.io/ethers.js/html/api-utils.html

In which package should the constant be?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
need to fix the tests, or else `install-than-set-state.spec.ts` will wait for 30 days

PR #1015, opened 2019-03-12
[playground] add seo-able images

patience-tema-baron:
Looks like twitter cards don't support svg, so I added a 1000x1000px jpeg for twitter and a 1200x630 banner image for og. This should work, but I can't validate against localhost.


snario:
From a quick glance those dimensions seem wrong — there are specific requirements for twitter, fb, LinkedIn, etc

PR #1020, opened 2019-03-13
[contracts] remove truffle compiler version to avoid confusion

AlexXiong97:
Since all contracts are upgraded by bot, and we are using waffle instead of truffle for contract build, it's better to remove the compiler version to avoid confusion.

### DIFF
--- a/packages/contracts/truffle-config.js
+++ b/packages/contracts/truffle-config.js
@@ -6,7 +6,7 @@ require("dotenv").config();
 module.exports = {
   compilers: {
     solc: {
-      version: "0.5.3",
+      version: "0.5.5",
     },
   },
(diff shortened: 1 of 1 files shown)


snario:
Actually we kind of need this because Truffle will otherwise default to the version locally installed which may not be the latest one that we want. (I think, if I recall why I added it)

AlexXiong97:
Oh. I see. `yarn migrate` is still using `truffle migrate`, which will fail without the statement. 

then, we should upgrade that to `0.5.5` as well then. I'll amend the commit later. But `yarn migrate` would still fail after the modification. It failed on deploying `Migration.sol`, I'm trying to figure out why.

PR #1031, opened 2019-03-13
[machine] Bump version

cf19drofxots:
Forgot to bump version in https://github.com/counterfactual/monorepo/pull/1029


cf19drofxots:
503 from yarn package registry - retrying once registry is available again

PR #1045, opened 2019-03-14
[playground] Start polling after 24 seconds for multisig

snario:


packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -417,12 +419,33 @@ export class AppRoot {
...
+        onMultisigMinedHasBeenCalled = true;
+      }
+    };
+
+    provider.once(transactionHash, onMultisigMined);
+
+    setTimeout(() => {
+      if (!onMultisigMinedHasBeenCalled) {
+        console.log("Tx event not emitted within 24s, polling every 5s now");
+        const poll = setInterval(async () => {
```
ebryn: i generally recommend avoiding setInterval in favor of setTimeout's that schedule another setTimeout. this tends to be better for testing.

PR #1063, opened 2019-03-15
[playground] Modifies deposit condition

joelalejandro:
If at any point the local balance or the counterparty's balance get below 0.01, you'll not be seeing the apps-list anymore. You'll get an error message offering you to redirect to the exchange path.


joelalejandro:
@snario Please test with a different mnemonic before merging is possible. I think it covers most cases.

PR #1073, opened 2019-03-15
[playground] [playground-server] Implements Heartbeats API

joelalejandro:
:heartbeat: The Heartbeats API is called every 30 seconds (and as soon as the Playground is opened) to check the current data schema version and if there is a maintenance underway.

If the API fails to reply, the frontend will consider the backend is not available, and therefore won't run the Playground.

**Maintenance mode**
image

If there's a change in the schema's version number, it'll trigger the following modal, and once the user clicks through the button, the Node storage of that user will be wiped out. 

**Schema mismatch**
image

### How to enable maintenance mode
Set the env-var `PLAYGROUND_MAINTENANCE_MODE` to `"true"` in the Playground Server.

### How to change schema version
Set the env-var `PLAYGROUND_SCHEMA_VERSION` to the new version number.


ebryn:
The upgrade modal should probably say something about the fact that their channels will be losing all of their state, right?

joelalejandro:
@ebryn Yeah. It should also probably do something about re-create the state channel, because otherwise this will render the account unusable.

snario:
We should link to a video or something (maybe my EthCC talk as a placeholder) or maybe the Medium blog post in the "we'll be back soon" modal. "While you're waiting for us to get the demo back up; check out this video" or something.

joelalejandro:
@snario Sure, tell me what you want it to show and I'll add it. Should this make the API delete the user account so it can re-register?

snario:
> Sorry! We're currently working on a few things behind the scenes to keep the demo functional. Please come back later. In the meantime, follow us on Twitter @statechannels to learn more and keep up to date on the project.

?

PR #1074, opened 2019-03-15
[dapp-high-roller]: Forcibly restarts the game if you click "Exit"

joelalejandro:


ebryn:
Why though?

joelalejandro:
@ebryn For some reason it remains in the same state even if remove the related state to the appInstance. Didn't want to delay this too much, we can iterate and do something better later.

PR #1080, opened 2019-03-16
[node] More versatile and verbose logging

cf19drofxots:


cf19drofxots:
Unfortunately not. PG server uses Logepi which doesn't have out-of-box support for running in browsers (its README even says it's meant for Node.js apps)

PR #1081, opened 2019-03-18
[pg] Rename email to twitter to avoid issues

snario:


emansipater:
You didn't remove the email validity check, which runs after the wallet authorisation modal.

packages/playground/src/components/account/account-edit/account-edit.tsx
```diff
@@ -76,7 +76,7 @@ export class AccountEdit {
             value={this.changeset.username}
           />
           <form-input
-            label="Email address"
+            label="Twitter handle"
```
ebryn: :-1:

ebryn: Can we please leave email? I want a way for us to contact users in the future

PR #1082, opened 2019-03-18
[pg] Different attempt at waiting for channel creation

snario:
Seems more logical that we should be doing this.


packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -447,33 +447,8 @@ export class AppRoot {
...
-        const poll = setInterval(async () => {
-          if (await provider.getTransactionReceipt(transactionHash)) {
-            clearInterval(poll);
-            await onMultisigMined();
-          }
-        }, 5000);
-      }
-    }, TWO_BLOCK_TIMES_ON_AVG_ON_KOVAN);
+    const node = CounterfactualNode.getInstance();
+    node.once(Node.EventName.CREATE_CHANNEL, this.fetchMultisig.bind(this));
```
snario: We actually don't need to fetch the multisig at all b/c the value will be in the args of the event from the node.

PR #1084, opened 2019-03-18
[node] store non-withdraw commitments in the same namespace

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/machine/src/protocol/install-virtual-app.ts
```diff
@@ -178,9 +192,7 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
       };
     },
 
-    Opcode.IO_SEND,
-
-    Opcode.WRITE_COMMITMENT
```
snario: Don't remove

packages/node/package.json
```diff
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "build": "rollup -c",
-    "test": "jest --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
+    "test": "jest --setupFiles dotenv-extended/config --runInBand --bail --forceExit --verbose",
```
snario: @cf19drofxots there was a reason you removed this no?

cf19drofxots: It was an attempt to speed up the tests if no logs needed to be buffered & printed. Didn't seem to make a difference though.

PR #1086, opened 2019-03-18
[node] pass --verbose to jest test

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Without these it seems that console.logs are not being printed (e.g. on master)

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "build": "rollup -c",
-    "test": "jest --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
+    "test": "jest --setupFiles dotenv-extended/config --runInBand --bail --forceExit --verbose",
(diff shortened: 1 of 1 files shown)


packages/node/package.json
```diff
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "build": "rollup -c",
-    "test": "jest --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
+    "test": "jest --setupFiles dotenv-extended/config --runInBand --bail --forceExit --verbose",
```
cf19drofxots: Typically these options get set in `jest.config.json`

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: including `--runInBand`, etc?

snario: `--runInBand` is an exception https://github.com/facebook/jest/issues/3215

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
implemented in #1082

PR #1096, opened 2019-03-18
[pg-s] Add some basic logs for user activity for observation

snario:


snario:
Note, this is temporary and we should do this kind of thing by writing functions against the main logs and not using the source code.

PR #1102, opened 2019-03-18
[pg-server, pg] Adds a corrupt waller state UI and delete account button

snario:
Fixed the error where your account no longer has a valid state channel in the playground DB

Capture d’écran 2019-03-18 à 17 26 28

### DIFF
--- a/packages/playground/src/components/app-root/app-root.tsx
+++ b/packages/playground/src/components/app-root/app-root.tsx
@@ -277,6 +277,7 @@ export class AppRoot {
     ) as string;
 
     if (!token) {
+      console.error("Couldn't delete account; no token was provided");
       return;
     }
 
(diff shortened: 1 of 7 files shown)


packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -271,6 +271,26 @@ export class AppRoot {
     return loggedUser;
   }
 
+  async deleteAccount(): Promise<void> {
+    const token = window.localStorage.getItem(
+      "playground:user:token"
+    ) as string;
+
+    if (!token) {
+      return;
```
cf19drofxots: Log something to give feedback about delete not being performed.

packages/playground/src/components/app-root/app-root.tsx
```diff
@@ -271,6 +271,26 @@ export class AppRoot {
...
+    }
+
+    const user = await PlaygroundAPIClient.getUser(token);
+
+    try {
+      await PlaygroundAPIClient.deleteAccount(user);
+    } finally {
+      this.updateAccount({ hasCorruptStateChannelState: false });
```
cf19drofxots: This should come immediately after the `deleteAccount` has successfully returned instead of being in the finally block.

PR #1110, opened 2019-03-19
[playground] change message

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/playground/src/components/account/account-register/account-register.tsx
```diff
@@ -217,7 +217,7 @@ export class AccountRegister {
       ready: "Register",
       awaitingForWallet: "Check Wallet...",
       creatingAccount: "Creating your account...",
-      deployingMultisig: "Creating state channel..."
+      deployingMultisig: "Deploying state channel contract..."
```
cf19drofxots: There's a tradeoff to be made here.

Since this is more user facing, a user ideally shouldn't care about the technical details (i.e. that it's a deployment or that it's a contract). To them it's just a realization of the concept of a channel being created.

cc @snario @ebryn

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: To a nontechnical user the confusion over what "creating state channel" means is even worse

snario: I think it isn't worth changing this right because CREATE2 will remove the entire process and the audience presently of the Playground demo still is developers.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: If the audience is developers, `deploying contract` is better for them as well

snario: OK @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll does that string fit within the button width without a newline?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `deploying contract` is shorter than `creating state channel`, but `deploying state channel contract` is longer

snario:
```suggestion
      deployingMultisig: "Deploying contract..."
```

PR #1126, opened 2019-03-21
[playground] Setup Sentry for Playground when TIER != dev

Alonski:
### Description

Setup Sentry for Playground when TIER != dev

### Related issues


Alonski:
@snario Do you mean having something like
```js
window["config"] = {
  ENV: "dev"
}
```

Alonski:
@snario 
One thing that I was thinking about now is that this:
    ```if (TIER === "dev") {
      // Do nothing
    } else {
      window["Sentry"].init({
        dsn: "https://6037586d37124e518f4718d9dd46b18b@sentry.io/1383439"
      });
    }```
Gets replaced at build time.
If `TIER === "dev"`, the final output will be `// Do nothing`
So I don’t think I will be replacing everything to use the Global I am creating. Which should only be used during runtime.
This is only for production builds btw

PR #1129, opened 2019-03-21
[contracts] fix #1019 renaming appState and appChallenge

### LINKED ISSUE #1019: [contracts] AppState, AppChallenge mixed naming caused confusion
**TL;DR**: there's inconsistency in terms of the naming of `appState`, `appChallenge` and consequently `stateHash` and `appStateHash`, which might cause confusion.

**Problem**:
Based on def given in [`LibStateChannelApp`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/libs/LibStateChannelApp.sol#L26):
- `AppChallenge` is the "overall" state of a certain app
- `appState` refers to the encoded appState specific to that application, (i.e. `abi.encode(AppState)` where `AppState` is a user-defined struct

Here's the confusion kicks in: sometimes in the `/mixin` suite, `appState` are used to refer to the "overall" state of an app, which is supposed to be `appChallenge` if we were to be consistent. 

E.g.: in [`MAppRegistryCore`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/mixins/MAppRegistryCore.sol#L14): the variable `appState` 

another demo of such possible confusion: in [`MixinSetState`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/mixins/MixinSetState.sol#L18), can you tell which state (challenge of app) is the `stateHash` referring to at first glance? What about [this line](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/mixins/MixinSetState.sol#L65)?

**Proposed Action**:
I do realize the change of naming might cause a bit cosmetic fix on multiple contracts, but for better readability, I propose the following changes: 
1. change `appState` mentioned above to `appChallenges`
2. change `computeStateHash` in the same file to `computeAppChallengeHash`
3. add additional `function getEncodedAppStateHash(AppState memory appState) public pure returns (bytes encodedState)` to `CounterfactualApp.sol`, because `appState` is required to pass to all function in `MAppCaller`, but in current version such helper function is not required for the developers to implement, which I think it should be. So this function should be put together with `getTurnTaker` etc. as one of the mandatory function interface.

AlexXiong97:
most changes in `contracts` and their testing scripts, with rippling effect to `machine` package (thus minor surgery there)


AlexXiong97:
@snario I'm not sure why CD failed. from my forked repo, all CI workflow passed. Can you help me take a look? Wish to get this PR into master soon, as it brings clarity in naming. Thanks

Issue #996, opened 2019-03-09
[playground-server] missing postgres db schema?


kosecki123:
Registration step throws 
```
@counterfactual/playground-server: JSONAPI-TS:  { error: relation "users" does not exist
```
after typing in the username/email and signing it with MetaMask.

Is that due to lack of DB initialization script?

I'm running the postgres using docker `docker run --name postgress -p 5432:5432 -d postgres`

Step to reproduce:
- clean monorepo clone
- `yarn` and `yarn build`
- `yarn run:playground`
- try to register

snario:
Sorry, this needs to be automated in the code.

```sql
CREATE SCHEMA playground_db AUTHORIZATION <user>;

CREATE TABLE playground_db."users"
(
    id uuid NOT NULL,
    username character varying(255) COLLATE pg_catalog."default" NOT NULL,
    email character varying(255) COLLATE pg_catalog."default" NOT NULL,
    eth_address character varying(42) COLLATE pg_catalog."default",
    multisig_address character varying(42) COLLATE pg_catalog."default",
    transaction_hash character varying(66) COLLATE pg_catalog."default",
    node_address character varying(128) COLLATE pg_catalog."default",
    CONSTRAINT users_pkey PRIMARY KEY (id),
    CONSTRAINT uk_users__username UNIQUE (username)

)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

ALTER TABLE playground_db.users OWNER to <user>;
```

snario:
I think #1143 has fixed this.

Issue #943, opened 2019-03-05
[playground] auto-start postgresql db when running playground locally


emansipater:
Currently running the playground locally requires a postgresql instance.  We should auto-start one on `run:playground` or in `yarn start` of `playground-server

emansipater:
note https://github.com/counterfactual/monorepo/issues/996

snario:
Addressed in #1143

PR #1143, opened 2019-03-26
[playground-server] Add postgres db on playground start

cf19drofxots:
### Description

If the server cannot connect to a database at the default connection string, it informs the user and exits:
```
[timestamp:2019-03-28T18:16:37.224Z] [level:error] [message:Failed to connect to database] [connection:postgresql://postgres@localhost:5432/postgres]
[timestamp:2019-03-28T18:16:37.225Z] [level:error] [message:Spin up a database at the connection string or specify an existing database via exporting the DB_CONNECTION_STRING environment variable]
```

The README is updated to inform the user on how to proceed if they hit the above error message. More specifically, pointing them to do one of:
- configure the `DB_CONNECTION_STRING` to a locally running database
- install postgres if they don't have it
- if they have docker then running a docker service via docker-compose

If a database connection is established, but the desired schema is not available, it tries to perform a migration to create the needed table, and informs the user if it fails to create table then exits.

### Related issues

Resolves the following issues:
https://github.com/counterfactual/monorepo/issues/943
https://github.com/counterfactual/monorepo/issues/996


cf19drofxots:
> Ideally we could offer it as an optional process you could run if there wasn't already a postgresql installation on your laptop. However if I already have it running from a brew install I'd rather just use that.

I'm not a fan of introducing dependencies if they're not needed either, but in this case I thought detecting a postgres installation cross-platform can become unwieldily hence introducing a dependency that'll provide a much more reliable dev experience being higher value.

@kosecki123 what are your thoughts on this approach?

kosecki123:
> > Ideally we could offer it as an optional process you could run if there wasn't already a postgresql installation on your laptop. However if I already have it running from a brew install I'd rather just use that.
> 
> I'm not a fan of introducing dependencies if they're not needed either, but in this case I thought detecting a postgres installation cross-platform can become unwieldily, hence introducing a dependency that'll provide a much more reliable dev experience being higher value.
> 
> @kosecki123 what are your thoughts on this approach?

I think having both Postgres and firebase is unnecessary unless there is a strong argument against that.

cf19drofxots:
> I think having both Postgres and firebase is unnecessary unless there is a strong argument against that.

Yes, using firebase was an earlier decision to get things up and running quickly since it could provide sufficient storage and message-relaying functionality.

We'll likely be moving away from firebase and will use a WebSocket server for messaging and Postgres for storage.

This PR was just an immediate attempt to mitigate the postgres start up issue before we get to the bigger overhaul.

kosecki123:
> > Ideally we could offer it as an optional process you could run if there wasn't already a postgresql installation on your laptop. However if I already have it running from a brew install I'd rather just use that.
> 
> I'm not a fan of introducing dependencies if they're not needed either, but in this case I thought detecting a postgres installation cross-platform can become unwieldily, hence introducing a dependency that'll provide a much more reliable dev experience being higher value.
> 
> @kosecki123 what are your thoughts on this approach?

As for the docker dependency, I use Docker whenever I can, it's convenient. On the other hand, some folks may have Postgres installed using other ways. I guess the easiest would be to make running Postgres from yarn as a separate command, so you can handle both cases.

snario:
@cf19drofxots A better approach might be to make it such that if `${POSTGRES_URI}` cannot be connected to by the software (i.e., it is not already running) then the start script could _ask_ the user: `PostgreSQL not detected. Start an instance using docker? [Y/n]` perhaps? Even simpler, we could just have a script called `start_dockerized_postgresql.sh` that we mention in the startup script as a backup to run in case there is no PostgreSQL running. However I do strongly advocate we not make docker a dependency.

cf19drofxots:
@snario @kosecki123 Updated the change set to reflect giving the user more informed feedback and optionality on getting the server running with a db.

packages/playground-server/docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+version: "3"
+services:
+  postgres:
+    image: postgres:11.2-alpine
+    ports:
+      - "5432:5432"
```
snario: This won't be persistent, will it? Don't we need to mount a volume?

cf19drofxots: https://github.com/counterfactual/monorepo/pull/1171

PR #1144, opened 2019-03-27
fix #1058 by adding package resolution to ethers.js

### LINKED ISSUE #1058: [contracts] typescript lint error Web3Provider type not assignable
[`packages/contracts/test/static-call.spec.ts`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/test/static-call.spec.ts):

Here's a weird error message from my editor:
> Type 'import(".../counterfactual/monorepo/node_modules/ethereum-waffle/node_modules/ethers/providers/web3-provider").Web3Provider' is not assignable to type 'import(".../counterfactual/monorepo/node_modules/ethers/providers/web3-provider").Web3Provider'.  Types have separate declarations of a private property '_sendAsync'.

Here's the screenshot:
Screenshot from 2019-03-15 16-49-26

Confirmed with @snario with the error, but as all tests pass, nothing serious was affected, left this issue for future investigation.

AlexXiong97:


AlexXiong97:
@snario ready for review and merge.

PR #1146, opened 2019-03-27
[specs] Add the specs repo as a package in monorepo

snario:


greenkeeper.json
```diff
@@ -12,6 +12,7 @@
         "packages/node-provider/package.json",
         "packages/node/package.json",
         "packages/playground/package.json",
+        "packages/specs/package.json",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: no such file - but I'm not sure what greenkeeper does with this

PR #1147, opened 2019-03-27
[playground-server] Optional persistence for local firebase db

### LINKED ISSUE #1136: [playground-server] Lack of local firebase persistence
Default development configuration assumes usage of local `firebase-server` instance which by default is in-mem only.

This means you need to set up your environment every time you restart the `playground-server`.

I guess at least 2 options could be considered:
- allow development setup to use remote firebase
- use Postgres to store latest firebase state (on firebase close) and then use it as the an initial state when starting firebase

kosecki123:
### Description

Uses local postgres db to store the snapshot from local firebase db. In order to enable the feature the env parameter `PLAYGROUND_PERSISTENCE_ENABLED=true` has to be set. 

### Related issues

Fixes #1136


cf19drofxots:
@kosecki123  see https://github.com/counterfactual/monorepo/pull/1143 :P

kosecki123:
> @kosecki123 see #1143 :P

Ow, yeah, forgot the DB scheme

cf19drofxots:
> > @kosecki123 see #1143 :P

> Ow, yeah, forgot the DB scheme. Will add after #1143 merge to master.

@kosecki123 #1143 has been merged

packages/playground-server/src/node.ts
```diff
@@ -40,30 +50,86 @@ export class LocalFirebaseServiceFactory extends FirebaseServiceFactory {
...
+
+  static async create(host: string, port: string) {
+    const snapshot = await getPlaygroundSnapshot();
+
+    return new LocalPersistentFirebaseServiceFactory(host, port, snapshot);
+  }
+
+  async closeServiceConnections() {
+    const snapshot = await this.firebaseServer.getValue();
+    storePlaygroundSnapshot(snapshot);
```
cf19drofxots: Does this not need to be awaited?

packages/playground-server/src/node.ts
```diff
@@ -40,30 +50,86 @@ export class LocalFirebaseServiceFactory extends FirebaseServiceFactory {
...
+      ]!
+    })
+  );
 } else {
   confirmLocalFirebaseConfigurationEnvVars();
-  serviceFactory = new LocalFirebaseServiceFactory(
-    process.env.FIREBASE_SERVER_HOST!,
-    process.env.FIREBASE_SERVER_PORT!
-  );
+  if (process.env.PLAYGROUND_PERSISTENCE_ENABLED) {
```
cf19drofxots: It's probably a good idea to mention the existence of this env var and its function as part of the README 

(similar to how you've described this feature in the OP)
> In order to enable the feature the env parameter PLAYGROUND_PERSISTENCE_ENABLED=true has to be set.

packages/playground-server/src/node.ts
```diff
@@ -40,30 +50,86 @@ export class LocalFirebaseServiceFactory extends FirebaseServiceFactory {
...
+
+  async closeServiceConnections() {
+    const snapshot = await this.firebaseServer.getValue();
+    storePlaygroundSnapshot(snapshot);
+
+    await this.firebaseServer.close();
+  }
+}
+
+export let serviceFactory: Promise<IClosableFirebaseServiceFactory>;
```
cf19drofxots: Might be better to name this something like `unresolvedServiceFactory` so that when we `await` it in its usage, we can refer to the resolution as just `serviceFactory` instead of `resolvedServiceFactory`

Small nit, not too opinionated either way but just mentioning it.

kosecki123: Good point, in the end I chose to name it `serviceFactoryPromise` so it's clear that's a Promise and has to be awaited before use.

Issue #1145, opened 2019-03-27
[machine] Fold the machine package as a subfolder / library within the Node


snario:
The node and machine code are always run in the same javascript process, but they are in separate packages and built separately. This has many disadvantages, e.g. certain type errors are only caught at the rollup stage.

snario:
#1218

PR #1149, opened 2019-03-28
[machine] upgrade jest

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Needed for `extraGlobals`

### DIFF
--- a/packages/machine/package.json
+++ b/packages/machine/package.json
@@ -26,7 +26,7 @@
     "dotenv-safe": "^6.1.0",
     "ethereum-waffle": "2.0.7",
     "ganache-cli": "6.4.1",
-    "jest": "23.6.0",
+    "jest": "24.1.0",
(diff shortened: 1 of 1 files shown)


snario:
There's some weird stuff as mentioned #841

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
I think this is a weirder issue than that - upgrading jest on **machine** causes a require to fail in **node**. But node was already using jest 24.

snario:
Let's do this: #1145

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
We will probably still run into the same problem, is my guess

snario:
Check my branch #1138, too.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Woot

PR #1203, opened 2019-04-01
[contracts, machine, specs] Fix mistakes in comments

nikeshnazareth:
### Description

While looking through the code I found some minor non-functional typos and mismatches between comments and code. This PR addresses them.


nikeshnazareth:
The PR template requests that I add labels. Am I correct in assuming this only applies to people with write access to the repository? I could not find a way to add a label to the PR using my own repo.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Thanks! I think you're right that labels require write access - I'll look into that

PR #1204, opened 2019-04-01
fix #393 use solc binary for fast contract compilation 

### LINKED ISSUE #393: [contracts] use `solc` for 10x faster contract compilation on CI
When #392 is done we can use `ethereum-waffle`'s latest version to use `solc` which speeds up contracts compilation 10x

AlexXiong97:


AlexXiong97:
I'm not so sure about the CD part, the Netlify deployment seems a probabilistic process, in earlier PR, I also have similar issue where they failed for no particular reason, and when I force push (amend without edit) to trigger another checking, they go through.

Not sure what can I do. CI tests should all pass though.

snario:
The build command is [failing](https://app.netlify.com/sites/high-roller-staging/deploys/5ca226be093c3ecd3e8e7237) on Netlify:

```
7:59:24 AM: /bin/sh: 1: solc: not found
```

It's because `netlify.sh` is calling `yarn build` on the `contracts` package but `solc` is not found in there because of this change:

Capture d’écran 2019-04-01 à 09 27 50

I think you need to make two `waffle.json` files; one for CI when it knows it has `solc` natively installed and one for local and Netlify when we don't know that we have a native `solc`.

Also shouldn't that change be in the `apps` package's `waffle.json` too?

packages/contracts/waffle.js
```diff
@@ -0,0 +1,85 @@
...
+  if (process.env.CI || process.env.NATIVE_SOLC) {
+    // use native solc binary for fast compilation in CI
+    waffleConfig.compiler = "native";
+
+  } else {
+
+    const solcjsVersion =
+    'v'+/([0-9a-z\.\+]+)\.Emscripten\.clang/g.exec(require("solc").version()).splice(1)[0] ;
+
+    const MigrationsContract = fs.readFileSync("contracts/Migrations.sol", "utf-8");
```
snario: What's the purpose of this?

AlexXiong97: regex matching. To get the version of solcjs installed, use `require.resolve` function from nodejs, 
which has a format of `0.x.x+commit.xxxx.Emscripten.clang` , however to feed it into the `waffle config`, we need to adhere to `v0.x.x+commit.xxx`, thus the cosmetic cutting

.circleci/config.yml
```diff
@@ -8,23 +8,45 @@ defaults: &defaults
...
+  restore_cache:
+    key: dependency-cache-{{ checksum "yarn.lock" }}
+
 jobs:
   build:
-    <<: *defaults
+    working_directory: /home/circleci/project
+    # use customized docker (ubuntu, yarn 1.15, node10, solc binary)
+    docker:
+      - image: web3docker/counterfactual:0.0.1
```
snario: Who owns this?

README.md
```diff
@@ -82,6 +82,9 @@ yarn
 
 ### Building packages
 
+Make sure you have installed `solc` compiler binary on your machine
```
snario: Shouldn't this state that it is optional?

packages/contracts/package.json
```diff
@@ -34,13 +34,15 @@
...
+    "ethereum-waffle": "2.0.9",
     "ethers": "4.0.27",
     "ethlint": "1.2.3",
+    "fs": "^0.0.1-security",
     "ganache-cli": "6.4.1",
     "openzeppelin-solidity": "2.2.0",
+    "request": "^2.88.0",
     "shx": "^0.3.2",
-    "solc": "0.5.7",
+    "solc": "^0.5.7",
```
snario: Keep it fixed vs `^`

.circleci/config.yml
AlexXiong97: ah.. That's me... security concern?

README.md
AlexXiong97: yes, I should change this to optional

AlexXiong97:
Sorry for the dragged-out PR history. The git diff in File Change is clear, 

In case anyone interested in the previous code review and change history:
1. I overthink on the task, my original logic is the following:
in CI environment of explicitly said `NATIVE_SOLC=true` in `.env`, then use solc binary;
otherwise use the solcjs specified in `package.json`; 
now in case the solcjs is not updated, it will detect (parse out) version in `Migrations.sol` (assumption is that this contract will use the same version with the rest of the contracts) and then fetch the long version of the solc name asynchronously and then feed into waffle config

^^^ this apparently is unnecessarily hairy and thanks to Liam's comment, we choose to require a `solcjs` version in _node_resolutions_ instead, and directly load the solcjs

2. because some very weird dependency versions in `contracts` and `machine`, result in CI test failure, which is too bizarre to me and took me quite some time to figure out. Thus the messy history. 

3. during the grind on a simple task, I accidentally git track a subfolder not mean for this PR... thus later removed it. 

Hope this clear up some confusion. In any case, just reviewing the final git diff would suffice. thanks !

.circleci/config.yml
snario: Could be build the `Dockerfile` in the CI process with a tag and then save the image in CircleCI cache to avoid needing to make an external image pull?

packages/contracts/package.json
```diff
@@ -34,13 +34,13 @@
...
     "eth-gas-reporter": "0.1.12",
-    "ethereum-waffle": "2.0.8",
+    "ethereum-waffle": "2.0.9",
     "ethers": "4.0.27",
     "ethlint": "1.2.3",
     "ganache-cli": "6.4.1",
     "openzeppelin-solidity": "2.2.0",
     "shx": "^0.3.2",
-    "solc": "0.5.7",
+    "solc": "^0.5.7",
```
snario:
```suggestion
    "solc": "0.5.7",
```

packages/contracts/waffle.js
```diff
@@ -0,0 +1,27 @@
...
+  if (process.env.CI || process.env.NATIVE_SOLC == "true") {
+    // use native solc binary for fast compilation in CI
+    waffleConfig.compiler = "native";
+
+  } else {
+    const solcjsVersion =
+    'v'+/([0-9a-z\.\+]+)\.Emscripten\.clang/g.exec(require("solc").version()).splice(1)[0];
+    waffleConfig.solcVersion = solcjsVersion;
+  }
+  console.log(waffleConfig);
```
snario:
```suggestion
```

AlexXiong97:
please don't merge yet. Still figuring out the docker image caching. will notify reviewer when ready.

AlexXiong97:
since [`docker_layer_caching`](https://circleci.com/docs/2.0/docker-layer-caching/) is a premium feature for circleci, optimization on that would not be addressed here. 

Just a result report:
Screenshot from 2019-04-06 14-33-27
Execution time for `build` job now is roughly 2'30'' whereas the master branch took 3' 

tiny improvement. but still @snario this PR should be ready for merge

PR #1218, opened 2019-04-02
[monorepo] move machine package into node

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR moves the machine package into the node package.

## Summary of changes

- `packages/machine/src` -> `packages/node/src/machine`
- `packages/machine/test` -> `packages/node/test/machine`
- entries in `packages/machine/package.json` added to `packages/node/package.json`
- `jest-dot-reporter` disabled
- machine test script moved to `node/scripts/test.sh` and now only tests the machine integration tests (no unit tests). Note that passing regex pattern now no longer works.
- imports updated
- `proxyFactory.on` changed to `proxyFactory.once`, otherwise the callback is called multiple times. Not sure why this didn't happen in the old codebase.


packages/contracts/contracts/libs/LibSignature.sol
```diff
@@ -46,7 +46,7 @@ contract LibSignature {
     address lastSigner = address(0);
     for (uint256 i = 0; i < signers.length; i++) {
       require(
-        signers[i] == recoverKey(signatures, txHash, i), "Invalid signature"
+        signers[i] == recoverKey(signatures, txHash, i), "Invalid signature passed to verifySignatures"
```
snario: Probably line too long

packages/node/package.json
```diff
@@ -6,27 +6,37 @@
   "types": "dist/src/index.d.ts",
   "license": "MIT",
   "scripts": {
-    "build": "rollup -c",
-    "test": "jest --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
+    "build": "tsc -p . && rollup -c",
+    "test": "yarn run test:no-ganache && yarn run test:with-ganache",
+    "test:with-ganache": "./scripts/test.sh",
+    "test:no-ganache": "jest --testPathIgnorePatterns=test/machine/integration --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
```
snario: maybe `test:machine` and `test:rpc`?

packages/node/tslint.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": ["../../tslint.json"]
+  "extends": ["../../tslint.json"],
+  "rules": {
+    "object-literal-shorthand": false
```
snario:
```suggestion
    // Needed because `ProtocolExecutionFlow` uses numbered object keys
    "object-literal-shorthand": false
```

packages/node/truffle-config.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
   },
   compilers: {
     solc: {
-      version: "../../node_modules/solc",
+      version: "../contracts/node_modules/solc",
```
snario: Stick with `../..`

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: In my testing, I found that that the versions of solcjs installed at `monorepo/node_modules` was older than the version of solcjs installed at `monorepo/contracts/node_modules`, hence this change was needed. Didn't dig into why there are 2 versions installed.

packages/node/package.json
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: actually, the machine unit tests are run under `no-ganache` now, which I think is the right thing to do

packages/node/truffle-config.js
snario: It's because of Truffle or ethereum-waffle's dependencies not being fixed I think. We ought to use `moduleResolutions` or perhaps add `solc` to the root `package.json`

packages/node/package.json
snario: ok

packages/node/truffle-config.js
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: seems to be fixed after rebase

cf19drofxots:
Looks like in the process of moving the machine files to the node package, the functionality of `yarn test <regex>` was broken.

For eg, running `yarn test test/integration/secure-withdraw.spec.ts` runs all the test suites across the Node and the machine.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
yeah @cf19drofxots sorry about that - `yarn test` now invokes subcommands since it runs tests in different ways. not sure how to fix this. one of the subcommands can't take a regex filter anyway.

Issue #1222, opened 2019-04-03
Transfers are forced to use the same token for all beneficiaries


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
The current `Transfer` struct doesn't support different tokens to different beneficiaries.

snario:
This effects #1157

AlexXiong97:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll but should be a relatively straightforward fix right?
https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/libs/Transfer.sol#L27
change from `uint8` to `uint8[]` and modify other files dependent/affected on this.

I could help with this one, any other better alternatives in mind?

Issue #1157, opened 2019-03-28
[node, cf.js] Create better abstractions for arbitrary state deposits


snario:
Presently in CF.js we use `AppInstanceInfo` which gets passed into the `Node` which translates it into a `ProposedAppInstanceInfo`. There are fields in this object, namely `myDeposit` and `peerDeposit` which are too strict for ETH-based 2-person deposits. We need to try to simplify this for the scenarios of n-party arbitrary assets. Ideally these nicely end up inside `Transfer.Transaction` object.

This is a **spike** so the outcome of this issue should be some well thought out action items.

PR #1263, opened 2019-04-11
[monorepo] replace Transfer.sol mechanism with Interpreters

### LINKED ISSUE #1258: [contracts] Upgrade StateChannelTransaction to accept an Interpreter argument
# todo
- [x] change "resolution" to be of type `bytes` instead of `Transfer.Transaction`
- [x] add an interpreter field to `executeAppConditionalTransaction`
- [x] write interpreters for existing apps
- [ ] virtual channels

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR refactors the contracts and node to fix the following issues: #1222, #1157, #1442

The design is as follows. App Definition resolve functions now return unstructured data which is to be consumed by an "interpreter" contract which runs in the multisig context and handles state deposit payouts in dispute scenarios. App Definitions are classified by "resolve type", e.g.,

- `TwoPartyOutcome`: the app is some kind of 2-party game, and the resolve function returns a `uint256` in the set `{0, 1, 2}` which means "player 1 won, player 2 won, draw" respectively
- `ETHTransfer`: the resolve functions returns a list of `(to, value)` tuples

State deposits constrain what apps they can be used with: for instance, a deposit of ETH can be used for apps of resolve type `ETHTransfer` or `TwoPartyOutcome`, but a deposit of two ERC721 NFTs can only be used for `TwoPartyOutcome`.

## hard-coded assumptions
- virtual apps always resolve to type 2-party outcome

## temporary hacks to be removed in the future
- A `resolveType` function is introduced in all app definitions. This is used by the uninstall protocol to determine how to update the free balance during uninstall.
- A `limitOrTotal` field is added to the AppInstance class. For app instances interpreted by `ETHInterpreter`, this represents a limit on the total ETH; for those interpreted by `TwoPartyEthAsLump`, this represents the total funds committed to the app.
- The ttt bot assumes that it is player 2
- `interpreterHash` is computed but not checked in `StateChannelTransaction`


packages/apps/contracts/CommitRevealApp.sol
```diff
@@ -115,23 +115,13 @@ contract CommitRevealApp is CounterfactualApp {
     address[] memory to = new address[](1);
     uint256 player;
     if (state.stage == Stage.DONE) {
-      player = uint256(state.winner);
+      return abi.encodePacked(player);
```
snario: Can you import the enum symbol

packages/apps/waffle.js
```diff
@@ -11,7 +11,7 @@ var waffleConfig = {
 };
 
 var selectSolc = () => {
-  if (process.env.CI || process.env.NATIVE_SOLC == "true") {
+  if (1) {
```
snario: hm

packages/contracts/contracts/libs/Transfer.sol
```diff
@@ -31,6 +31,12 @@ library Transfer {
     bytes[] data;
   }
 
+  struct Transaction2 {
```
snario: Remove `2`

packages/contracts/contracts/libs/TwoPartyLumpAsEth.sol
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  struct Params {
+    address payable[2] playerAddrs;
+    uint256 limit;
+  }
+
+  function interpret(
+    bytes memory resolution, bytes memory params
+  ) public {
```
snario: Unnecessary multi-line

packages/contracts/contracts/test-fixtures/ResolveToPay5WeiApp.sol
```diff
@@ -11,26 +11,16 @@ import "../CounterfactualApp.sol";
...
-    bytes[] memory data;
-
-    return Transfer.Transaction(
-      0,
-      address(0),
-      to,
-      amounts,
-      data
-    );
+    return abi.encode(500);
```
snario: Why sometimes packed and sometimes not?

packages/contracts/waffle.js
```diff
@@ -11,10 +11,9 @@ var waffleConfig = {
 };
 
 var selectSolc = () => {
-  if (process.env.CI || process.env.NATIVE_SOLC == "true") {
+  if (1) {
```
snario: hm

packages/contracts/contracts/Interpreter.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity 0.5.7;
+pragma experimental "ABIEncoderV2";
+
+import "./libs/Transfer.sol";
+
+
+contract Interpreter {
+
+  function interpret(bytes memory, bytes memory)
```
AlexXiong97: can we add commentary notes on what this function should achieve?

packages/contracts/waffle.js
AlexXiong97: shouldn't change this, because Netlify doesn't have `solc` binary natively installed, which will fail all checks.

packages/contracts/contracts/StateChannelTransaction.sol
```diff
@@ -48,16 +47,13 @@ contract StateChannelTransaction {
...
       appIdentityHash
     );
 
-    require(
-      Transfer.meetsTerms(txn, terms),
-      "Transfer details do not meet terms"
-    );
-
-    txn.execute();
+    bytes memory payload = abi.encodeWithSignature("interpret(bytes,bytes)", resolution, interpreterParams);
```
AlexXiong97: what should be included in interpreterParams? and in any specific order?

AlexXiong97:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll two extra comments:
1. better comments would help. (e.g. I don't really know what am I supposed to feed into `resolution` and `interpreterParams`
2. in the future, I would be more friendly if we provide a predefined set of interpreter options (i.e. `preset`) for most use case such that they can just choose one (from an `enum`, which contains a list of common interpreters) by specifying a number, or they can customize their own for more complicated app.

in general, I do think current change is great and will support coinshuffle app requirement. Great work!

packages/apps/contracts/CommitRevealApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: what would this change to after the import?

snario: The code needs to be as readable as possible so ideally you would be able to declare `player` variable using the enum.

packages/contracts/contracts/StateChannelTransaction.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: depends on the interpreter

packages/contracts/waffle.js
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this is just for me, since my .env file has never worked...

packages/apps/contracts/CommitRevealApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: ah

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: hmm isn't it a bit strange to have players be represented by an enum with value `SEND_TO_ADDR_ONE`?

snario: Perhaps the name is bad; should try to find a better name

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
might be blocked on https://github.com/duaraghav8/Ethlint/issues/261

packages/contracts/contracts/UintAsEthVirtualApp.sol
```diff
@@ -0,0 +1,50 @@
+pragma solidity 0.5.7;
+pragma experimental "ABIEncoderV2";
+
+import "./AppRegistry.sol";
+import "./NonceRegistry.sol";
+
+
+contract UintAsEthVirtualApp {
```
snario: Should be "AsWei"?

packages/contracts/contracts/interfaces/CounterfactualApp.sol
```diff
@@ -21,9 +18,14 @@ interface CounterfactualApp {
     pure
     returns (bytes memory);
 
-  function resolve(bytes calldata, Transfer.Terms calldata)
+  function resolve(bytes calldata)
+    external
+    pure
+    returns (bytes memory);
+
+  function resolveSelector()
```
snario: We can avoid this, right? By enforcing return types and args be `bytes`?

packages/contracts/contracts/mixins/MAppCaller.sol
```diff
@@ -64,20 +62,20 @@ contract MAppCaller {
...
   function resolve(
     address appDefinitionAddress,
-    bytes memory appState,
-    bytes memory terms
+    bytes memory appState
   )
     internal
-    pure
-    returns (Transfer.Transaction memory)
+    view
```
snario: Good catch, maybe its own PR?

packages/apps/contracts/PreimageApp.sol
```diff
@@ -0,0 +1,70 @@
...
+
+  function resolve(bytes memory encodedState)
+    public
+    pure
+    returns (bytes memory)
+  {
+    AppState memory state = abi.decode(encodedState, (AppState));
+
+    if (state.preimageRevealed) {
+      return abi.encode(1);
```
snario: Replace with enum?

packages/contracts/contracts/StateChannelTransaction.sol
```diff
@@ -1,10 +1,9 @@
 pragma solidity 0.5.8;
 pragma experimental "ABIEncoderV2";
 
-import "./libs/Transfer.sol";
-
 import "./NonceRegistry.sol";
 import "./AppRegistry.sol";
+import "./interfaces/Interpreter.sol";
```
snario: You import this but it is not used, why?

packages/contracts/contracts/StateChannelTransaction.sol
```diff
@@ -48,16 +45,15 @@ contract StateChannelTransaction {
...
-    require(
-      Transfer.meetsTerms(txn, terms),
-      "Transfer details do not meet terms"
-    );
-
-    txn.execute();
+    bytes memory payload = abi.encodeWithSignature(
+      "interpret(bytes,bytes)", resolution, interpreterParams);
+    (bool success, bytes memory returnData) =
+      interpreterAddress.delegatecall(payload);
```
snario: cc @chriseth, kind of an interesting use case.

packages/contracts/contracts/default-apps/ETHBalanceRefundApp.sol
```diff
@@ -13,27 +12,28 @@ contract ETHBalanceRefundApp {
     uint256 threshold;
   }
 
-  function resolve(bytes memory encodedState, Transfer.Terms memory terms)
-    public
+  function resolve(bytes calldata encodedState)
+    external
```
snario: Again with `external` and `calldata`. Can you make a definitive call on this and add it to the specs, please?

packages/contracts/contracts/default-apps/ETHBalanceRefundApp.sol
```diff
@@ -13,27 +12,28 @@ contract ETHBalanceRefundApp {
...
   {
-    AppState memory state = abi.decode(encodedState, (AppState));
+    AppState memory appState = abi.decode(encodedState, (AppState));
 
-    uint256[] memory amounts = new uint256[](1);
-    amounts[0] = address(state.multisig).balance - state.threshold;
+    ETHInterpreter.ETHTransfer[] memory ret =
+      new ETHInterpreter.ETHTransfer[](1);
+    ret[0].amount = address(appState.multisig).balance - appState.threshold;
+    ret[0].to = appState.recipient;
```
snario: Is this `ETHInterpreter.ETHTransfer[]` object instantiable inline?

packages/contracts/contracts/default-apps/ETHBalanceRefundApp.sol
```diff
@@ -13,27 +12,28 @@ contract ETHBalanceRefundApp {
...
+  }
 
-    return Transfer.Transaction(
-      terms.assetType,
-      terms.token,
-      to,
-      amounts,
-      data
-    );
+  function resolveType()
```
snario: Is this needed?

packages/contracts/contracts/default-apps/ETHBucket.sol
```diff
@@ -1,66 +1,31 @@
...
 pragma experimental "ABIEncoderV2";
 
-import "../libs/Transfer.sol";
-import "../CounterfactualApp.sol";
+import "../interfaces/CounterfactualApp.sol";
+import "../interfaces/Interpreter.sol";
+import "../interpreters/ETHInterpreter.sol";
 
 
-contract ETHBucket is CounterfactualApp {
```
snario: Inheritance?

packages/contracts/contracts/default-apps/ETHBucket.sol
```diff
@@ -1,66 +1,31 @@
...
+contract ETHBucket {
 
   struct AppState {
-    address alice;
-    address bob;
-    uint256 aliceBalance;
-    uint256 bobBalance;
+    ETHInterpreter.ETHTransfer[] transfers;
```
snario: Seems clever but makes the data structure unnecessarily nested I think? To be specific, this is now `tuple(tuple(address[], uint256[])[]))` instead of `tuple(address[], uint256[])` right?

packages/contracts/contracts/default-apps/ETHBucket.sol
```diff
@@ -1,66 +1,31 @@
...
-    returns (address)
+    returns (bytes memory)
   {
-    revert("Not implemented");
+    return encodedState;
   }
 
-  function applyAction(bytes memory, bytes memory)
-    public
+  function resolveType()
```
snario: What are the tradeoffs we incur by forcing the interpreter to a contract-level decision versus a software level decision?

packages/contracts/contracts/interfaces/CounterfactualApp.sol
```diff
@@ -21,9 +18,15 @@ interface CounterfactualApp {
...
     returns (bytes memory);
 
-  function resolve(bytes calldata, Transfer.Terms calldata)
+  function resolve(bytes calldata)
+    external
+    pure
+    returns (bytes memory);
+
+  // a hack
+  function resolveType()
```
snario: As I have mentioned above, this parameter is confusing me a bit. Could you explain it in more detail inside the comments of this pull request please?

packages/contracts/contracts/interfaces/Interpreter.sol
```diff
@@ -0,0 +1,15 @@
...
+contract Interpreter {
+
+  enum ResolutionType {
+    TWO_PARTY_OUTCOME,
+    ETH_TRANSFER,
+    UINT256
+  }
+
+  function interpret(bytes memory, bytes memory)
+    public;
```
snario: Unnecessary indentation

packages/contracts/contracts/interfaces/Interpreter.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity 0.5.8;
+pragma experimental "ABIEncoderV2";
+
+
+contract Interpreter {
```
snario: Likewise, can you write a description for this object (and in general the approach you took that led to this) inside of the pull-request? Will also be useful for @AlexXiong97 who is reviewing this PR, too.

packages/contracts/contracts/interpreters/ETHInterpreter.sol
```diff
@@ -0,0 +1,39 @@
+pragma solidity 0.5.8;
+pragma experimental "ABIEncoderV2";
+
+
+contract ETHInterpreter {
+
+  struct ETHTransferWrapper {
+    ETHTransfer[] wrapped;
```
snario: Why this versus `address[]` and `uint256[]`?

packages/contracts/contracts/interpreters/ETHInterpreter.sol
```diff
@@ -0,0 +1,39 @@
...
+    uint256 amount;
+  }
+
+  struct Param {
+    uint256 limit;
+  }
+
+  function interpret(
+    bytes memory input, bytes memory params
+  ) public {
```
snario: Unnecessary indentation?

packages/contracts/contracts/interpreters/ETHInterpreter.sol
```diff
@@ -0,0 +1,39 @@
...
+    uint256 limitRemaining = abi.decode(params, (Param)).limit;
+
+    for (uint i=0; i<transfers.length; i++) {
+      address payable to = address(uint160(transfers[i].to));
+      uint256 amount = transfers[i].amount;
+
+      require(amount <= limitRemaining, "hit the limit");
+      limitRemaining -= amount;
+
+      // note: send() is deliberately used instead of transfer() here
```
snario: Instead of saying it is deliberately used, explain _why_ it is deliberately used.

packages/contracts/contracts/interpreters/TwoPartyEthAsLump.sol
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  function interpret(
+    bytes memory encodedResolution, bytes memory encodedParams
+  ) public {
+
+    Params memory params = abi.decode(encodedParams, (Params));
+    Resolution resolution = abi.decode(encodedResolution, (Resolution));
+
+    if (resolution == Resolution.SEND_TO_ADDR_ONE) {
```
snario: I see you used enums here but integer literals in the above place where this code is used that I commented on. Stick with enums?

packages/contracts/contracts/mixins/MAppCaller.sol
snario: bump

packages/contracts/contracts/test-fixtures/AppWithAction.sol
```diff
@@ -28,24 +30,12 @@ contract AppWithAction is CounterfactualApp{
...
-    bytes[] memory data = new bytes[](2);
-
-    return Transfer.Transaction(
-      terms.assetType,
-      terms.token,
-      to,
-      amounts,
-      data
-    );
+    return abi.encode(0);
```
snario: Still, recommend using enums always here.

packages/contracts/contracts/test-fixtures/ResolveTo0App.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity 0.5.8;
+pragma experimental "ABIEncoderV2";
+
+import "../interfaces/Interpreter.sol";
+
+
+contract ResolveTo0App {
```
snario: I recommend renaming to `ResolveToZeroApp` just to make it more readable

packages/contracts/contracts/test-fixtures/ResolveTo500App.sol
```diff
@@ -0,0 +1,28 @@
+pragma solidity 0.5.8;
+pragma experimental "ABIEncoderV2";
+
+import "../interfaces/Interpreter.sol";
+
+// todo: adding is CounterfactualApp makes empty bytecode...
```
snario: Ah, now I see this. Can you explain this in the PR?

packages/contracts/contracts/test-fixtures/ResolveTo500App.sol
```diff
@@ -0,0 +1,28 @@
...
+// todo: rename to resolveToFirstPlayer
```
snario: Marking this TODO.

packages/contracts/contracts/test-fixtures/ResolveToPay5WeiApp.sol
```diff
@@ -1,59 +1,28 @@
 pragma solidity 0.5.8;
 pragma experimental "ABIEncoderV2";
 
-import "../libs/Transfer.sol";
-import "../CounterfactualApp.sol";
+import "../interfaces/Interpreter.sol";
 
+// todo: adding is CounterfactualApp makes empty bytecode...
+// todo: rename to resolveToFirstPlayer
```
snario: Duplicate code.

packages/contracts/contracts/test-fixtures/ResolveToPay5WeiApp.sol
snario: Is this still relevant @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll ?

packages/contracts/test/eth-virtual-app-agreement.spec.ts
```diff
@@ -172,9 +141,11 @@ describe("ETHVirtualAppAgreement", () => {
       appIdentityHash,
       0,
       bigNumberify(10),
-      0,
       HashZero
     );
+
+    // actually, 500 is treated as "split"
```
snario: This comment lacks context, why is it here?

packages/contracts/waffle.js
```diff
@@ -11,11 +11,11 @@ var waffleConfig = {
 };
 
 var selectSolc = () => {
-  if (process.env.CI || process.env.NATIVE_SOLC == "true") {
+  if (process.env.NATIVE_SOLC == "true") {
```
snario: Why did you remove the `CI` env variable?

packages/dapp-high-roller/src/components/app-provider/app-provider.tsx
```diff
@@ -54,6 +54,8 @@ export class AppProvider {
   nodeProvider: MockNodeProvider = {} as MockNodeProvider;
 
   @Prop({ mutable: true }) cfProvider: cf.Provider = {} as cf.Provider;
+
+  // A factory for HighRoller
```
snario: Seems unnecessary?

packages/dapp-high-roller/src/components/app-provider/app-provider.tsx
```diff
@@ -78,10 +80,11 @@ export class AppProvider {
     this.cfProvider.on("uninstall", this.onUninstall.bind(this));
     this.cfProvider.on("installVirtual", this.onInstall.bind(this));
 
-    const appId = "0x91907355C59BA005843E791c88aAB80b779446c9";
+    const highRollerAppDefinitionAddr =
+      "0x91907355C59BA005843E791c88aAB80b779446c9";
```
snario: I know this was here before but, what is this magic number?

packages/node/jest.config.js
```diff
@@ -33,6 +32,6 @@ module.exports = {
   "transform": {
     "^.+\\.tsx?$": "ts-jest"
   },
-  "verbose": true,
+  "verbose": false,
```
snario: We keep moving this around. Can you make it its own PR?

packages/node/package.json
```diff
@@ -11,9 +11,9 @@
     "test": "yarn run test:node && yarn run test:machine",
     "test:machine": "yarn run test:machine:integration && yarn run test:machine:unit",
     "test:machine:integration": "./scripts/test.sh",
-    "test:machine:unit": "jest --testPathPattern=test/machine/unit --bail --forceExit",
-    "test:node": "jest --testPathIgnorePatterns=test/machine --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
-    "test:coverage": "jest --runInBand --detectOpenHandles --bail --coverage",
```
snario: Can you make this its own PR?

packages/node/scripts/test.sh
```diff
@@ -63,8 +63,9 @@ yarn run truffle migrate --network machine --reset > /dev/null
 echo "🎬 Starting jest test suites"
 
 jest \
-  --testPathPattern=test/machine/integration/ \
+  --testPathPattern=test/machine/integration/protocols.spec.ts \
```
snario: Why? What happened here? Is this just for you testing?

packages/node/scripts/test.sh
```diff
@@ -63,8 +63,9 @@ yarn run truffle migrate --network machine --reset > /dev/null
...
   --runInBand `#integration tests fail parallelized (tx nonce out of sync)` \
   --detectOpenHandles \
   --forceExit \
+  -- verbose \
```
snario: Also you added this here but marked it as false in the config 😆

packages/node/src/ethereum/install-commitment.ts
```diff
@@ -15,21 +16,20 @@ export class InstallCommitment extends MultiSendCommitment {
...
     public readonly freeBalanceAppIdentity: AppIdentity,
-    public readonly freeBalanceTerms: Terms,
     public readonly freeBalanceStateHash: string,
     public readonly freeBalanceNonce: number,
     public readonly freeBalanceTimeout: number,
     public readonly dependencyNonce: number,
-    public readonly rootNonceValue: number
+    public readonly rootNonceValue: number,
+    public readonly interpreterAddr?: string,
+    public readonly interpreterParams?: string
```
snario: Shouldn't both of these be required? Why are they optional?

packages/node/src/ethereum/install-commitment.ts
```diff
@@ -68,12 +68,13 @@ export class InstallCommitment extends MultiSendCommitment {
...
-        this.networkContext.AppRegistry,
-        this.networkContext.NonceRegistry,
-        uninstallKey,
-        this.rootNonceValue,
-        appIdentityHash,
-        this.terms
+        /* appRegistry */ this.networkContext.AppRegistry,
+        /* nonceRegistry */ this.networkContext.NonceRegistry,
+        /* uninstallKey */ uninstallKey,
+        /* rootNonceExpectedValue */ this.rootNonceValue,
```
snario: If these are different, perhaps rename the JS variable to the expected name of the Solidity variable?

packages/node/src/ethereum/setup-commitment.ts
```diff
@@ -60,7 +61,7 @@ export class SetupCommitment extends MultisigCommitment {
 
   private getSaltForDependencyNonce() {
     return keccak256(
-      defaultAbiCoder.encode(["uint256"], [DependencyValue.NOT_UNINSTALLED])
+      defaultAbiCoder.encode(["uint256"], [DependencyValue.NOT_CANCELLED])
```
snario: What does `CANCELLED` mean exactly? I feel like `UNINSTALLED` has a specific meaning.

Maybe `NOT_INVALIDATED` is more precise?

packages/node/src/ethereum/uninstall-commitment.ts
```diff
@@ -44,11 +41,11 @@ export class UninstallCommitment extends MultiSendCommitment {
...
-        0, // Timeout is 0 for dependencyNonce!
-        keccak256(defaultAbiCoder.encode(["uint256"], [this.dependencyNonce])),
-        // Hard coded the update to 1 because that is the value
-        // that represents an app as being "uninstalled"
-        DependencyValue.UNINSTALLED
+        /* timeout */ 0,
+        /* salt */ keccak256(
+          defaultAbiCoder.encode(["uint256"], [this.dependencyNonce])
+        ),
+        /* nonceValue */ DependencyValue.CANCELLED
```
snario: Hey @ricmoo is there any way `encode` could take an object with named arguments so we didn't have to do this weird commenting pattern as you see above?

packages/node/src/ethereum/utils/eth-bucket.ts
```diff
@@ -23,16 +17,45 @@ export function getETHBucketAppInterface(addr: string): AppInterface {
...
+      ret[key] = a[key];
+    }
+  }
+  for (const key of Object.keys(b)) {
+    if (!a[key]) {
+      throw Error(`mismatch ${Object.keys(a)} ${Object.keys(b)}`);
+    }
+  }
+  return ret;
+}
```
snario: What is any of this stuff above?

snario: I think it makes sense but maybe name them more explicitly? Since they are not methods on a class I would advise against super short names like `flip` and `merge`.

packages/node/src/machine/types.ts
```diff
@@ -106,7 +104,7 @@ export type UninstallVirtualAppParams = {
   respondingXpub: string;
   intermediaryXpub: string;
   targetAppIdentityHash: string;
-  targetAppState: SolidityABIEncoderV2Type;
+  targetAppState: SolidityABIEncoderV2Type; // todo: remove
```
snario: Marking this TODO.

packages/node/src/methods/app-instance/get-free-balance/controller.ts
```diff
@@ -25,16 +26,15 @@ export default class GetFreeBalanceController extends NodeController {
...
       .state as ETHBucketAppState;
-    return {
-      [alice]: aliceBalance,
-      [bob]: bobBalance
-    };
+
+    const ret: Node.GetFreeBalanceStateResult = {};
+
+    for (const { amount, to } of appState) {
+      ret[to] = bigNumberify(amount._hex);
```
snario: Do you need to specify `._hex`? I think the function works correctly without doing that.

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -56,6 +56,11 @@ export default class CreateChannelController extends NodeController {
       MinimumViableMultisig.abi
     ).functions.setup.encode([multisigOwners]);
 
+    let bytecode = Proxy.bytecode;
+    if (bytecode.slice(0, 2) !== "0x") {
+      bytecode = `0x${bytecode}`;
+    }
```
snario: This seems like a type failure. Why would it sometimes include `0x` and sometimes not?

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -66,7 +71,7 @@ export default class CreateChannelController extends NodeController {
           keccak256(
             solidityPack(
               ["bytes", "uint256"],
-              [`0x${Proxy.bytecode}`, networkContext.MinimumViableMultisig]
+              [`${bytecode}`, networkContext.MinimumViableMultisig]
```
snario:
```suggestion
              [bytecode, networkContext.MinimumViableMultisig]
```

packages/node/src/models/app-instance.ts
```diff
@@ -78,27 +77,30 @@ export class AppInstance {
...
-    terms: Terms,
     isVirtualApp: boolean,
     appSeqNo: number,
     rootNonceValue: number,
     latestState: any,
     latestNonce: number,
-    latestTimeout: number
+    latestTimeout: number,
+    beneficiaries: [string, string],
+    limitOrTotal: BigNumber
```
snario: Add docstring comments to this class for both `beneficiaries` and `limitOrTotal` please.

packages/node/src/models/app-instance.ts
```diff
@@ -212,6 +205,14 @@ export class AppInstance {
     return this.json.latestNonce;
   }
 
+  public get limitOrTotal() {
+    return bigNumberify(this.json.limitOrTotal._hex);
```
snario: Same comment, do you need to specify `._hex`?

packages/node/src/models/proposed-app-instance-info.ts
```diff
@@ -128,13 +121,15 @@ export class ProposedAppInstanceInfo implements AppInstanceInfo {
...
       proposedAppInterface,
-      proposedTerms,
       isVirtualApp,
       isVirtualApp ? 1337 : stateChannel.numInstalledApps,
       stateChannel.rootNonceValue,
       this.initialState,
       0,
-      bigNumberify(this.timeout).toNumber()
+      bigNumberify(this.timeout).toNumber(),
+      // not used in app identity computation
```
snario: Is this comment trying to say "the below 2 parameters are not used in app identity computation"? It is unclear.

packages/node/src/protocol/uninstall-virtual-app.ts
```diff
@@ -21,6 +22,10 @@ import { getChannelFromCounterparty } from "./utils/get-channel-from-counterpart
 import { computeFreeBalanceIncrements } from "./utils/get-resolution-increments";
 import { validateSignature } from "./utils/signature-validator";
 
+const zA = (xpub: string) => {
```
snario: `zA`?

packages/node/src/protocol/utils/get-resolution-increments.ts
```diff
@@ -1,80 +1,74 @@
...
-    bigNumberify(appInstance.terms.limit).gt(Zero) &&
-    resolution.value.every(v => v.eq(Zero)) &&
-    attempts < 10
-  ) {
-    log.info(`Empty resolution. Querying chain again. Attempt #${attempts}`);
+  /// todo(xuanji)
+  /// there used to be a bunch of retry logic here, to handle the case where
+  /// the appDefinition contract was deployed recently (and hence the provider
+  /// sometimes returning an incorrect definition). This TODO is to handle
+  /// that situation.
```
snario: > the appDefinition contract was deployed recently 

No, this is not why the retry logic was here.

The retry logic is to handle the fact that one provider might return one thing (because it observed a balance change that mattered in the `resolve view` function of the ETH balance refund app from the POV of Party A but then Party B might be using a provider where that state has not been observed to have changed yet (it only happened like 1 or 2 blocks ago for example and the chain did not sync yet).

We need to resolve this TODO before merging or the Playground will break.

packages/node/src/protocol/utils/get-resolution-increments.ts
```diff
@@ -1,80 +1,74 @@
...
+  if (resolveType.eq(One)) {
+    /* ETHInterpreter */
+
+    const decoded = defaultAbiCoder.decode(
+      ["tuple(address,uint256)[]"],
+      resolution
     );
 
-    attempts += 1;
+    // todo(xuanji): check shape
```
snario: Marking this TODO.

packages/node/src/protocol/utils/get-resolution-increments.ts
```diff
@@ -1,80 +1,74 @@
...
+    const [[[address, to]]] = decoded;
 
-    await wait(1000 * attempts);
+    return { [address]: to };
   }
-  // END FIXME
+  if (resolveType.eq(Zero)) {
+    /* TWO_PARTY_OUTCOME */
```
snario: Use enums?

packages/node/src/protocol/utils/get-resolution-increments.ts
```diff
@@ -1,80 +1,74 @@
...
-    bob: channel.multisigOwners[1]
-  };
+    const i0 = total.div(2);
+    const i1 = total.sub(i0);
+    return {
+      [appInstance.beneficiaries[0]]: i0,
+      [appInstance.beneficiaries[1]]: i1
+    };
+  }
+  throw Error("unknown interpreter");
```
snario: The above logic would be nice with a switch statement.

packages/node/test/machine/unit/ethereum/set-state-commitment.spec.ts
```diff
@@ -68,15 +68,15 @@ describe("Set State Commitment", () => {
         owner,
         signingKeys,
         appDefinitionAddress,
-        termsHash,
+        {} /* interpreter */,
+        {} /* interpreter */,
```
snario: `{}`? Not `AddressZero` and `HashZero`?

snario: Seems like a type error.

packages/node/test/machine/unit/ethereum/virtual-app-set-state-commitment.spec.ts
```diff
@@ -29,6 +28,7 @@ describe("Virtual App Set State Commitment", () => {
 
   const networkContext = generateRandomNetworkContext();
 
+  // todo(xuanji): use createAppInstance
```
snario: Marking this TODO.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "noImplicitAny": false,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
snario: Its own PR, please

packages/contracts/contracts/default-apps/ETHBucket.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It should be `tuple(address, uint256)[]`

packages/contracts/contracts/UintAsEthVirtualApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why?

packages/contracts/waffle.js
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: IIRC because the selector in master never does native solc, even on a dev's local laptop. Will fix that separately

packages/apps/contracts/CommitRevealApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I will close this thread unless a better name is proposed for consideration

packages/contracts/contracts/test-fixtures/ResolveToPay5WeiApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: deleted

packages/contracts/contracts/StateChannelTransaction.sol
chriseth: Not sure what you mean - what should it represent?

packages/apps/contracts/HighRollerApp.sol
```diff
@@ -10,7 +11,7 @@ import "@counterfactual/contracts/contracts/CounterfactualApp.sol";
 ///         The winner is the player whose sum of dice outcomes is highest.
 /// @dev This contract is an example of a dApp built to run on
 ///      the CounterFactual framework
-contract HighRollerApp is CounterfactualApp {
```
snario: Why do you remove this inheritance?

packages/apps/contracts/HighRollerApp.sol
```diff
@@ -33,7 +34,6 @@ contract HighRollerApp is CounterfactualApp {
   }
 
   struct AppState {
-    address[2] playerAddrs;
```
snario: This was done in another PR, I think. Can you rebase on `master`?

packages/apps/contracts/HighRollerApp.sol
```diff
@@ -120,59 +120,39 @@ contract HighRollerApp is CounterfactualApp {
     return abi.encode(nextState);
   }
 
-  function resolve(bytes calldata encodedState, Transfer.Terms calldata terms)
+  function resolve(bytes calldata encodedState)
     external
     pure
-    returns (Transfer.Transaction memory)
+    returns (TwoPartyEthAsLump.Resolution)
```
snario: Shouldn't this include `memory`?

packages/apps/contracts/HighRollerApp.sol
```diff
@@ -120,59 +120,39 @@ contract HighRollerApp is CounterfactualApp {
...
     if (expectedCommitHash == appState.commitHash) {
-      amounts = getWinningAmounts(
-        appState.playerFirstNumber, appState.playerSecondNumber, terms.limit
+      return getWinningAmounts(
+        appState.playerFirstNumber, appState.playerSecondNumber
       );
     } else {
-      amounts[0] = 0;
-      amounts[1] = terms.limit;
+      return TwoPartyEthAsLump.Resolution.SEND_TO_ADDR_TWO;
```
snario: Why did you change the `else` to send it to address two? It's meant to split it, right?

packages/apps/contracts/HighRollerApp.sol
```diff
@@ -120,59 +120,39 @@ contract HighRollerApp is CounterfactualApp {
...
-      data
-    );
   }
 
-  function getWinningAmounts(uint256 num1, uint256 num2, uint256 termsLimit)
+  function getWinningAmounts(uint256 num1, uint256 num2)
     public
     pure
-    returns (uint256[] memory)
+    returns (TwoPartyEthAsLump.Resolution)
```
snario: `memory`?

packages/apps/contracts/HighRollerApp.sol
```diff
@@ -120,59 +120,39 @@ contract HighRollerApp is CounterfactualApp {
...
-      amounts[1] = 0;
+      return TwoPartyEthAsLump.Resolution.SEND_TO_ADDR_ONE;
     } else if (playerFirstTotal < playerSecondTotal) {
-      amounts[0] = 0;
-      amounts[1] = termsLimit;
+      return TwoPartyEthAsLump.Resolution.SEND_TO_ADDR_TWO;
     } else {
-      amounts[0] = termsLimit / 2;
-      amounts[1] = termsLimit / 2;
+      return TwoPartyEthAsLump.Resolution.SPLIT_AND_SEND_TO_BOTH_ADDRS;
```
snario: Ah, ignore the earlier message, I see now.

packages/apps/contracts/NimApp.sol
```diff
@@ -1,22 +1,22 @@
...
+import "@counterfactual/contracts/contracts/interfaces/CounterfactualApp.sol";
+import "@counterfactual/contracts/contracts/interpreters/TwoPartyEthAsLump.sol";
 
 
 /*
 Normal-form Nim
 https://en.wikipedia.org/wiki/Nim
 */
-contract NimApp is CounterfactualApp {
+contract NimApp {
```
snario: Inheritance?

packages/apps/contracts/NimApp.sol
```diff
@@ -64,35 +64,20 @@ contract NimApp is CounterfactualApp {
...
   }
 
-  function resolve(
-    bytes calldata encodedState, Transfer.Terms calldata terms
-  )
+  function resolve(bytes calldata encodedState)
     external
     pure
-    returns (Transfer.Transaction memory)
+    returns (TwoPartyEthAsLump.Resolution)
```
snario: `memory`?

packages/apps/contracts/PreimageApp.sol
```diff
@@ -0,0 +1,70 @@
...
+    public
+    pure
+    returns (bool)
+  {
+    AppState memory state = abi.decode(encodedState, (AppState));
+    return state.preimageRevealed;
+  }
+
+  function getTurnTaker(
+    bytes memory /* encodedState */, address[] memory signingKeys
```
snario: Does omitting the variable instantiation do anything?

packages/apps/contracts/PreimageApp.sol
```diff
@@ -0,0 +1,70 @@
...
+  )
+    public
+    pure
+    returns (address)
+  {
+    return signingKeys[0];
+  }
+
+  function applyAction(bytes memory encodedState, bytes memory encodedAction)
```
snario: Why do you use `memory` sometimes and `calldata` other times? When we call these methods from the `Node` it will be `calldata` but when we call from `AppRegistry` it will be `memory`... I think? Or is it always `calldata`? In any case, this is the kind of standardization we should be getting from inheritance, no?

packages/apps/contracts/PreimageApp.sol
```diff
@@ -0,0 +1,70 @@
...
+
+    require(
+      !state.preimageRevealed,
+      "No actions possible after preimage revealed");
+
+    AppState memory nextState = state;
+
+    if (keccak256(action.preimage) == state.image) {
+      nextState.preimageRevealed = true;
+    }
```
snario:
```suggestion
    }
    
```

packages/apps/contracts/PreimageApp.sol
```diff
@@ -0,0 +1,70 @@
...
+    returns (bytes memory)
+  {
+    AppState memory state = abi.decode(encodedState, (AppState));
+    Action memory action = abi.decode(encodedAction, (Action));
+
+    require(
+      !state.preimageRevealed,
+      "No actions possible after preimage revealed");
+
+    AppState memory nextState = state;
```
snario: What is the purpose of instantiating a new variable versus re-using `state`?

packages/apps/contracts/PreimageApp.sol
```diff
@@ -0,0 +1,70 @@
...
+    public
+    pure
+    returns (bytes memory)
+  {
+    AppState memory state = abi.decode(encodedState, (AppState));
+
+    if (state.preimageRevealed) {
+      return abi.encode(1);
+    } else {
+      return abi.encode(0);
```
snario: Replace with enum?

packages/apps/contracts/TicTacToeApp.sol
```diff
@@ -1,10 +1,12 @@
 pragma solidity 0.5.8;
 pragma experimental "ABIEncoderV2";
 
-import "@counterfactual/contracts/contracts/CounterfactualApp.sol";
+import "@counterfactual/contracts/contracts/interfaces/Interpreter.sol";
+import "@counterfactual/contracts/contracts/interfaces/CounterfactualApp.sol";
+import "@counterfactual/contracts/contracts/interpreters/TwoPartyEthAsLump.sol";
 
 
-contract TicTacToeApp is CounterfactualApp {
```
snario: Inheritance?

packages/apps/contracts/TicTacToeApp.sol
```diff
@@ -223,4 +194,12 @@ contract TicTacToeApp is CounterfactualApp {
     }
   }
 
+  function resolveType()
```
snario: I haven't seen this used in `HighRollerApp.sol`, why not?

packages/apps/test/high-roller-app.spec.ts
```diff
@@ -114,8 +111,8 @@ describe("HighRollerApp", () => {
     );
   }
 
-  async function resolve(state: SolidityABIEncoderV2Type, terms: Terms) {
-    return await highRollerApp.functions.resolve(encodeState(state), terms);
+  async function resolve(state: SolidityABIEncoderV2Type) {
```
snario: Perhaps a better name for this function is `computeResolution`?

packages/apps/test/high-roller-app.spec.ts
```diff
@@ -236,27 +229,16 @@ describe("HighRollerApp", () => {
...
-
-      expect(transaction.assetType).to.eq(AssetType.ETH);
-      expect(transaction.token).to.eq(AddressZero);
-      expect(transaction.to).to.deep.eq([AddressZero, AddressZero]);
-      expect(transaction.value[0]).to.eq(Zero);
-      expect(transaction.value[1]).to.eq(parseEther("2"));
-      expect(transaction.data).to.deep.eq(["0x", "0x"]);
+      const result = await resolve(preState);
+
+      expect(result).to.eq(1);
```
snario: Why `1`? Perhaps name this with an enum for better readability.

packages/apps/test/high-roller-app.spec.ts
```diff
@@ -271,27 +253,16 @@ describe("HighRollerApp", () => {
...
-
-      expect(transaction.assetType).to.eq(AssetType.ETH);
-      expect(transaction.token).to.eq(AddressZero);
-      expect(transaction.to).to.deep.eq([AddressZero, AddressZero]);
-      expect(transaction.value[0]).to.eq(parseEther("1"));
-      expect(transaction.value[1]).to.eq(parseEther("1"));
-      expect(transaction.data).to.deep.eq(["0x", "0x"]);
+      const result = await resolve(preState);
+
+      expect(result).to.eq(2);
```
snario: Same as above.

packages/apps/test/high-roller-app.spec.ts
```diff
@@ -301,27 +272,16 @@ describe("HighRollerApp", () => {
...
-
-      expect(transaction.assetType).to.eq(AssetType.ETH);
-      expect(transaction.token).to.eq(AddressZero);
-      expect(transaction.to).to.deep.equal([AddressZero, AddressZero]);
-      expect(transaction.value[0]).to.eq(parseEther("2"));
-      expect(transaction.value[1]).to.eq(Zero);
-      expect(transaction.data).to.deep.equal(["0x", "0x"]);
+      const result = await resolve(preState);
+
+      expect(result).to.eq(0);
```
snario: Same as above.

packages/apps/test/nim.spec.ts
```diff
@@ -73,6 +72,9 @@ describe("Nim", () => {
   before(async () => {
     const provider = waffle.createMockProvider();
     const wallet = (await waffle.getWallets(provider))[0];
+    if (!NimApp.bytecode) {
+      throw Error("empty bytecode");
+    }
```
snario: Why is this necessary? The build-test flow is meant to handle this. An error of this sort should be thrown at the very beginning of the entire test process I think. Perhaps in the jest entrypoint file.

packages/contracts/contracts/ETHVirtualAppAgreement.sol
```diff
@@ -1,36 +1,29 @@
 pragma solidity 0.5.8;
 pragma experimental "ABIEncoderV2";
 
-import "./libs/Transfer.sol";
 import "./AppRegistry.sol";
 import "./NonceRegistry.sol";
 
+/// todo(xuanji): rename to TwoPartyVirtualEthAsLump
```
snario: Highlighting this TODO.

packages/contracts/contracts/ETHVirtualAppAgreement.sol
```diff
@@ -1,36 +1,29 @@
...
 /// @notice Commitment target to support "virtual apps", i.e., apps which have
 /// ETH committed to them via intermediary lock-up instead of having ETH directly
 /// committed in a ledger channel.
-/// The `target` contract must return via getResolution the outcome of the app
-/// with the same asset type as `agreement.terms`. The value of the resolution
-/// to the target's first beneficiary (i.e., `resolution.value[0]`) is
-/// treated as the amount assigned to our first beneficiary (i.e.,
-/// `agreement.beneficiaries[0]`). Note that `terms.limit` is explicitly
-/// ignored since limits are enforced by `capitalProvided`.
+/// The `target` contract must return via getResolution the outcome of the app,
```
snario: The name `getResolution` is out-of-date

packages/contracts/contracts/ETHVirtualAppAgreement.sol
```diff
@@ -40,50 +33,31 @@ contract ETHVirtualAppAgreement {
...
-    require(
-      agreement.terms.token == resolution.token,
-      "returned incompatible resolution based on bad token value"
-    );
-
-    require(
-      agreement.capitalProvided > resolution.value[0],
-      "returned incompatible resolution"
-    );
+    uint256 resolutionAsValue = abi.decode(resolution, (uint256));
```
snario: Does `uint256(resolution)` do the same thing as this line?

packages/contracts/contracts/ETHVirtualAppAgreement.sol
```diff
@@ -40,50 +33,31 @@ contract ETHVirtualAppAgreement {
...
-
-    amount[0] = resolution.value[0];
-    amount[1] = agreement.capitalProvided - amount[0];
+    if (resolutionAsValue == 0) {
+      agreement.beneficiaries[0].send(agreement.capitalProvided);
+      return;
+    } else if (resolutionAsValue == 1) {
+      agreement.beneficiaries[1].send(agreement.capitalProvided);
+      return;
+    }
```
snario: Can re-write this as:

```solidity
if (resolutionAsValue <= 1) {
    agreement.beneficiaries[resolutionAsValue].send(agreement.capitalProvided);
    return;
}
```

packages/contracts/contracts/ETHVirtualAppAgreement.sol
```diff
@@ -40,50 +33,31 @@ contract ETHVirtualAppAgreement {
...
-      beneficiaries,
-      amount,
-      data
-    );
+    agreement.beneficiaries[0].transfer(agreement.capitalProvided / 2);
+    agreement.beneficiaries[1].transfer(
+      agreement.capitalProvided - agreement.capitalProvided / 2);
 
-    ret.execute();
+    return;
```
snario: So you are choosing to basically embed the interpreter's execution function inside of this special case contract for virtual apps. Does this hurt our ability to fully generalize virtual apps with `StateChannelTransaction` in the future?

packages/contracts/contracts/StateChannelTransaction.sol
snario: Is there any syntax for something like `Interpreter(...).interpret(...).delegatecall()`?

packages/contracts/contracts/UintAsEthVirtualApp.sol
snario: bump

packages/contracts/contracts/UintAsEthVirtualApp.sol
```diff
@@ -0,0 +1,50 @@
+pragma solidity 0.5.8;
```
snario: Is this file even used?

packages/apps/contracts/HighRollerApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It's not done in master

packages/apps/contracts/PreimageApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It's an unused parameter, so this silences the warning

packages/apps/contracts/PreimageApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: When public functions are called, the params passed in calldata are copied into memory IIRC. We should use `external/calldata` where possible

snario: OK, can you change all instances to use `external` and `calldata` then?

packages/contracts/contracts/default-apps/ETHBucket.sol
snario: That is not what this is right now as far as I can tell. It is `tuple(tuple(address[], uint256[])[]))`.

packages/contracts/contracts/interpreters/ETHInterpreter.sol
```diff
@@ -3,7 +3,10 @@ pragma experimental "ABIEncoderV2";
 
 import "../interfaces/Interpreter.sol";
 
-
+/*
+This file is excluded from ethlint/solium because of this issue:
+https://github.com/duaraghav8/Ethlint/issues/261
+*/
```
snario: Use this style of comments:

```solidity
/**
 * text
 */
```

packages/apps/contracts/TicTacToeApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It's used in the client-side code during uninstall, but never used in solidity

packages/apps/contracts/HighRollerApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this is fixed to be `bytes memory` but in general, "simple" return types are not annotated with `memory`

packages/apps/contracts/PreimageApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: https://github.com/counterfactual/monorepo/issues/1449

packages/node/src/methods/app-instance/get-free-balance/controller.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: it does not:

image

packages/contracts/contracts/default-apps/ETHBalanceRefundApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I don't think so

packages/contracts/contracts/default-apps/ETHBalanceRefundApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: yes

packages/apps/test/high-roller-app.spec.ts
```diff
@@ -237,8 +237,8 @@ describe("HighRollerApp", () => {
       };
 
       const result = await resolve(preState);
-
-      expect(result).to.eq(1);
+      const [decodedResult] = defaultAbiCoder.decode(['uint256'], result);
```
snario: Might as well put the decoder inside the `resolve` helper fn.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "noImplicitAny": false,
-    "noUnusedLocals": false,
+    "noUnusedLocals": true,
```
snario: Is there an issue for the fundamental problem that causes this issue? It should be a lint error not a build error right?

packages/contracts/contracts/default-apps/ETHBucket.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `ETHTransfer` is a static type with 2 fields. Anyway, I have removed the wrapper functionality and disabled linting for this file.

packages/contracts/contracts/default-apps/ETHBucket.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It should be a client-level decision - this design is a hack listed in the PR message.

packages/contracts/contracts/ETHVirtualAppAgreement.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The exact same set of execution functions cannot be shared, because of the fact that for e.g. NFTs cannot be routed across a channel network

packages/contracts/contracts/StateChannelTransaction.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `Interpreter(addr).interpret(p1, p2).delegatecall()`would delegatecall the `interpret` function of `addr` using the `Interpreter` contract interface, similarly to how `Interpreter(addr).interpret(p1, p2)` calls the `interpret` function of `addr` using the `Interpreter` contract interface

packages/node/src/models/app-instance.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: yes

packages/node/src/protocol/uninstall-virtual-app.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: stands for "zeroeth address"

packages/contracts/contracts/TwoPartyVirtualEthAsLump.sol
```diff
@@ -4,16 +4,14 @@ pragma experimental "ABIEncoderV2";
...
-
-/// @title ETHVirtualAppAgreement
+/// @title TwoPartyVirtualEthAsLump
 /// @notice Commitment target to support "virtual apps", i.e., apps which have
 /// ETH committed to them via intermediary lock-up instead of having ETH directly
 /// committed in a ledger channel.
 /// The `target` contract must return via resolution the outcome of the app,
 /// and the resolve function must be TwoPartyOutcome type.
-contract ETHVirtualAppAgreement {
+contract TwoPartyVirtualEthAsLump {
```
snario: Can you write briefly, perhaps as a comment on the pull-request, _why_ we need this file versus re-using the `StateChannelTransaction.sol` file? Of course I know, but it is an interesting point worth explaining

packages/node/src/protocol/utils/get-resolution-increments.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Not possible to do so directly since they are compared with `.eq` and not `==`

packages/apps/contracts/HighRollerApp.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The old code did not split it

packages/contracts/contracts/ETHVirtualAppAgreement.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Not sure - used `abi.decode` for consistency

packages/contracts/contracts/interpreters/ETHInterpreter.sol
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `(address, uint256)[]` enforces that they are of the same length

packages/dapp-high-roller/src/components/app-provider/app-provider.tsx
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: no idea

packages/node/src/ethereum/setup-commitment.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The commitment is `cancelled` when the app is `uninstalled`

packages/node/src/ethereum/install-commitment.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: please open a separate issue for this if needed

PR #1290, opened 2019-04-11
[cf-wallet] add initial cf-wallet work

patience-tema-baron:


package.json.orig
```diff
@@ -0,0 +1,61 @@
...
+    "publish": "lerna publish --yes from-package patch",
+    "postinstall": "patch-package"
+  },
+  "keywords": [
+    "ethereum",
+    "counterfactual",
+    "state channels"
+  ],
+  "devDependencies": {
+<<<<<<< HEAD
```
snario: merge conflict

patience-tema-baron:
Current state: registration works and deposits seem to go through, though getFreeBalanceState so we don't exit the deposit screen. Additionally, the balance never updates in the menu bar. At present, the pg ui seems to be making a proper request for the balances, but it's getting a response in an incomplete format.

I've also wired up the other methods (withdraw, installVirtual, and rejectInstall), though I haven't been able to test them since I need a positive balance to do any of those actions.

@ebryn @joelalejandro

joelalejandro:
@patience-tema-baron @ebryn 

Figured out that the problem is serialization-related. The controller is working but when the payload goes through JSONAPI-TS, it loses all data. Apparently there's something wrong with the type's schema or with JSONAPI-TS's `convertToResources()`.

Issue #1250, opened 2019-04-08
[node]  De-couple the messaging layer from the Node


joelalejandro:
This is an actionable issue for #1164.

Initial attempts on implementing a first iteration of #1242 revealed that there's some previous work that needs to be done. 

The current protocol uses `NodeMessage` as an intrinsic piece of the `IMessagingService` interface. A first logical step would be to apply a generic type that allows to change the type of the `msg` parameter in both `send` and `onReceive`:

```diff
-export interface IMessagingService {
-  send(to: string, msg: NodeMessage): Promise<void>;
-  onReceive(address: string, callback: (msg: NodeMessage) => void): void;
+export interface IMessagingService<MessageFormat = NodeMessage> {
+  send(to: string, msg: MessageFormat): Promise<void>;
+  onReceive(address: string, callback: (msg: MessageFormat) => void): void;
}
```

Let's suppose we want to implement a new message called, conspicuosly called `Operation` (referring to a JSONAPI operation). The previous class, NodeMessage, has the following structure:

```ts
{
  from: string;
  type: string;
  params: { [key: string]: string | number | boolean };
}
```

While our new `Operation` message format goes like:

```ts
{
  op: "get" | "add" | "update" | "remove";
  ref: {
    id: string;
    type: string;
    relationship: string;
    lid: string;
  };
  data: {
    id: string;
    type: string;
    attributes: { 
      [key: string]: string | number | boolean; 
    };
    relationships: { 
      [key: string]: {
        data: { 
          id: string; 
          type: string; 
        } | Array<{ id: string; type: string; }>;
      }
    };
  };
  params: {};
  included: [];
  meta: { 
    [key: string]: string | number | boolean; 
  }; 
}
```

The immediate effect is that the `FirebaseMessagingService` class would no longer work, since both the `send()` and `receive()` methods expect a specific structure. Notice the `msg.from` reference:

```ts
async send(to: string, msg: NodeMessage) {
  await this.firebase
    .ref(`${this.messagingServerKey}/${to}/${msg.from}`)
    .set(JSON.parse(JSON.stringify(msg)));
}

onReceive(address: string, callback: (msg: NodeMessage) => void): void {
  // ...
  await this.firebase
    .ref(`${this.messagingServerKey}/${address}/${msg.from}`)
    .remove();
  // ...
}
```

So, one could implement a variant of `FirebaseMessagingService` which implements a `fromGetter`:

```ts
[...]

PR #1292, opened 2019-04-12
[node] Implement JSONAPI operations

joelalejandro:
This is a draft implementation of what's being discussed in #1250.


packages/node/src/api-router.ts
```diff
@@ -32,6 +32,41 @@ import {
...
+  UninstallController: { type: "app", op: "uninstall" },
+  UninstallVirtualController: { type: "app", op: "uninstallVirtual" },
+  TakeActionController: { type: "app", op: "takeAction" },
+  UpdateStateController: { type: "app", op: "updateState" },
+  GetAppInstanceController: { type: "app", op: "get", params: ["id"] },
+  GetAppInstanceStateController: { type: "app", op: "getState" },
+  GetInstalledAppInstancesController: { type: "app", op: "get" }
+};
+
 const controllers = [
```
ebryn: `const controllers = Object.keys(controllersToOperations);`

joelalejandro: Not sure about that. Notice that here we store the Controller constructors...

packages/node/src/api-router.ts
```diff
@@ -31,6 +31,41 @@ import {
 } from "./methods";
 import { NODE_EVENTS } from "./types";
 
+// Maps controllers to JSONAPI operations.
+export const controllersToOperations = {
```
snario: Why are some ops in a fixed set like `get` or `add` but others are strings like `getFreeBalanceState`?

snario: Is there an intro to `@ebryn/jsonapi-ts` with motivation?

package.json
```diff
@@ -21,6 +21,7 @@
     "lint": "lerna run lint --parallel --no-bail",
     "lint:fix": "lerna run lint:fix --parallel --no-bail",
     "run:playground": "lerna run start --parallel --scope=**/playground --scope=**/playground-server --scope=**/dapp-high-roller --scope=**/dapp-tic-tac-toe",
+    "run:playground:server": "lerna run start --parallel --scope=**/playground-server",
```
cf19drofxots: Could simply do `yarn start` from the playground-server package, no?

packages/node/src/events/propose-install-virtual/controller.ts
```diff
@@ -34,10 +34,24 @@ export default async function proposeInstallVirtualEventController(
...
   }
 
-  const relayedProposalMsg: ProposeVirtualMessage = {
-    from: requestHandler.publicIdentifier,
-    type: NODE_EVENTS.PROPOSE_INSTALL_VIRTUAL,
-    data: nodeMsg.data
+  const relayedProposalMsg: NodeOperation = {
+    meta: {
+      from: requestHandler.publicIdentifier,
+      requestId: ""
```
cf19drofxots: Probably outside the scope of this PR but we should discuss needing `requestId` - from what I've seen so far it's been empty / not referenced at all.

joelalejandro:
Cancelled in favour of JSONRPC.

PR #1312, opened 2019-04-16
Introduction of an Adjudication Layer specification

snario:
- Clearly separates the adjudication layer from the app definition layer
- Removes a (mostly useless) section, to be re-added in a better way later


packages/specs/source/02-adjudication-layer.md
```diff
@@ -0,0 +1,136 @@
+# Adjudication Layer
+
+Counterfactual's adjudication layer uses a singleton contract called the [`AppRegistry`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/AppRegistry.sol).
+
+The implementation that is inside of this repository has been designed to be exclusively compatible with applications that implement the [`CounterfactualApp`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/CounterfactualApp.sol) interface. Although this is the case, the core concepts are agnostic to the underlying interface that a state channels application might implement. A future version of the `AppRegistry` might support other types of state channel architectures such as [Force Move Games](https://github.com/magmo/force-move-games), for example.
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
```suggestion
The implementation that is inside of this repository has been designed to only be compatible with applications that implement the [`CounterfactualApp`](https://github.com/counterfactual/monorepo/blob/master/packages/contracts/contracts/CounterfactualApp.sol) interface. Although this is the case, the core concepts are agnostic to the underlying interface that a state channels application might implement. A future version of the `AppRegistry` might support other types of state channel architectures such as [Force Move Games](https://github.com/magmo/force-move-games), for example.
```

PR #1357, opened 2019-04-25
[node] Fix node test scripts

cf19drofxots:
### Description

During the transition of moving the machine package under the node package, the test script broke, preventing test regexes from being used to filter tests to run.

Now test regex for node tests can be specified via:
```
yarn test:node <regex>
```

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -8,8 +8,10 @@
   "scripts": {
     "build": "tsc -p . && rollup -c",
     "test": "yarn run test:machine && yarn run test:node",
-    "test:machine": "./scripts/test.sh",
-    "test:node": "jest --testPathIgnorePatterns=test/machine/integration --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
+    "test:machine": "yarn run test:machine:integration && yarn run test:machine:unit",
+    "test:machine:integration": "./scripts/test.sh",
+    "test:machine:unit": "jest --testPathPattern=test/machine/unit --bail --forceExit",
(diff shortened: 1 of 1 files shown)


packages/node/package.json
```diff
@@ -7,9 +7,9 @@
   "license": "MIT",
   "scripts": {
     "build": "tsc -p . && rollup -c",
-    "test": "yarn run test:no-ganache && yarn run test:with-ganache",
-    "test:with-ganache": "./scripts/test.sh",
-    "test:no-ganache": "jest --testPathIgnorePatterns=test/machine/integration --setupFiles dotenv-extended/config --runInBand --bail --forceExit",
+    "test": "yarn run test:machine && yarn run test:node",
+    "test:machine": "./scripts/test.sh",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: These are not the right names since they just check on `test/machine/integration`

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
How can test regexes be specified after this PR?

packages/node/package.json
cf19drofxots: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll actually just noticed that the existing script names are totally accurate either since the node tests do in fact spin up a ganache instance.

I'll update the naming again

cf19drofxots: Updated

cf19drofxots:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll updated the OP to include the exact command format

PR #1358, opened 2019-04-25
[monorepo] allow encoding array types

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/types/src/simple-types.ts
```diff
@@ -10,22 +10,24 @@ export type Bytes32 = string;
 // around in protocol messages and include this in transaction data in disputes,
 // we impose some restrictions on the type; they must be serializable both as
 // JSON and as solidity structs.
-export type SolidityABIEncoderV2Struct = {
+export type SolidityABIEncoderV2Type = SolidityABIEncoderV2Struct | SolidityABIEncoderV2SArray;
```
cf19drofxots: Is the union necessary here? Does the union with the last type for `SolidityABIEncoderV2Struct` not suffice?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `SolidityABIEncoderV2Struct` disallows top-level arrays (e.g. the type `[1]` is not of type `SolidityABIEncoderV2Struct`) and unfortunately `tuple([1])` seems to be deserialized as `[1]`

snario: I believe it is the difference between:

```solidity
struct State { bytes[] state; }
```

and

```solidity
bytes[] State;
```

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: no, I mean that `struct State { bytes[] state; }` gets treated as `tuple(bytes[])` which gets treated as `bytes[]`

snario: Ah.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: actually, now that I think about it, this sounds like a bug since we pass the `tuple` wrapper into ethers...

PR #1368, opened 2019-04-26
[node] remove GetMyFreeBalanceForState

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This method is not documented in API_REFERENCE and not used anywhere


snario:
Historical note: It was used in the Playground before a refactor Joel did to clean up the usage of the more abstract method which gets both users free balance state.

PR #1370, opened 2019-04-26
[node] Fix setting of deposits for non-matching amounts

cf19drofxots:
### Description

There's a bug that's causing the Node to not account for deposits of differing amounts between two participants of a channel.

For example, with a channel A<->B that's funded with 2 ETH (1 ETH from each participant respectively), A initiating an installation of an app for `(A: 0.1 ETH, B: 0 ETH)` would incorrectly leave B's state as `(myDeposit: 0.1 ETH, peerDeposit: 0 ETH)` when it should obviously be `(myDeposit: 0 ETH, peerDeposit: 0.1 ETH)`


snario:
We'll want to re-think this with #1157

PR #1393, opened 2019-05-01
[node] Postgres storage service implementation

rhlsthrm:
### Description

Create a node storage service that uses a Postgres table. The table name is `storeServiceKey` which is passed into the factory. The table has a simple schema defined as:
```sql
CREATE TABLE table_name (
    id SERIAL PRIMARY KEY,
    key text UNIQUE,
    value jsonb
);
```

### Related issues

[Link here any issues relevant to this PR, using the GitHub `fixes/resolves/closes` keywords to close related issues automatically.]


cf19drofxots:
@rhlsthrm looks like `package.json` wasn't included in the commit to include the new package (`sql-template-strings`) being used which is causing the build to fail.

rhlsthrm:
Added the packages, looks like there's still some errors and conflicts. I'll pull in upstream and resolve the conflicts.

packages/node/src/services.ts
```diff
@@ -202,3 +204,99 @@ export function confirmLocalFirebaseConfigurationEnvVars() {
...
+  async get(key: string): Promise<any> {
+    const result = await this.pgClient.query(
+      SQL`SELECT * FROM "`
+        .append(this.storeServiceKey)
+        .append(SQL`" WHERE "key" = ${key}`)
+    );
+    return result.rows[0] && result.rows[0].value[key];
+  }
+
+  async set(pairs: { key: string; value: any }[]): Promise<any> {
```
cf19drofxots: @rhlsthrm can we make this a transaction across the set of pairs that's being written?

Seems like the current implementation loses atomicity which could potentially cause state dissonance.

rhlsthrm: This is a transaction, that's what the `BEGIN` and `COMMIT` queries are for. The query will be atomic because of this.

rhlsthrm: Ah but I still need to catch the error and `ROLLBACK`. I'll address this.

cf19drofxots: Eh I should have been more specific. I meant across all pairs, instead of doing the write per iteration of the loop.

Similar to this: https://github.com/counterfactual/monorepo/blob/master/packages/node/src/services.ts#L180

rhlsthrm:
Our team has always found more flexibility in using raw SQL strings as opposed to ORMs, but I am fine with moving this to an ORM if you are more comfortable with that @snario. Also, please advise on how to create it as a plugin? I can work with @cf19drofxots on that.

ebryn:
We are fans of Knex.js & use it heavily

rhlsthrm:
Thanks for the tip on Knex.js. I'll look into that and refactor this appropriately.

cf19drofxots:
> I believe this should end up being a standalone package, no? Perhaps as its own package node-postgresql-store-plugin or something like this? This would avoid the dependencies being added into the node itself.

I was thinking about this too, and we definitely should start moving towards a plugin system. Though to unblock the integration this is fine, then we can refactor to use a plugin when that exists.

packages/node/src/services.ts
```diff
@@ -202,3 +204,99 @@ export function confirmLocalFirebaseConfigurationEnvVars() {
...
+  }
+}
+
+class PostgresStoreService implements IStoreService {
+  constructor(
+    private readonly pgClient: Client,
+    private readonly storeServiceKey: string
+  ) {}
+
+  async get(key: string): Promise<any> {
```
joelalejandro: @rhlsthrm I think we can type more accurately here by returning `Promise<StringKeyValue>`:

```ts
type StringKeyValue = { [key: string]: StringKeyValue };

async get(key: string): Promise<StringKeyValue>
```

We're trying to stear clear of `any` as much as possible.

packages/node/src/services.ts
joelalejandro: @rhlsthrm Same remark about typing as before :)

cf19drofxots:
hey @rhlsthrm any updates here?

rhlsthrm:
Hi all, sorry for the huge delay in pushing this. I finally went ahead and made the requested changes to get it into an ORM.

packages/node/src/entity/Node.ts
```diff
@@ -0,0 +1,12 @@
+import { Column, Entity, PrimaryColumn } from "typeorm";
+
+@Entity()
+export class Node {
+  @PrimaryColumn()
+  // @ts-ignore
```
rhlsthrm: Typescript doesn't seem to like these declarations.

packages/node/src/services.ts
```diff
@@ -232,3 +236,102 @@ export function confirmLocalFirebaseConfigurationEnvVars() {
...
+    if (!allowDelete && containsNull(pairs)) {
+      throw new Error(WRITE_NULL_TO_FIREBASE);
+    }
+
+    const connection = this.connectionMgr.get();
+
+    await connection.transaction(async transactionalEntityManager => {
+      for (const pair of pairs) {
+        const storeKey = `${this.storeServiceKey}_${pair.key}`;
+        // wrap value so its always JSON
```
rhlsthrm: This is necessary for Postgres because the JSON column breaks if you use anything other than JSON (i.e. a raw string). In some cases, the node code is inserting strings as values instead of objects.

packages/node/src/services.ts
```diff
@@ -232,3 +236,102 @@ export function confirmLocalFirebaseConfigurationEnvVars() {
...
+        const storeValue = {
+          [pair.key]: JSON.parse(JSON.stringify(pair.value))
+        };
+        let record = await transactionalEntityManager.findOne(
```
rhlsthrm: I tried hard to make this "upset" instead of read/write in two queries. It was almost impossible without using raw SQL. We can revisit this in the future.

packages/node/src/services.ts
```diff
@@ -232,3 +236,102 @@ export function confirmLocalFirebaseConfigurationEnvVars() {
...
+        );
+        if (!record) {
+          record = new NodeEntity();
+          record.key = storeKey;
+        }
+        record.value = storeValue;
+        await transactionalEntityManager.save(record);
+      }
+    });
+    return true;
```
rhlsthrm: Not sure what exactly to return here, so I punted to return true. Is this value used anywhere?

packages/playground/src/index.html
```diff
@@ -40,6 +40,10 @@
       constants = ethers.constants;
       providers = ethers.providers;
       database = firebase.database;
+      // nullify these as they're expected as globals for the node
+      // but aren't used browserside
+      pg = null;
```
rhlsthrm: Might not need this, or it might not be the same @cf19drofxots.

rhlsthrm:
I still have some questions on how to move forward with this. It would be good for the rest of the team to start using the Postgres implementation instead of Firebase so we can eventually remove that dependency. The model right now is very simple, it is a key/value store. I think we should move to expanding the node schema to store the node data as proper relational data instead of purely key/value stores, what do you guys think @cf19drofxots, @snario, @ebryn?

Also another thing I haven't tackled is migrations. Right now this requires a database table set up correctly to work. I'd also like some thoughts on how to structure that. Should we have a migration flow in this package?

snario:
@rhlsthrm the strategy @cf19drofxots and I spoke about was to update the `playground-server` package to use this postgresql store service and ship the Playground with that update as a means of ensuring this is production ready.

cf19drofxots:
> I still have some questions on how to move forward with this. It would be good for the rest of the team to start using the Postgres implementation instead of Firebase so we can eventually remove that dependency. The model right now is very simple, it is a key/value store. I think we should move to expanding the node schema to store the node data as proper relational data instead of purely key/value stores

This seems like a 3-phase procedure:
1. This PR implementing the postgres functionality
2. Getting it used in the playground-server with the initial schema that we have (high-level key/value imitation)
3. Expanding the schema. The key/value usage right now is simple enough that converting the format to be relational won't pose a challenge. Given that phase 2 puts the native schema in production, it seems that introducing migrations seems appropriate as a part of this task / PR.

rhlsthrm:
Thanks for the notes, @snario and @cf19drofxots. I fixed the latest conflict, so if CI runs successfully, this should be good to merge.

packages/node/src/services.ts
```diff
@@ -232,3 +236,102 @@ export function confirmLocalFirebaseConfigurationEnvVars() {
...
+  }
+
+  async get(key: string): Promise<StringKeyValue | string> {
+    const storeKey = `${this.storeServiceKey}_${key}`;
+    const res = await this.connectionMgr
+      .get()
+      .manager.findOne(NodeEntity, storeKey);
+    return res && res.value[key];
```
cf19drofxots: Maybe if `res` isn't defined return a rejection, otherwise get the value?

packages/playground/src/index.html
cf19drofxots: This is a workaround that we unfortunately still need. It's a small thing so I wouldn't worry about it for now.

packages/node/src/node.ts
```diff
@@ -4,6 +4,7 @@ import { SigningKey } from "ethers/utils";
 import { fromExtendedKey, HDNode } from "ethers/utils/hdnode";
 import EventEmitter from "eventemitter3";
 import * as log from "loglevel";
+import "reflect-metadata";
```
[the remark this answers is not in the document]
rhlsthrm: https://typeorm.io/#/, unfortunately yes. It does say to "import it somewhere in the global place of your app (for example in app.ts)", so maybe this isn't the best place to put this import.

packages/node/src/entity/Node.ts
```diff
@@ -0,0 +1,12 @@
+import { Column, Entity, PrimaryColumn } from "typeorm";
```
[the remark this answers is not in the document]
rhlsthrm: That's the convention with the entities, we could change it though.

packages/node/src/services.ts
```diff
@@ -232,3 +236,102 @@ export function confirmLocalFirebaseConfigurationEnvVars() {
...
+    private readonly connectionMgr: ConnectionManager,
+    private readonly storeServiceKey: string
+  ) {}
+
+  async set(
+    pairs: { key: string; value: any }[],
+    allowDelete?: Boolean
+  ): Promise<boolean> {
+    if (!allowDelete && containsNull(pairs)) {
+      throw new Error(WRITE_NULL_TO_FIREBASE);
```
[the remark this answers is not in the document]
rhlsthrm: Ok, I assumed this was so we can control deletions to avoid removing items. Will remove.

packages/node/src/services.ts
rhlsthrm: Added `undefined` as a possible type to make this more clear.

packages/node/src/entity/Node.ts
rhlsthrm: https://github.com/typeorm/typeorm/issues/2936 found a weird workaround, the issue is the `strictPropertyInitialization` flag in the tsconfig.

rhlsthrm:
@snario are there any other changes you're requiring for this to get merged? It looks like everything has been addressed.

PR #1419, opened 2019-05-07
[node] generate network contexts using networkContextProps

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/node/src/network-configuration.ts
```diff
@@ -40,26 +40,13 @@ interface Migration {
...
-      migrations,
-      "ETHVirtualAppAgreement"
-    ),
-    MinimumViableMultisig: getContractAddress(
-      migrations,
-      "MinimumViableMultisig"
-    ),
-    ProxyFactory: getContractAddress(migrations, "ProxyFactory")
-  };
+  const ret = {} as any;
```
cf19drofxots: Why `any` if the expected type is `NetworkContext`?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `{}` does not conform to the type `NetworkContext`

cf19drofxots: `tsc` doesn't error on that, but carry on. Just generally avoidant of using `any`

PR #1426, opened 2019-05-07
[playground] Skip PG tests

cf19drofxots:
This temporarily removes the PG tests from running as no change is being made to the PG.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,10 +92,6 @@ jobs:
       - run:
           name: Run Tic-Tac-Toe tests
           command: cd packages/dapp-tic-tac-toe && yarn test
-      - run:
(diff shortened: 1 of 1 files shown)


snario:
Make sure there is an issue to add these back in that is P1.

Alonski:
Is it not an issue that we could potentially break the Playground?

cf19drofxots:
@Alonski not a big concern atm since we're not changing the playground much but the root cause is non-trivial to track down and it's causing all other PRs to continuously get delayed for irrelevant reasons.

cf19drofxots:
Created this https://github.com/counterfactual/monorepo/pull/1428 as a placeholder to revert this change once playground modification starts again.

PR #1443, opened 2019-05-08
[documentation] Add encoding string examples to the API_REFERENCE

cf19drofxots:
This adds example strings for the ABI encodings to make it easier to newer developers with understanding how the encoding is supposed to be expressed for an App.

### DIFF
--- a/packages/cf.js/API_REFERENCE.md
+++ b/packages/cf.js/API_REFERENCE.md
@@ -494,10 +494,12 @@ An instance of an installed app.
 ### Data Type: `AppABIEncodings`
 - `stateEncoding: string`
     - ABI encoding of the app state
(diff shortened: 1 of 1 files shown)


packages/cf.js/API_REFERENCE.md
```diff
@@ -494,10 +494,12 @@ An instance of an installed app.
 ### Data Type: `AppABIEncodings`
 - `stateEncoding: string`
     - ABI encoding of the app state
+        - For eg, for the Tic Tac Toe App (https://github.com/counterfactual/monorepo/blob/master/packages/apps/contracts/TicTacToeApp.sol), the state encoding string is `"tuple(address[2] players, uint256 turnNum, uint256 winner, uint256[3][3] board)"`.
```
snario:
```suggestion
        - For example, for the Tic Tac Toe App (https://github.com/counterfactual/monorepo/blob/master/packages/apps/contracts/TicTacToeApp.sol), the state encoding string is `"tuple(address[2] players, uint256 turnNum, uint256 winner, uint256[3][3] board)"`.
```

PR #1446, opened 2019-05-09
[node] More refactoring of tests

cf19drofxots:
This refactors more of the Node tests to make them more straightforward.


packages/node/test/integration/channel-creation.spec.ts
```diff
@@ -28,63 +24,21 @@ describe("Node can create multisig, other owners get notified", () => {
...
-      const txHash2 = await getMultisigCreationTransactionHash(
-        nodeA,
-        ownersACPublicIdentifiers
-      );
-
-      expect(txHash1).toBeDefined();
-      expect(txHash2).toBeDefined();
+    it("Node A can create multiple back-to-back channels with Node B and Node C", async () => {
+      await createChannel(nodeA, nodeB);
+      let openChannelsNodeA = await getChannelAddresses(nodeA);
```
snario: We can avoid using `let` and keep things functional here. I generally prefer to only use mutable variables when there is a deep need / using `consts` is ridiculously over-complicated.

packages/node/test/integration/channel-creation.spec.ts
```diff
@@ -28,63 +24,21 @@ describe("Node can create multisig, other owners get notified", () => {
...
-
-      const txHash2 = await getMultisigCreationTransactionHash(
-        nodeA,
-        ownersACPublicIdentifiers
-      );
-
-      expect(txHash1).toBeDefined();
-      expect(txHash2).toBeDefined();
+    it("Node A can create multiple back-to-back channels with Node B and Node C", async () => {
+      await createChannel(nodeA, nodeB);
```
snario: Don't you kind of want to _not_ use `await` here so you can issue two instantly back to back calls to the node and not result on the promise resolving? I think that's what the purpose of the test is, right?

packages/node/test/integration/utils.ts
```diff
@@ -450,12 +450,21 @@ export async function collateralizeChannel(
 
 export async function createChannel(nodeA: Node, nodeB: Node) {
   return new Promise(async (resolve, reject) => {
+    let nodeBChannelCreationAck = false;
     nodeA.on(NODE_EVENTS.CREATE_CHANNEL, async () => {
       expect(await getInstalledAppInstances(nodeA)).toEqual([]);
       expect(await getInstalledAppInstances(nodeB)).toEqual([]);
+
+      while (!nodeBChannelCreationAck) {
+        await sleep(500);
```
snario: This is a wacky loop. Why do this? Just only resolve the promise after the event returns in side this block:

```ts
 nodeB.once(NODE_EVENTS.CREATE_CHANNEL, async () => {
      nodeBChannelCreationAck = true;
// resolve() here?
    });
```

packages/node/test/integration/channel-creation.spec.ts
cf19drofxots: I'll have to refactor this more in general which will address all of your comments.

packages/node/src/models/state-channel.ts
```diff
@@ -441,7 +441,7 @@ export class StateChannel {
     }
     if (updatedBobBalance.lt(Zero)) {
       throw Error(
-        `\n${bob} cannot install virtual app agreement instance. Its balance in channel with ${alice} is insufficient ${formatEther(
+        `\n${bob} cannot install virtual app agreement instance. Its balance in channel with ${alice} is insufficient by ${formatEther(
```
snario: Newline?

PR #1453, opened 2019-05-09
Run yarn upgrade and regenerate yarn.lock

### LINKED ISSUE #845: Use the same version of Jest across the entire repo
`dapp-tic-tac-toe` requires Jest 23 because `react-scripts` has not updated yet. See [here](https://github.com/facebook/create-react-app/pull/6278) for more. Since `react-scripts` has a dependency on an older version, it causes some issues with the monorepo correctly installing both.

To resolve this (and until [`react-scripts` upgrades to Jest 24](https://github.com/facebook/create-react-app/pull/6278) we install Jest 23 as default on the majority of packages (so it is hoisted to the root) and Jest 24 only on `node` and `playground` because it uses the `globalSetup` and `globalTeardown` tranformer functionality [released in Jest 24](https://github.com/facebook/jest/issues/5164).

snario:
Ran `yarn upgrade` and upgraded everything except `@ebryn/jsonapi-ts` and `truffle`. The `jsonapi-ts` has breaking changes (cc: @joelalejandro) and `truffle` seems to have introduced a major bug in their compiler 😢.

This fixes #845 too incidentally.

I had to regenerate `yarn.lock` because of this: https://stackoverflow.com/questions/55681358/browserslisterror-unknown-browser-kaios


snario:
Most of them are blocked by [`renovate.json`](https://github.com/counterfactual/monorepo/blob/master/renovate.json) because of #845.

PR #1454, opened 2019-05-09
[contracts] fix some visibilities

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/apps/contracts/HighRollerApp.sol
```diff
@@ -155,7 +155,7 @@ contract HighRollerApp is CounterfactualApp {
   }
 
   function getWinningAmounts(uint256 num1, uint256 num2, uint256 termsLimit)
-    public
```
snario: I believe this is called by the JS application at the moment... could be wrong but I'd check.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: it's unused

snario: okay

Issue #669, opened 2019-02-05
[monorepo] Contract tests fail on second consecutive run, probably b/c of machine tests from first run


emansipater:
Most likely this is because the machine uses truffle to build artifacts but the contracts folder uses waffle, and we were previously relying on the assumption that the build output was the same for both (when both were on truffle).

snario:
Potential fix: have `machine` make a second folder to put build artifacts in vs relying on the `build` folder in contracts. The reason it uses that now is because on the _first_ run it can re-use the compile step from the contracts `yarn build` command instead of doing another compilation.

PR #1479, opened 2019-05-14
[node] selectively add 0x prefix

### LINKED ISSUE #1461: Running tests twice fails with `0x0x` in bytecode
Running the following from a fresh clone errors out

```
yarn && yarn build && yarn test && cd packages/node && yarn test
```

with the following

```
invalid hexidecimal string (arg="value", value="0x0x608060405234801561001057600080fd5b506040516020806101a88339810180604052602081101561003057600080fd5b8101908080519060200190929190505050600073ffffffffffffffffffffffffffffffffffffffff168173ffffffffffffffffffffffffffffffffffffffff1614156100c7576040517f08c379a00000000000000000000000000000000000000000000000000000000081526004018080602001828103825260248152602001806101846024913960400191505060405180910390fd5b806000806101000a81548173ffffffffffffffffffffffffffffffffffffffff021916908373ffffffffffffffffffffffffffffffffffffffff16021790555050606e806101166000396000f3fe608060405273ffffffffffffffffffffffffffffffffffffffff600054163660008037600080366000845af43d6000803e6000811415603d573d6000fd5b3d6000f3fea165627a7a72305820694407262d1bf8a12b96a9ab0006c182bd726d829087da293c23bcb1e4bd03930029496e76616c6964206d617374657220636f707920616464726573732070726f7669646564", version=4.0.27)
```

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
fixes #1461, #669

### DIFF
--- a/packages/node/src/methods/state-channel/create/controller.ts
+++ b/packages/node/src/methods/state-channel/create/controller.ts
@@ -56,6 +56,11 @@ export default class CreateChannelController extends NodeController {
       MinimumViableMultisig.abi
     ).functions.setup.encode([multisigOwners]);
 
+    let bytecode = Proxy.bytecode;
+    if (!bytecode.startsWith("0x")) {
+      bytecode = `0x${bytecode}`;
+    }
+
     const multisigAddress = getAddress(
       solidityKeccak256(
         ["bytes1", "address", "uint256", "bytes32"],
@@ -66,7 +71,7 @@ export default class CreateChannelController extends NodeController {
           keccak256(
             solidityPack(
               ["bytes", "uint256"],
-              [`0x${Proxy.bytecode}`, networkContext.MinimumViableMultisig]
+              [bytecode, networkContext.MinimumViableMultisig]
             )
           )
         ]


snario:
Feels like a patch, not a fix?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
what's the fix?

snario:
On the first-time around of running `yarn test` we compile the contracts for the first time time with `ethereum-waffle` (in the `contracts` package) and then a second time with `truffle` (in the `node` package). The second time around, when the tests run for the Node, it runs `truffle migrate` and tries to _recompile_ the contracts and there is some discontinuity with how `ethereum-waffle` and `truffle` handle compilation that leads to this weird result.

I think the root cause is this compilation discontinuity.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
So what if we use two methods to recompile? AFAICT it's not specified anywhere whether the `bytecode` field should start with `0x`, so the current code on master is relying on implementation details.

snario:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll I think a temporary fix would be to delete the compiled contracts on `yarn test`. The specification is based on what the Truffle output is (`ethereum-waffle` is set to use "legacyOutput" which is _supposed to be_ the Truffle output; maybe double check that flag is working).

https://github.com/counterfactual/monorepo/blob/0c78d92716f975d5d183aed7ff4bc590be63d504/packages/contracts/waffle.js#L7

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
> The specification is based on what the Truffle output is

again, this looks like implementation behaviour to me, since truffle doesn't seem (to me) to make a guarantee about what the bytecode field should look like

snario:
Yes @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll, correct, but that is not the issue here. The issue is that running it twice changes the behaviour, right?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Incorrect, the issue is what is stated in the title of https://github.com/counterfactual/monorepo/issues/1461

snario:
Sorry that is what I mean. 

I still recommend deleting the build artifacts when you start the tests.

cf19drofxots:
This was fixed in https://github.com/counterfactual/monorepo/pull/1524

PR #1480, opened 2019-05-14
[node] throw error on null values

### LINKED ISSUE #1416: Firebase silently throws away keys with null/undefined value
This leads to hard-to-debug errors if we try to save and restore an object (e.g. AppInstanceJson) with a field with a value like that. We should warn about this when serializing.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
fixes #1416

### DIFF
--- a/packages/node/src/services.ts
+++ b/packages/node/src/services.ts
@@ -149,6 +149,20 @@ class FirebaseMessagingService implements IMessagingService {
   }
 }
 
+function containsNull(obj) {
+  for (const key in obj) {
+    if (typeof obj[key] === "object") {
+      if (containsNull(obj[key])) {
+        return true;
+      }
+    }
+    if (obj[key] === null || obj[key] === undefined) {
+      return true;
+    }
+  }
+  return false;
+}
+
 class FirebaseStoreService implements IStoreService {
   constructor(
     private readonly firebase: firebase.database.Database,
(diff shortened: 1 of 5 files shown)


packages/node/src/services.ts
```diff
@@ -172,11 +189,23 @@ class FirebaseStoreService implements IStoreService {
...
+  async set(
+    pairs: { key: string; value: any }[],
+    allowDelete?: Boolean
+  ): Promise<any> {
     const updates = {};
     for (const pair of pairs) {
       updates[pair.key] = JSON.parse(JSON.stringify(pair.value));
     }
+    if (!allowDelete && containsNull(updates)) {
+      throw new Error("Firebase store service found null/undefined value");
```
cf19drofxots: This description is a little misleading since the store service itself hasn't found `null` or `defined` - we're just not writing these values to the store by choice.

So maybe, "The records being set contain `null` values" or something like that.

cf19drofxots: And maybe add this as an ERROR constant to the set of errors we have, so we don't copy the string literal as we're doing in the test.

cf19drofxots: Also this directly throws - I thought the suggestion was to warn and not error?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I think developers naturally ignore warnings. Since there's a switch anyway, it should error unless explicitly told not to.

cf19drofxots: Makes sense - I think the comment about the error message still stands tho

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: fixed

PR #1486, opened 2019-05-15
Move docker image for CI to our image + autobuilding

### LINKED ISSUE #1434: [repo] solc must be dynamically updateable via circle.yml
Presently `circle.yml` relies on `web3docker/counterfactual:0.0.1` which is owned by @AlexXiong97. We need to decentralize this :)

snario:
Fixes #1434.

Made an account on Docker Hub and `counterfactual` organization we can re-use.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
     working_directory: /home/circleci/project
     # use customized docker (ubuntu, yarn 1.15, node10, solc binary)
     docker:
-      - image: web3docker/counterfactual:0.0.1
+      - image: counterfactual/circleci-environment:latest
     steps:
       - <<: *restore_code
       - checkout


cf19drofxots:
We should put the recipe in the repo so it's easy to update by anyone

snario:
Basically when they are rolled up two parties A and B only need to do one round trip to execute a protocol. A signs and sends to B, B countersigns and sends to A, done. The problem is that you are signing a transaction which does multiple things in one go, not all of which will strictly need to be executed on chain in one go.

When they are not rolled up, you end up with multiple commitments which can be put on-chain piecemeal but the tradeoff is that the protocol becomes multiple round trips. A to B, B to A, A back to B, B back to A, done, for example.

We have decided the one-round trip goal is not that advantageous / worth it for the trouble this causes handling on chain challenges.

snario:
@cf19drofxots [.circleci/images/build/Dockerfile](https://github.com/counterfactual/monorepo/blob/master/.circleci/images/build/Dockerfile)

PR #1516, opened 2019-05-20
[specs] clean up intro

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- try to reduce word count a bit, improve english etc
- reorder "Framework Design Goals"
- move "Protocol Structure" to protocol.md


packages/specs/source/app-definition.md
```diff
@@ -18,7 +18,7 @@ A slightly more complicated example would be a Tic-Tac-Toe game:
...
+There is, of course, an important difference in the example of a Tic-Tac-Toe game though. That difference is that in the Tic-Tac-Toe game, the final resolution of the state of the application is not always defined given only the current state of the application. In the payment channel example, if one user were to become unresponsive it is is easy to see how the resolution would play out; Alice would receive `aBal` and Bob would receive `bBal`. In Tic-Tac-Toe, however, if the game is not finished yet, there are some moves left to be made to reach a "terminal" state which can then be resolved. Terminal states comprise those in which X wins, O wins, or there is a draw.
```
cf19drofxots:
```suggestion
There is, of course, an important difference in the example of a Tic-Tac-Toe game though. That difference is that in the Tic-Tac-Toe game, the final resolution of the state of the application is not always defined given only the current state of the application. In the payment channel example, if one user were to become unresponsive it is easy to see how the resolution would play out; Alice would receive `aBal` and Bob would receive `bBal`. In Tic-Tac-Toe, however, if the game is not finished yet, there are some moves left to be made to reach a "terminal" state which can then be resolved. Terminal states comprise those in which X wins, O wins, or there is a draw.
```

Issue #1221, opened 2019-04-02
[repo] Investigate flaky builds on Netlify


snario:
Occasionally Netlify builds seem to throw errors in inconsistent ways. Let's figure out why.

joelalejandro:
Seems that with #2033 in progress, this is no longer needed.

PR #1536, opened 2019-05-23
Rename dispute to challenge everywhere

snario:
Generally we've agreed on this new naming, so let's use the word challenge instead of dispute.


snario:
Grrr. #1221

PR #1544, opened 2019-05-23
Add .nvmrc and regenerate yarn.lock

snario:
I tried to regenerate `yarn.lock` and some of the dependencies threw errors because I was using `node` at version `10.5.0`. So, I added a `.nvmrc` and did `nvm use 10.15.3` and then running `yarn` worked and produced the `yarn.lock` file in this PR.


snario:
Also I updated the `Dockerfile` because it is used in CircleCI but uses 10.4.0.

URL to see build: https://cloud.docker.com/u/counterfactual/repository/docker/counterfactual/circleci-environment/builds

PR #1545, opened 2019-05-23
[node] Extract firebase implementations of store & messaging service from Node package

cf19drofxots:
This extracts the implementation of the store and messaging services that were placed in the Node package to separate packages.

The package `firebase-client` package implements client functionality for the store and messaging services.

The package `firebase-server` extends the client's factory interface to spin up a firebase server such that whoever is running a `firebase-client` can connect to a firebase server.


snario:
Can you add some detail on what the two new packages introduced are?

Issue #950, opened 2019-03-05
Remove external eth_calls by using a JS-based EVM interpreter


snario:
Calling `eth_call` for every `applyAction` and `resolve` sucks. It is slow.

We can use a JS-based EVM interpreter instead. https://github.com/ethereumjs/ethereumjs-vm offers a `runCode` method we could use.

There is a subtle problem which is that for some specific apps we allow `view` functions (e.g., `ETHBalanceRefundApp` to observe `address(multisigAddress).balance`). So, those ones need to use `eth_call`. Maybe you can observe the ABI of the contract and conditionally use `eth_call` for those.

snario:
@ricmoo how's this coming along with `ethers`? @pkieltyka any ideas?

ricmoo:
Still trying to get v5 into public beta, then I’ll start making the EVM public through its experimental package.

snario:
Sweet.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 750.0 DAI (750.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/counterfactual/monorepo/950/2584).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/counterfactual/monorepo/950/2584).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $41,591.17 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [iamonuwa](https://gitcoin.co/profile/iamonuwa) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/counterfactual/monorepo/950/2584?mutate_worker_action=approve&worker=iamonuwa) | [reject worker](https://gitcoin.co/issue/counterfactual/monorepo/950/2584?mutate_worker_action=reject&worker=iamonuwa))_.**

[...]

PR #1550, opened 2019-05-24
[JSON-RPC][Run eth_call in local EVM]

StevenJNPearce:
### Description

Use ethereumjs-vm to run pure functions on local machine. Reduces unnecessary calls to node. I've implemented this a child of Ethers' JsonRpcProvider; it should work as a drop-in replacement.

### Related issues

Fixes: #950


packages/ethers-cf-purelocal/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const vm = new VM({
+  enableHomestead: true,
+  activatePrecompiles: true
+});
+export class PureEvmProvider extends ethers.providers.JsonRpcProvider {
+  contracts;
+  constructor(...args) {
+    super(...args);
+    this.contracts = require(`@counterfactual/contracts/networks/${
```
snario: Ought to remove the dependence on our contracts.

packages/ethers-cf-purelocal/index.ts
```diff
@@ -0,0 +1,75 @@
+import { ethers } from "ethers";
+import VM from "ethereumjs-vm";
+
+const vm = new VM({
+  enableHomestead: true,
+  activatePrecompiles: true
+});
+export class PureEvmProvider extends ethers.providers.JsonRpcProvider {
```
snario: Intuitively the thing to extend seems like `Contract` (so you can access the `abi` property)

packages/ethers-cf-purelocal/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "ethers-cf-purelocal",
+  "version": "1.0.0",
```
snario: Maybe start at a lower version.

packages/ethers-cf-purelocal/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "tsc && node test.js"
```
snario: Useful to split up the build and test steps. Also, we can use `jest` to run the tests (which will read `.ts` files)

packages/ethers-cf-purelocal/package.json
```diff
@@ -0,0 +1,16 @@
...
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@counterfactual/contracts": "0.0.3",
```
snario: Should remove this dependency

packages/ethers-cf-purelocal/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "ethers-cf-purelocal",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "tsc && node test.js"
+  },
+  "author": "",
+  "license": "ISC",
```
snario: I don't know what ISC is but everything we use is MIT

packages/ethers-cf-purelocal/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "dependencies": {
+    "@counterfactual/contracts": "0.0.3",
+    "ethereumjs-vm": "^3.0.0",
+    "ethers": "^4.0.27"
```
snario: Can fix this to 4.0.27 instead of allowing above

packages/ethers-cf-purelocal/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "tsc && node test.js"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@counterfactual/contracts": "0.0.3",
+    "ethereumjs-vm": "^3.0.0",
```
snario: Should also fix this

packages/ethers-cf-purelocal/test.ts
```diff
@@ -0,0 +1,17 @@
+import { ethers } from "ethers";
+import { PureEvmProvider } from "./index";
+import assert = require("assert");
+
+let provider = new PureEvmProvider("https://ropsten.infura.io/v3/aa7e165c61a84f25a85905e498267ca7", 'ropsten');
```
snario: Why use Ropsten in the test?

packages/ethers-cf-purelocal/tsconfig.json
```diff
@@ -0,0 +1,60 @@
+{
```
snario: You can inherit from [`tsconfig.json`](https://github.com/counterfactual/monorepo/blob/master/tsconfig.json)

StevenJNPearce:
@snario Thank you for the review, I'll go through it and make the changes tomorrow.

packages/ethers-cf-purelocal/index.ts
StevenJNPearce: @snario Without including the contracts package we don't actually have the code we want to run on EVM; I will add a call to getCode() to the contract constructor, unless you have a better idea for how to implement?

p.s. I recognize the face; I think we met briefly at ETHSingapore :)

snario: The `Contract` object has `this.address` and `this.abi` parameters. We could fetch the bytecode from the chain. Alternatively we could use the `ContractFactory` object to pass in the bytecode but no address.

snario: Hm. This is where I'd love to have @ricmoo chime in...

StevenJNPearce: @snario One solution I can see here is too add the methods under their own object, e.g. Contract.localFunctions._functionName_ . Would this be acceptable for your use case?

PR #1540, opened 2019-05-23
Rename "resolution" to "outcome" and "resolve" to "computeOutcome"

snario:
Based on discussions in standardization workshop.

PR #1553, opened 2019-05-24
Fixes the issue created on the Playground after Interpreters PR

snario:
Root cause was that the timeout was removed: https://github.com/counterfactual/monorepo/commit/3f71398a1584d37e20ffed362fa5649def398473#diff-11bca43397c368ea24a97a786e791205L56

However in the meantime we have also changed the names of functions in the #1540, so needed to redeploy addresses anyway.

### DIFF
--- a/packages/node/src/methods/state-channel/create/controller.ts
+++ b/packages/node/src/methods/state-channel/create/controller.ts
@@ -11,6 +11,7 @@ import {
   solidityKeccak256,
   solidityPack
 } from "ethers/utils";
+import log from "loglevel";
 import Queue from "p-queue";
 
 import { xkeysToSortedKthAddresses } from "../../../machine";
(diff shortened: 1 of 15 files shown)


packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -76,11 +77,14 @@ export default class CreateChannelController extends NodeController {
...
     );
 
+    console.log("multisig address: ", multisigAddress);
+
     const tx = await this.sendMultisigDeployTx(
       wallet,
       setupData,
       networkContext
     );
+    console.log();
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: ?

cf19drofxots: Ignore the logs, they'll all be removed

joelalejandro:
I see no unintended side-effects on PG UI/PG Server with these changes. ✅

cf19drofxots:
@joelalejandro this was strictly an issue caused by a call to the Node when carrying out safe deposits

PR #1561, opened 2019-05-24
[apps] Unidirectional ETH payment app

rhlsthrm:
### Description

We need an app that can be installed virtually to transfer ETH across intermediaries. Eventually this will be integrated into the framework, but for now I would like this to be pulled into master so that it can stay up to date with the interface changes.


rhlsthrm:
Wrote a test for the outcome which is failing: 
```ts
it("can compute the outcome", async () => {
  const senderAddr = mkAddress("0xa");
  const receiverAddr = mkAddress("0xb");
  const senderAmt = new BigNumber(10000);
  const preState: EthPaymentAppState = {
    transfers: [
      {
        to: senderAddr,
        amount: senderAmt
      },
      {
        to: receiverAddr,
        amount: Zero
      }
    ]
  };

  expect(await computeOutcome(preState)).to.eq(OutcomeType.ETH_TRANSFER);
});
```

I believe I'm probably doing something wrong with this assertion, which was copied from the `high-roller-app.spec.ts`. Any suggestions on how to fix or other tests I should put in place would be great.

packages/apps/contracts/EthPaymentApp.sol
```diff
@@ -0,0 +1,90 @@
...
+    AppState memory state = abi.decode(encodedState, (AppState));
+    return state.transfers[0].to;
+  }
+
+  function computeOutcome(bytes calldata encodedState)
+    external
+    pure
+    returns (bytes memory)
+  {
+    return encodedState;
```
snario: I suppose this is a nice property though of the app state structure

packages/apps/contracts/EthPaymentApp.sol
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  function isStateTerminal(bytes calldata encodedState)
+    external
+    pure
+    returns (bool)
+  {
+    // always terminal
+    return true;
+  }
```
snario: I think you might want to restructure it such that there is a receiver "finalize payment" action where that moves it into a terminal state but whenever the sender sends funds it is not in a finalized state. The reason being is that technically right now the Node is written such that the counterparty will auto-sign the new state and return it to the sender every time. Thus, the sender in this app would have a final state for every time he sent money. So he could submit state with nonce = 1 where he only sent a small amount and it would execute the effect on chain immediately because it is terminal even though a nonce = 10 might exist.

That being said, if the Node did not auto-sign everything and the receiver simply did not return back the doubly signed copy to the sender, having every state being final would work fine in the way this contract is written already. However with one change. The actions would need to be updated in a way where the `paymentAmount` is the _total_ amount that has been sent thus far, not just the amount sent in this one payment. If you go with the approach I mention in the first paragraph though, you don't need to do this.

packages/apps/package.json
```diff
@@ -29,7 +29,7 @@
     "ethlint": "1.2.4",
     "mocha": "6.1.4",
     "solc": "0.5.8",
-    "truffle": "5.0.10",
+    "truffle": "5.0.19",
```
snario: ?

packages/apps/contracts/EthPaymentApp.sol
```diff
@@ -0,0 +1,90 @@
+pragma solidity 0.5.8;
+pragma experimental "ABIEncoderV2";
+
+import "@counterfactual/contracts/contracts/interfaces/CounterfactualApp.sol";
+import "@counterfactual/contracts/contracts/interfaces/TwoPartyOutcome.sol";
+import "@counterfactual/contracts/contracts/interfaces/Interpreter.sol";
+import "@counterfactual/contracts/contracts/interpreters/ETHInterpreter.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+contract EthPaymentApp is CounterfactualApp {
```
snario: Maybe `EthUnidirectionalPaymentApp`?

packages/apps/contracts/EthPaymentApp.sol
```diff
@@ -0,0 +1,90 @@
...
+
+  using SafeMath for uint256;
+
+  struct AppState {
+    ETHInterpreter.ETHTransfer[2] transfers; // [sender, receiver]
```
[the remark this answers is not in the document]
rhlsthrm: From looking at the FreeBalance app, it looked like it needed to be an array of ETHTransfers, can you type out the suggestion for how to fix?

packages/apps/package.json
rhlsthrm: I had an issue that I thought was truffle related so I tried bumping the version. Turned out it wasn't truffle-related. I'll roll this back.

packages/apps/contracts/EthPaymentApp.sol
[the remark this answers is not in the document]
snario: It doesn't _need_ to be, but the return type of `computeOutcome` does need to be. https://github.com/counterfactual/monorepo/pull/1561#discussion_r287837125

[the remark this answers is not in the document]
rhlsthrm: Ah okay, based on that, I think it would probably be better to leave it as is.

packages/apps/contracts/EthPaymentApp.sol
rhlsthrm: https://github.com/counterfactual/monorepo/pull/1561/commits/550bd5da4436ac11ec54c887cdcdeaf74234e340

rhlsthrm: This totally makes sense, I implemented this the way you mentioned: https://github.com/counterfactual/monorepo/pull/1561/commits/550bd5da4436ac11ec54c887cdcdeaf74234e340.

However, thinking about it more, we might not want the limitation of having to wait for a signature to be able to finalize the state on a unidirectional payment. I'll leave as is and we can revisit this.

PR #1563, opened 2019-05-25
Fix Dockerfile so it builds correctly

snario:
Previous PR #1555 added a Dockerfile that did not build

### DIFF
--- a/.circleci/images/build/Dockerfile
+++ b/.circleci/images/build/Dockerfile
@@ -5,7 +5,6 @@ RUN add-apt-repository ppa:ethereum/ethereum
 
 RUN apt-get update
 
-# install solc binary
 RUN apt-get install -y solc
 
 # install node 10.x
--- a/.circleci/images/build/Dockerfile
+++ b/.circleci/images/build/Dockerfile
@@ -3,7 +3,6 @@ FROM resinci/npm-x86_64-ubuntu-node10
 # add ethereum ppa
 RUN add-apt-repository ppa:ethereum/ethereum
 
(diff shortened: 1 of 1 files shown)


.circleci/images/build/Dockerfile
```diff
@@ -1,11 +1,18 @@
-FROM node:10.15.3-jessie
+FROM resinci/npm-x86_64-ubuntu-node10
 
-# install solc binary
+# add ethereum ppa
 RUN add-apt-repository ppa:ethereum/ethereum
+
+# update apt repository
```
cf19drofxots:
```suggestion
```

.circleci/images/build/Dockerfile
```diff
@@ -1,11 +1,18 @@
...
 RUN apt-get update
+
+# install solc binary
```
cf19drofxots:
```suggestion
```

.circleci/images/build/Dockerfile
```diff
@@ -1,11 +1,16 @@
...
-# install solc binary
+# add ethereum ppa
 RUN add-apt-repository ppa:ethereum/ethereum
+
 RUN apt-get update
+
 RUN apt-get install -y solc
-RUN node -v & solc --version
+
+# install node 10.x
```
cf19drofxots: You can run all 3 commands in the same `RUN` step to cache the result in one layer instead of 3 if the execution of these 3 is tightly coupled. Obviously not a blocker, just mentioning it.

PR #1572, opened 2019-05-27
Update truffle

snario:
Noticed @rhlsthrm was doing this in #1393


snario:
Yeah, I just wanted to isolate something from that other branch

cf19drofxots:
So I ran migrate from the latest release and the issue seems to have been fixed. This seems fine to merge.

cf19drofxots:
There seems to be an issue with fetching some packages from yarn package registry.

```
7:01:33 PM: [3/5] Fetching packages...
7:02:05 PM: error https://registry.yarnpkg.com/@firebase/firestore/-/firestore-1.3.2.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "Unexpected end of data"
```

cf19drofxots:
Same issue has persisted. @snario thoughts on switching to npm's package registry instead of yarn's?

PR #1573, opened 2019-05-27
Rename AppStatus to ChallengeStatus

snario:
Seems like a more explicit and descriptive name for this variable


packages/contracts/contracts/libs/LibStateChannelApp.sol
```diff
@@ -6,11 +6,11 @@ pragma solidity 0.5.8;
 /// @notice Contains the structures and enums needed for the ChallengeRegistry
 contract LibStateChannelApp {
 
-  // The mode that the App is currently in from POV of the blockchain
-  enum AppStatus {
-    ON,
-    IN_CHALLENGE,
-    OFF
+  // The status of any challenge in the AppRegistry
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `AppRegistry`?

PR #1574, opened 2019-05-27
Rename AppRegistry to ChallengeRegistry

snario:
This is what is actually stored in the registry


snario:
Do we need to re-deploy the contracts? The name change will change the bytecode I think?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
There are a few places in the codebase which hard-code the assumption that the contract name is equal to the name of the key in `NetworkContext`. If they are unequal there would have to be some place in the code to store them.

snario:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll but based on this PR, they would be equal, right?

PR #1602, opened 2019-05-31
Specify node engine in package.json

snario:
In response to this thing I saw

https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
     ]
   },
   "engines": {
-    "yarn": "^1.10"
+    "yarn": "^1.10",
+    "node": "10.15.3"
   },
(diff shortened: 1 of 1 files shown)


snario:
Note that this causes `yarn` to fail if the version of `node` on the host computer is not 10.15.3

```
$ yarn               
yarn install v1.10.1
[1/5] 🔍  Validating package.json...
error @: The engine "node" is incompatible with this module. Expected version "10.15.3". Got "10.5.0"
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

PR #1608, opened 2019-05-31
[cf-wallet] initial implementation

patience-tema-baron:
Adds the cf-wallet.js package, which is a specialized version of cf.js including wallet specific functionality and removing app specific functionality.


packages/apps/contracts/EthUnidirectionalPaymentApp.sol
```diff
@@ -7,6 +7,7 @@ import "@counterfactual/contracts/contracts/interfaces/Interpreter.sol";
 import "@counterfactual/contracts/contracts/interpreters/ETHInterpreter.sol";
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
+
```
joelalejandro: 🔵 This might be an unintended whitespace.

packages/cf-wallet.js/src/app-instance.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.peerDeposit = info.peerDeposit;
+    this.timeout = info.timeout;
+    this.intermediaries = info.intermediaries;
+  }
+
+  /**
+   * Whether this app is virtual i.e. installation was routed through intermediaries
+   */
+  get isVirtual(): boolean {
+    return !!(this.intermediaries && this.intermediaries.length !== 0);
```
joelalejandro: 🔵 I think we can do a better cast:

```suggestion
    return Array.isArray(this.intermediaries) && this.intermediaries.length > 0;
```

packages/cf-wallet.js/test/provider.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+
+    nodeProvider.onMethodRequest(
+      Node.MethodName.GET_FREE_BALANCE_STATE,
+      request => {
+        expect(request.type).toBe(Node.MethodName.GET_FREE_BALANCE_STATE);
+
+        nodeProvider.simulateMessageFromNode({
+          requestId: request.requestId,
+          type: Node.ErrorType.ERROR,
+          data: { errorName: "music_too_loud", message: "Music too loud" }
```
joelalejandro: lol

packages/cf-wallet.js/test/provider.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+          result: {}
+        });
+      });
+      await provider.rejectInstall(TEST_APP_INSTANCE_INFO.id);
+    });
+
+    it("can create a channel between two parties", async () => {
+      expect.assertions(3);
+
+      const transactionHash = "t_hash";
```
joelalejandro: 💡 Might be interesting to use more "realistic" sample data in this case.

packages/cf-wallet.js/test/provider.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+      });
+
+      const response = await provider.createChannel(TEST_OWNERS);
+      expect(response).toEqual(transactionHash);
+    });
+
+    it("can deposit eth to a channel", async () => {
+      expect.assertions(3);
+
+      const multisigAddress = "multisig_address";
```
joelalejandro: 💡 Might be interesting to use more "realistic" sample data in this case.

packages/cf-wallet.js/test/provider.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+        });
+      });
+
+      await provider.deposit(multisigAddress, amount);
+    });
+
+    it("can withdraw eth from a channel", async () => {
+      expect.assertions(3);
+
+      const multisigAddress = "multisig_address";
```
joelalejandro: 💡 Might be interesting to use more "realistic" sample data in this case.

packages/cf-wallet.js/test/provider.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+        });
+      });
+
+      await provider.withdraw(multisigAddress, amount, TEST_OWNERS[0]);
+    });
+
+    it("can query for a channel's freeBalance", async () => {
+      expect.assertions(3);
+
+      const multisigAddress = "multisig_address";
```
joelalejandro: 💡 Might be interesting to use more "realistic" sample data in this case.

packages/playground/src/components/node-listener/node-listener.tsx
```diff
@@ -81,19 +84,13 @@ export class NodeListener {
...
-        } as Node.InstallVirtualParams,
-        requestId: uuid()
-      };
-
-      const installedApp = (await this.node.call(
-        Node.MethodName.INSTALL_VIRTUAL,
-        request
-      )).result as Node.InstallVirtualResult;
+      const appInstanceId = this.currentMessage.data.appInstanceId;
+      const intermediaryIdentifier = this.currentMessage.data.params
```
joelalejandro: 🔵 We can use array destructuring here:

```ts
const { appInstanceId, intermediaries } = this.currentMessage.data;
const [intermediaryIdentifier] = intermediaries;
```

packages/playground/src/data/counterfactual.ts
```diff
@@ -30,12 +30,21 @@ export interface NodeConfig {
 }
 
 export default class CounterfactualNode {
+  private static port: MessagePort;
   private static node: Node;
+  // @ts-ignore
+  private static nodeProvider: cfWallet.NodeProvider;
```
joelalejandro: ❓ Why are we using `@ts-ignore`?

patience-tema-baron: Since we're using the bundled version of cf-wallet, ts doesn't have access to the relevant types (NodeProvider and Provider). It errors here because it doesn't recognize `cfWallet`, which might also be expressed as `window.cfWallet` (which also would error). Are you familiar w/ a way around this?

joelalejandro: Just wanted to confirm that was the case!

PR #1615, opened 2019-06-03
[playground-server] Add build script

cf19drofxots:
Bumps package version and adds build script.

### DIFF
--- a/packages/playground-server/package.json
+++ b/packages/playground-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@counterfactual/playground-server",
-  "version": "0.1.16",
+  "version": "0.1.17",
   "description": "A backend for the Playground app.",
   "author": "Counterfactual",
   "homepage": "https://github.com/counterfactual/monorepo",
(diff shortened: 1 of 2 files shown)


cf19drofxots:
We don't _need_ to build it, but building it prevents errors from being caught too late if `tsc` fails in the heroku deployment phase.

joelalejandro:
I'm ok with adding this as long as it's restricted to deployment builds (we shouldn't build locally).

snario:
Yeah like @joelalejandro said, this is only needed on Heroku, it seems to me like it should be in `heroku-postbuild`.

cf19drofxots:
> Yeah like @joelalejandro said, this is only needed on Heroku, it seems to me like it should be in heroku-postbuild.

That's exactly the point of this PR though. The postbuild already invokes `tsc` but if it ends up failing, it'll be too late and incur another cycle to fix.

> (we shouldn't build locally).

@joelalejandro the start script invokes `ts-node` (which isn't bothered by any built files, if I understand correctly), so what's the downside of having a build script that invokes `tsc` if it doesn't effect anything else?

cf19drofxots:
This isn't a high-enough priority item to spend time on atm, closing.

PR #1618, opened 2019-06-04
[postgres-connector] Add proper build and testing around package

cf19drofxots:
This ensures the postgres package can be built and consumed as expected, as well as introducing tests to verify expected functionality by the implementation.

This also achieves a couple other things:
- uses in-memory store and messaging services for the Node tests instead of Firebase
  - follow up PRs can fully eliminate the need for Firebase across the stack
- exposes another method on the Node for retrieving a proposed `AppInstance` (primarily for easier testing purposes)


cf19drofxots:
To confirm that the postgres implementation actually works, this PR adds a test to the node package consuming the postgres package.

This is currently failing because the relational APIs don't offer value lookup based on partial path keys.

.circleci/config.yml
```diff
@@ -7,6 +7,15 @@ defaults: &defaults
   working_directory: /home/circleci/project
   docker:
     - image: circleci/node:10.15.3
+      environment:
```
snario: Why are you adding all of these to the `defaults`?

packages/node/.env.defaults
```diff
@@ -4,3 +4,9 @@ FIREBASE_MESSAGING_SERVER_KEY="messageKey"
 FIREBASE_STORE_SERVER_KEY="storeKey"
 FIREBASE_STORE_PREFIX_KEY="dev-store"
 GANACHE_PORT="8545"
+POSTGRES_USER="postgres"
```
snario: It feels a bit odd that we have default environment variables for the `node` package that have to do with plugins being used by it (presumably in tests?)

packages/node/package.json
```diff
@@ -16,12 +16,12 @@
   "devDependencies": {
     "@counterfactual/apps": "0.1.1",
     "@counterfactual/firebase-server": "0.0.1",
+    "@counterfactual/postgresql-node-connector": "0.0.2",
```
snario: Seems odd that the Node has a dev dependency on a plugin to it.

Maybe the plugin should have a dev dependency on the node for testing purposes instead.

packages/node/test/integration/get-app-instance.spec.ts
```diff
@@ -4,7 +4,7 @@ import { Node as NodeTypes } from "@counterfactual/types";
 
 import { Node } from "../../src";
 
-import { setup } from "./setup";
+import { setupWithFirebaseServiceFactory } from "./setup";
```
snario: On second thought, why are we using Firebase for local Node testing? Why can't we just use in-memory messaging and storage?

packages/node/test/integration/setup.ts
```diff
@@ -14,7 +16,33 @@ import { MNEMONIC_PATH, Node } from "../../src";
 import { CF_PATH } from "../global-setup.jest";
 import { A_MNEMONIC, B_MNEMONIC } from "../test-constants.jest";
 
-export async function setup(
+export async function setupWithFirebaseServiceFactory(
```
snario: In general the concept of "service factories" feels like it shouldn't exist. It just _happens_ to be the case that Firebase can be used in a hacky way to do messaging, but in general messaging and storage are fundamentally different things and trying to introduce this kind of "service factory" wrapper concept feels like way too much overhead for this weird edge case.

packages/node/test/integration/setup.ts
```diff
@@ -26,6 +54,37 @@ export async function setup(
   const messagingService = firebaseServiceFactory.createMessagingService(
     process.env.FIREBASE_MESSAGING_SERVER_KEY!
   );
+
+  const postgresServiceFactory = new PostgresServiceFactory({
```
snario: As evidenced here. Why do we have a "PostgresServiceFactory"? We just need a "PostgresStoreService".

.circleci/config.yml
cf19drofxots: Just to get it working. This is still in draft.

packages/node/package.json
cf19drofxots: That was my initial thought and is probably the better way to do it. Only reason I opted for this in the immediate future is because the main objective was to confirm the postgres implementation works as expected, but it's also because setting up proper e2e tests that use in the Node in the postgres package requires a lot of context

packages/node/test/integration/get-app-instance.spec.ts
cf19drofxots: Yeah we could. I'll remove this.

packages/node/test/integration/setup.ts
rhlsthrm: This was my fault, I tried to abide by the given interfaces as much as possible, the factory did seem unnecessary to me then as well. I will remove this.

rhlsthrm: Or it looks like you still might be deep into this @cf19drofxots, let me know if you want me to do any of this.

packages/node/test/integration/setup.ts
cf19drofxots: The nice property of a service factory is that it enables much easier testing as we'd need multiple instances of a service (this is only true for the store service, not the messaging service).

packages/node/test/integration/setup.ts
cf19drofxots: As said in the other comment, this gives more flexibility in creating services, as compared to creating multiple services separately which would amount to effectively the same thing as having a factory producing the services.

cf19drofxots:
So this can be merged as is, though I'm currently working on separating out the ganache environment dependency from the Node package so that the proper context can be injected into the Node _in another_ package, which is the motivation for moving the postgres tests outside of the Node package and into the postgres package.

/cc @snario

.circleci/config.yml
```diff
@@ -52,12 +52,25 @@ jobs:
             - packages/*/build
 
   run-non-playground-tests:
-    <<: *defaults
+    docker:
+      - image: circleci/node:10.15.3
```
snario: Use `circleci-environment` image

.circleci/config.yml
```diff
@@ -52,12 +52,25 @@ jobs:
...
+        environment:
+          POSTGRES_HOST: 127.0.0.1
+          POSTGRES_USER: postgres
+          POSTGRES_DATABASE: postgres
+          POSTGRES_PORT: 5432
+
+      - image: circleci/postgres:9.6.2-alpine
```
snario: Why are there two images?

cf19drofxots: The second image runs in the background and provides a Postgres server.

PR #1632, opened 2019-06-05
Remove address[2] playerAddrs from client HighRollerApp data structure

snario:
Similar to #1629


cf19drofxots:
@snario you've tested this on the playground?

snario:
Yes @cf19drofxots

PR #1636, opened 2019-06-05
Move external method atop internal methods in one file

snario:
This gives a Solidity linting warning


cf19drofxots:
Is the point of this PR to expose this warning?

snario:
It fixes it.

PR #1637, opened 2019-06-05
Add better error messages in install protocol

snario:
Better error messages.


packages/node/src/protocol/install.ts
```diff
@@ -120,7 +131,9 @@ async function proposeStateTransition(
       break;
     }
     default: {
-      throw Error("unrecognized");
+      throw Error(
+        "The outcome type in this application logic contract is not supported yet."
```
cf19drofxots: Does it not suffice to say `application contract`?

PR #1639, opened 2019-06-06
Remove variable name appId, replace w/ appDefinition

snario:
`appId` is an outdated name. `appDefinition` is the more clear and commonly used term to refer to the address of a contract.


ebryn:
This is going to conflict with @joelalejandro's RPC branch

PR #1645, opened 2019-06-07
Add limitOrTotal and beneficiaries to client API wrapper class (also rename them to be explicit)

snario:
These fields were not added in the Interpreter PR #1263 (cc: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll for reference) to the clientside API wrapper class for an `AppInstance` inside `cf.js` and `cf-wallet.js`. This adds them.

**Field rename**
- `limitOrTotal` was `amount` if TwoPartyOutcome interpreter type or `limit` if ETHTransfer interpreter type.
- `beneficiaries` was `playerAddrs` if TwoPartyOutcome else meaningless (but set to `[0x0, 0x0]`)

### DIFF
--- a/packages/cf-wallet.js/src/app-instance.ts
+++ b/packages/cf-wallet.js/src/app-instance.ts
@@ -4,6 +4,10 @@ import {
   AppInstanceID,
   AppInstanceInfo
 } from "@counterfactual/types";
+import {
+  ETHTransferInterpreterParams,
+  TwoPartyOutcomeInterpreterParams
+} from "@counterfactual/types/dist/src/data-types";
 import { BigNumber } from "ethers/utils";
 
 import { Provider } from "./provider";
@@ -26,18 +30,8 @@ export class AppInstance {
   readonly myDeposit: BigNumber;
   readonly peerDeposit: BigNumber;
 
-  readonly twoPartyOutcomeInterpreterParams?: {
-    // Derived from:
-    // packages/contracts/contracts/interpreters/TwoPartyEthAsLump.sol#L10
-    playerAddrs: [string, string];
-    amount: BigNumber;
-  };
-
-  readonly ethTransferInterpreterParams?: {
-    // Derived from:
-    // packages/contracts/contracts/interpreters/ETHInterpreter.sol#L18
-    limit: BigNumber;
-  };
+  readonly twoPartyOutcomeInterpreterParams?: TwoPartyOutcomeInterpreterParams;
+  readonly ethTransferInterpreterParams?: ETHTransferInterpreterParams;
 
   readonly intermediaries?: Address[];
 
--- a/packages/cf-wallet.js/src/app-instance.ts
+++ b/packages/cf-wallet.js/src/app-instance.ts
@@ -25,8 +25,20 @@ export class AppInstance {
   // Funding-related fields
   readonly myDeposit: BigNumber;
   readonly peerDeposit: BigNumber;
-  readonly beneficiaries?: string[];
-  readonly limitOrTotal?: BigNumber;
+
+  readonly twoPartyOutcomeInterpreterParams?: {
+    // Derived from:
+    // packages/contracts/contracts/interpreters/TwoPartyEthAsLump.sol#L10
+    playerAddrs: [string, string];
+    amount: BigNumber;
+  };
+
+  readonly ethTransferInterpreterParams?: {
+    // Derived from:
+    // packages/contracts/contracts/interpreters/ETHInterpreter.sol#L18
+    limit: BigNumber;
+  };
+
   readonly intermediaries?: Address[];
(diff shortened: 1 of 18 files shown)


packages/cf-wallet.js/src/app-instance.ts
```diff
@@ -17,11 +17,16 @@ export class AppInstance {
...
+  // Application-specific fields
   readonly appDefinition: Address;
   readonly abiEncodings: AppABIEncodings;
+  readonly timeout: BigNumber;
+
+  // Funding-related fields
   readonly myDeposit: BigNumber;
   readonly peerDeposit: BigNumber;
-  readonly timeout: BigNumber;
+  readonly beneficiaries?: string[];
```
cf19drofxots: Why optional?

snario: If `ETH_TRANSFER` is the interpreter type, then `beneficiaries` is `[AddressZero, AddressZero]` which is meaningless data. If `TWO_PARTY_OUTCOME` then it has correct data.

cf19drofxots: Hmm right. Just seems a bit awkward to be required to know interpreter-level behavior to specify parameters this high level.

snario: Yeah. I have another PR coming to make it more explicit.

snario:
Renamed all of the fields in the codebase also to much more clearly labelled variables.

packages/cf-wallet.js/src/app-instance.ts
```diff
@@ -17,11 +17,28 @@ export class AppInstance {
...
+
+  // Funding-related fields
   readonly myDeposit: BigNumber;
   readonly peerDeposit: BigNumber;
-  readonly timeout: BigNumber;
+
+  readonly twoPartyOutcomeInterpreterParams?: {
+    // Derived from:
+    // packages/contracts/contracts/interpreters/TwoPartyEthAsLump.sol#L10
+    playerAddrs: [string, string];
```
cf19drofxots: It's kind of funny how we still don't have one definitive word for parties of a channel (thought of this because not all apps will be games i.e. have a notion of `players`)

packages/cf-wallet.js/src/app-instance.ts
```diff
@@ -17,11 +17,28 @@ export class AppInstance {
...
   readonly appDefinition: Address;
   readonly abiEncodings: AppABIEncodings;
+  readonly timeout: BigNumber;
+
+  // Funding-related fields
   readonly myDeposit: BigNumber;
   readonly peerDeposit: BigNumber;
-  readonly timeout: BigNumber;
+
+  readonly twoPartyOutcomeInterpreterParams?: {
```
cf19drofxots: This seems like it's repeated multiple times. Maybe define it in the `types` package?

(Same with `ethTransferInterpreterParams`)

packages/types/src/data-types.ts
```diff
@@ -13,6 +13,22 @@ export type AppInstanceInfo = {
   proposedByIdentifier: string; // xpub
   proposedToIdentifier: string; // xpub
   intermediaries?: string[];
+
```
cf19drofxots: So they _are_ defined here. Why not use them across the other packages then?

packages/cf-wallet.js/src/app-instance.ts
snario: This is different, though, this is a custom parameter for the `TwoPartyOutcome` interpreter where the outcomes are basically WIN_A, WIN_B or DRAW, so "player" makes more sense. The term we agreed on at the channel level is "participant". See: https://github.com/statechannels/SCIPs/blob/master/TERMINOLOGY.md#participant

packages/cf-wallet.js/src/app-instance.ts
snario: I was hesitant to do this because I want to remove this code in a future PR so it is a bit more general. Right now it is explicitly ugly because this _is_ an ugly field to add to this object.

packages/types/src/data-types.ts
snario: This is the data type for the client side object. I suppose it could be getting re-used.

packages/cf-wallet.js/src/app-instance.ts
cf19drofxots: I share the hesitation, just think keeping things consistent even in the interim is generally good until this future PR does happen which can refactor things more generally then.

snario: Added types.

packages/types/src/data-types.ts
snario: Added.

PR #1627, opened 2019-06-05
Add yarn.lock so that Heroku environments are consistent

snario:
The bots were not running because of the following error:

```
2019-06-05T15:14:35.304306+00:00 app[worker.1]: Error: Cannot find module '@firebase/app'
2019-06-05T15:14:35.304309+00:00 app[worker.1]:     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
2019-06-05T15:14:35.304311+00:00 app[worker.1]:     at Function.Module._load (internal/modules/cjs/loader.js:562:25)
2019-06-05T15:14:35.304313+00:00 app[worker.1]:     at Module.require (internal/modules/cjs/loader.js:690:17)
2019-06-05T15:14:35.304315+00:00 app[worker.1]:     at require (internal/modules/cjs/helpers.js:25:18)
2019-06-05T15:14:35.304316+00:00 app[worker.1]:     at /app/node_modules/@firebase/auth/dist/auth.js:1:29
2019-06-05T15:14:35.304319+00:00 app[worker.1]:     at Object.<anonymous> (/app/node_modules/@firebase/auth/dist/auth.js:361:432)
2019-06-05T15:14:35.304320+00:00 app[worker.1]:     at Module._compile (internal/modules/cjs/loader.js:776:30)
2019-06-05T15:14:35.304322+00:00 app[worker.1]:     at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
2019-06-05T15:14:35.304324+00:00 app[worker.1]:     at Module.load (internal/modules/cjs/loader.js:653:32)
2019-06-05T15:14:35.304326+00:00 app[worker.1]:     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
2019-06-05T15:14:35.319884+00:00 app[worker.1]: npm ERR! code ELIFECYCLE
2019-06-05T15:14:35.321102+00:00 app[worker.1]: npm ERR! errno 1
2019-06-05T15:14:35.327467+00:00 app[worker.1]: npm ERR! @counterfactual/tic-tac-toe-bot@0.0.15 serve: `node dist`
2019-06-05T15:14:35.327709+00:00 app[worker.1]: npm ERR! Exit status 1
2019-06-05T15:14:35.328090+00:00 app[worker.1]: npm ERR! 
2019-06-05T15:14:35.328320+00:00 app[worker.1]: npm ERR! Failed at the @counterfactual/tic-tac-toe-bot@0.0.15 serve script.
[...]

PR #1646, opened 2019-06-07
[monorepo] Update lock file

cf19drofxots:
[Last merge](https://github.com/counterfactual/monorepo/pull/1618) failed at the heroku deployment phase due to outdated lock file. 

```
remote: -----> Installing dependencies        
remote:        Installing node modules (yarn.lock)        
remote:        yarn install v1.12.3        
remote:        [1/4] Resolving packages...        
remote:        error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.        
remote:        info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.        
remote: 
remote: -----> Build failed        
remote: 
remote:  !     Outdated Yarn lockfile        
```

### DIFF
--- a/packages/high-roller-bot/yarn.lock
+++ b/packages/high-roller-bot/yarn.lock
@@ -154,10 +154,10 @@
   resolved "https://registry.yarnpkg.com/@counterfactual/contracts/-/contracts-0.1.1.tgz#f2045b281b0ab11b99417dc978dcf458aeba221c"
   integrity sha512-RyPxbzBfdhbmPjYLYHviB4aCdN0mqgcH63d0d9OY6TH0h2wThzLi05+elBsXnK6KekKg4cUWwyFtGPvbMHY5SQ==
 
-"@counterfactual/firebase-client@0.0.2":
-  version "0.0.2"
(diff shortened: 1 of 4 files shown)


snario:
This is probably my fault because I added `yarn.lock`s for each sub package on Heroku to fix their dependencies and we don't have a way of updating these locally easily. See #1627

snario:
I don't think updating the root `yarn.lock` will do anything

cf19drofxots:
@snario so what do you suggest as a fix here then?

snario:
We need to update each individual `yarn.lock`. Unfortunately the easiest way to do this is to:

1. Comment out these lines:

https://github.com/counterfactual/monorepo/blob/07a118bd6f18be441adff4f56724a2d152b3e366/package.json#L3-L7

2. `cd` into `playground-server`, `tic-tac-toe-bot`, and `high-roller-bot` and for each run `yarn`

3. Uncomment those lines

4. Commit the updates `yarn.lock` files

cf19drofxots:
Updated according to instructions

PR #1666, opened 2019-06-10
Proposal (w/ PR): Rename playground-server to simple-hub-server

snario:
I find there is still some confusion with the term "playground". For example, when we migrate the demo apps to run inside of MetaMask and not inside the `playground` environment, you might think we are "deprecating the Playground" however in truth we would only deprecate the `playground` package and the `playground-server` would probably remain running.

Anyway this is just a suggestion that took 2 min to implement the PR for.

Requires `yarn publish` of `@counterfactual/simple-hub-server` if we decide to do it.


packages/playground/README.md
```diff
@@ -19,7 +19,7 @@ In order for the Playground to run at your workstation, you'll need:
 - A PostgreSQL database (>= 10.x)
 - A BIP39 mnemonic for the Server's funds account
 
-Check on the [Playground Server requirements](../playground-server/README.md#usage) **before** running the Playground.
+Check on the [Playground Server requirements](../simple-hub-server/README.md#usage) **before** running the Playground.
```
joelalejandro: @snario Shouldn't we rename this to `Hub Server`?

cf19drofxots:
```suggestion
Check on the [Hub Server requirements](../simple-hub-server/README.md#usage) **before** running the Playground.
```

packages/playground/README.md
```diff
@@ -32,7 +32,7 @@ yarn run:playground
 This command will run:
 
 - The Playground app itself (this package)
-- The [Playground Server](../playground-server)
+- The [Playground Server](../simple-hub-server)
```
cf19drofxots:
```suggestion
- The [Hub Server](../simple-hub-server)
```

PR #1669, opened 2019-06-11
Adds turnNum to all existing apps

snario:
Had to do some big changes to the `HighRoller` contract cc: @Alonski 

Does not handle the ETHUnidirectional app @rhlsthrm wrote, yet.

Have not been able to test Playground, for some reason the `playground:matchmakeWith` flag isn't working for me locally and I haven't tried with bots, yet. Also I did not update the bot code for HR yet.

### DIFF
--- a/packages/dapp-high-roller/src/components/app-game/app-game.tsx
+++ b/packages/dapp-high-roller/src/components/app-game/app-game.tsx
@@ -5,9 +5,9 @@ import { RouterHistory } from "@stencil/router";
 
 import CounterfactualTunnel from "../../data/counterfactual";
 import {
-  Action,
-  ActionType,
   GameState,
+  HighRollerAction,
+  HighRollerActionType,
   HighRollerAppState,
   HighRollerStage,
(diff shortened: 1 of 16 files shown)


packages/dapp-high-roller/src/components/app-game/app-game.tsx
```diff
@@ -134,30 +135,26 @@ export class AppGame {
...
-        actionType: ActionType.COMMIT_TO_HASH,
-        actionHash: hash
+      const commitHashAction: HighRollerAction = {
+        actionType: HighRollerActionType.COMMIT_TO_HASH,
+        actionHash: hash,
+        // NOTE: `number` is unused with COMMIT_TO_HASH but cannot be undefined
+        number: 0
       };
 
+      // TODO: Figure out why this is using ...
```
Alonski: This is just spreading out the response from `takeAction` so the properties on the `HighRollerAppState` object get applied as is to `this.highRollerState`

PR #1670, opened 2019-06-11
add typedocs to cf and cf-wallet

patience-tema-baron:
Adds typedoc to the build script for cf.js and cf-wallet.js.

Currently the docs are themed to use markdown, that way they can be viewed [on github](https://github.com/counterfactual/monorepo/tree/519b3872b7c09803775e815b449c097923f667c1/packages/cf-wallet.js/docs). In order to view them on github, we obviously can't gitignore them, even though that might otherwise be preferable.

Alternatively, we could theme them as html (though we'll need to find someplace else to publish them to):

Screenshot from 2019-06-11 15-06-05


package.json
```diff
@@ -35,6 +35,8 @@
     "tslint-config-airbnb": "5.11.1",
     "tslint-config-prettier": "1.18.0",
     "tslint-plugin-prettier": "2.0.1",
+    "typedoc": "^0.14.2",
```
snario: Do we need this in the root `package.json`?

packages/cf-wallet.js/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "sourceRoot": ".",
```
snario: Why is this added?

packages/cf-wallet.js/package.json
```diff
@@ -15,11 +15,11 @@
   "repository": "https://github.com/counterfactual/monorepo/tree/master/packages/cf.js",
   "license": "MIT",
   "scripts": {
-    "build": "tsc -b . && rollup -c",
+    "build": "tsc -b . && rollup -c && yarn docs",
```
snario: Maybe have a separate `docgen` script?

package.json
patience-tema-baron: Is the goal to eventually use typedocs throughout the project? If so, it makes sense here, right? If no, I could just move it into the relevant projects.

packages/cf-wallet.js/tsconfig.json
patience-tema-baron: For parity w/ cf.js. I'd removed it from cf-wallet.js when I wrote it initially, since it was leading to a build error. This pr includes a fix for that build error, and so I could bring this line back. That being said, cf-wallet.js seems to work fine without it. Same with cf.js. Perhaps we could just remove it from both.

packages/cf-wallet.js/package.json
patience-tema-baron: We currently have a `docs` script which works independently of the other build commands. I'm just calling it from here to ensure that the docs get regenerated whenever we make a change to these packages. I could remove it, but I'm worried about us failing to regenerate the docs if we have to do so manually.

package.json
snario: Always makes sense to use it in the relevant projects

packages/cf-wallet.js/tsconfig.json
snario: Why do we need it?

patience-tema-baron: As far as I can tell, we don't use it at present because we're not actually using the sourcemaps we generate, at least not within the context of the playground. That being said, even if we were using sourcemaps, it looks like it's just the tsconfig default value. I'm not sure why we added it in the first place, but it's probably safe to remove.

patience-tema-baron: Went ahead and removed this line just to unblock things.

PR #1671, opened 2019-06-12
[node] ERC20 testing environment

cf19drofxots:
This sets up the testing environment for ensuring an ERC20 token is minted and supply is as expected.

This is task 0 in verifying the framework's functioning support with ERC20 token usage.


snario:
Why do we need `DolphinCoin` on testnet?

cf19drofxots:
We don't. Right now the tests we have in place check for whether the contracts listed in the `NetworkContext` have been deployed or not, and this just makes things more consistent across all contracts instead of having to have conditionals on whether a contract should be deployed or not (we could just have an `if` condition in the `check-deployed-contracts` test instead of deploying `DolphinCoin` - I'm largely indifferent)

cf19drofxots:
@snario removed the deployments and put in the check I described above.

snario:
@cf19drofxots on `master` I still see `DolphinCoin` inside `ARTIFACTS`:

https://github.com/counterfactual/monorepo/blob/master/packages/contracts/migrations/3_deploy_contracts.js#L17-L21

PR #1672, opened 2019-06-12
postgresql-node-connector: Fix hard-coded username

bohendo:
### Description

The `node_records` table creation query had a line that sets the owner to a hard-coded "postgres".

I replaced "postgres" with the username provided by the configuration.

### Related issues

none

^ idk what this means but this change fixed a bug while deploying `ConnextProject/indra-v2` so there's that

### DIFF
--- a/packages/postgresql-node-connector/src/index.ts
+++ b/packages/postgresql-node-connector/src/index.ts
@@ -14,13 +14,13 @@ export const POSTGRES_CONFIGURATION_ENV_KEYS = {
   port: "POSTGRES_PORT"
 };
 
-export interface PostgresConnectionOptions {
+export type PostgresConnectionOptions = ConnectionOptions & {
   username: string;
   host: string;
   database: string;
   password: string;
   port: number;
-}
+};
 
 export const EMPTY_POSTGRES_CONFIG: ConnectionOptions = {
   type: "postgres",
@@ -36,7 +36,7 @@ export class PostgresServiceFactory implements Node.ServiceFactory {
   private connection: Connection;
 
   constructor(
-    readonly configuration: ConnectionOptions | PostgresConnectionOptions,
+    readonly configuration: PostgresConnectionOptions,
     readonly tableName: string = "node_records"
   ) {
     this.connectionManager = new ConnectionManager();
--- a/packages/postgresql-node-connector/src/index.ts
+++ b/packages/postgresql-node-connector/src/index.ts
@@ -36,7 +36,7 @@ export class PostgresServiceFactory implements Node.ServiceFactory {
   private connection: Connection;
 
   constructor(
-    configuration: ConnectionOptions,
+    readonly configuration: PostgresConnectionOptions,
     readonly tableName: string = "node_records"
   ) {
     this.connectionManager = new ConnectionManager();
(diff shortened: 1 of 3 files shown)


bohendo:
I assume you mean the version in `packages/postgresql-node-connector/package.json`, I'll push that real quick

bohendo:
Ohh, this is weird: https://circleci.com/gh/counterfactual/monorepo/32815?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I was getting errors that `ConnectionOptions` (from typeorm) doesn't have the username property so I changed it to `PostgresConnectionOptions` & my build passed/. But now CI says that one doesn't have a type property 🤔

cf19drofxots:
You can probably convert `PostgresConnectionOptions` to a type then union it with the `ConnectionOptions` type.

bohendo:
I tried to update `PostgresConnectionOptions` to extend `ConnectionOptions` and that caused all kinda of stuff to blow up, is that the same as doing a type union?

I added an optional `type?: string` to `PostgresConnectionOptions`.. I'd like to test it but can't figure out how these tests work. `npm test` spits out a bunch of stuff like:
```
    test/save-load.spec.ts:5:1 - error TS2593: Cannot find name 'describe'. Do you need to install type definitions for a test runner? Try `npm i @types/jest` or `npm i @types/mocha` and then add `jest` or `mocha` to the types field in your tsconfig.
```

bohendo:
nvmd my build was only working because I edited tsconfig to not include tests.. Idk how these tests are supposed to compile

cf19drofxots:
I just locally tried doing a type union and the build passed fine.

Adding specific, optional parameters is probably not ideal because of the parameters that `ConnectionOptions` has that's expected to be passed in, hence my suggestion of the type union. And because `ConnectionOptions` itself is a type, `PostgresConnectionOptions` being an interface can't extend it.

bohendo:
Cool, let's give the union a shot

packages/postgresql-node-connector/src/index.ts
```diff
@@ -36,7 +36,7 @@ export class PostgresServiceFactory implements Node.ServiceFactory {
   private connection: Connection;
 
   constructor(
-    configuration: ConnectionOptions,
+    readonly configuration: ConnectionOptions | PostgresServiceFactory,
```
bohendo: Is this the kind of union you were talking about @cf19drofxots ?

bohendo:
Hmm, no good: https://circleci.com/gh/counterfactual/monorepo/32838?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Is the username even supposed to be supplied via this configuration object?

packages/postgresql-node-connector/src/index.ts
bohendo: Oh wow, autocomplete got me. Let me fix this lol

cf19drofxots: Ah not exactly. More like, for the actual `PostgresConnectionOptions`, converting the interface to a type, like:
```
export type PostgresConnectionOptions = ConnectionOptions | { <the-current-properties> };
```

cf19drofxots: And then passing this updated `PostgresConnectionOptions` to this constructor.

bohendo:
Alright, fixed the types this morning now that my brain is a little fresher. Think this PR should be good to go for real now.

PR #1680, opened 2019-06-14
Some fixes for Playground

snario:
- Some contract addresses were out of date
- Some missing parameters in uninstall virtual protocol messaging
- Logic for checking if you were p1 or p2 was quite convoluted in HR; this does not fix it but it makes it work in the case of playing against the bot. Future PR can use `interpreterParams` to do it deterministically based on the `appInstance`
- Bumped node version
- Install virtual app protocol used `interpreterParams` incorrectly; also as a note it it hard coded for the two party outcome interpreter for all virtual apps atm

Verified HR works as expected against bot on PG.


packages/node/src/protocol/install-virtual-app.ts
```diff
@@ -232,7 +232,7 @@ function createAndAddTarget(
     0, // app nonce
     defaultTimeout,
     {
-      playerAddrs: [initiatingAddress, intermediaryAddress],
+      playerAddrs: [initiatingAddress, respondingAddress],
```
snario: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll can you ack this

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: LGTM

snario:
Blocked on being able to update the sub `yarn.lock` files for Heroku because of an issue running `yarn` on `playground-server` because of `prototypal/jsonrpc-server` throwing this error:

```
[5/5] 📃  Building fresh packages...
[1/9] ⠂ protobufjs
[2/9] ⠂ grpc
[3/9] ⠂ rpc-server
[4/9] ⠂ sha3
error /Users/liam/src/code/monorepo/packages/playground-server/node_modules/rpc-server: Command failed.
Exit code: 127
Command: rm -rf dist/ && tsc -b
Arguments: 
```

cc @joelalejandro

packages/node/src/protocol/uninstall-virtual-app.ts
```diff
@@ -114,6 +114,7 @@ export const UNINSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
       Opcode.IO_SEND_AND_WAIT,
       {
         // m4
+        ...context.message,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why is this here?

snario: Because when `...context.message` was removed last time it did not correctly add in every explicit parameter it was responsible for adding; this lead to `undefined` protocol messages being passed around.

snario:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll Most other files use `...context.message`. At the moment this fixes a bug and so I think in another PR we can be explicit everywhere.

packages/node/src/protocol/install-virtual-app.ts
```diff
@@ -51,12 +52,14 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
...
-        ...context.message,
-        signature: s1,
-        signature2: s5,
+        protocol: Protocol.InstallVirtualApp,
+        protocolExecutionID: context.message.protocolExecutionID,
         toXpub: intermediaryXpub,
-        seq: 1
-      }
+        seq: 1,
+        params: context.message.params,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why does this message need to pass params?

packages/node/src/protocol/install-virtual-app.ts
```diff
@@ -100,12 +103,14 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
     const m3 = yield [
       Opcode.IO_SEND_AND_WAIT,
       {
-        ...context.message,
+        protocol: Protocol.InstallVirtualApp,
+        protocolExecutionID: context.message.protocolExecutionID,
+        params: context.message.params,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why does this message need to pass params?

packages/node/src/protocol/install.ts
```diff
@@ -73,11 +80,13 @@ export const INSTALL_PROTOCOL: ProtocolExecutionFlow = {
     yield [
       Opcode.IO_SEND,
       {
-        ...context.message,
+        protocol: Protocol.Install,
+        protocolExecutionID: context.message.protocolExecutionID,
+        params: context.message.params,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why does this message need to pass params?

PR #1701, opened 2019-06-17
Remove outcomeType solidity function in favour of InstallParams field

snario:
This PR removes `outcomeType` as a mandatory field on `CounterfactualApp`.

Outstanding issues with the whole thing we will address in future PRs:
- Virtual apps still have a hard-coded requirement of `TwoPartyFixedOutcome`
- `interpreterAddress` and `interpreterParameters` are still hard-coded based on `outcomeType` in the `InstallParams`, but ought to be in the `InstallParams`

### DIFF
--- a/packages/apps/contracts/ETHUnidirectionalTransferApp.sol
+++ b/packages/apps/contracts/ETHUnidirectionalTransferApp.sol
@@ -76,14 +76,6 @@ contract ETHUnidirectionalTransferApp is CounterfactualApp {
     return appState.finalized;
   }
 
-  function outcomeType()
-    external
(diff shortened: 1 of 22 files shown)


snario:
> How does this handle the case where the incorrect outcomeType is specified? (i.e. previously the contract said it should be X, now this is simply trusting the call on the client?)

Yes, the caller is trusted

cf19drofxots:
I mean that's probably fine but what I'm trying to say is that there's no existing mechanism to mitigate against an incorrectly-specified outcome type. Is this mitigation part of a future PR?

PR #1702, opened 2019-06-17
Add teardown function for PostgreSQL table in tests

snario:


packages/node/test/integration/setup.ts
```diff
@@ -52,6 +54,21 @@ export async function setupWithMemoryMessagingAndPostgresStore(
   );
 }
 
+export async function teardownPostgresStore() {
```
cf19drofxots: We have a [`reset`](https://github.com/counterfactual/monorepo/blob/master/packages/types/src/node.ts#L46) function on the store interface that we could implement [in the postgres package](https://github.com/counterfactual/monorepo/blob/master/packages/postgresql-node-connector/src/index.ts#L77) instead of here.

snario: Can that be done separately to this PR? Currently this fixes an annoying bug as-is with small lines of code

PR #1707, opened 2019-06-18
Remove DolphinCoin in ARTIFACTS

snario:


packages/contracts/migrations/3_deploy_contracts.js
```diff
@@ -16,7 +16,6 @@ const TwoPartyEthAsLump = artifacts.require("TwoPartyEthAsLump");
 
 const ARTIFACTS = [
   ChallengeRegistry,
-  DolphinCoin
```
cf19drofxots: Will need to remove the `require` above as well.

PR #1710, opened 2019-06-18
Remove custom ethers from typescript-typings

snario:
We were maintaining a custom typings file for `ethers` instead of using the one that (now) ships with the npm package. As a result of removing it, the `playground` source code now uses `window["ethers"]` when referring to an `ethers` library, which is closer to the truth of the matter anyway. Previously it was re-using the fact that `ethers` happened to be a global symbol.


cf19drofxots:
Lint fix

PR #1719, opened 2019-06-19
Add GetStateDepositHolderAddress method on Node

snario:


snario:
@cf19drofxots I'm thinking of bumping it in bulk with a couple other changes

cf19drofxots:
I generally lean towards bumping it with every incremental change to keep consistent with the notion of the package code actually changing, but effectively speaking, I suppose I digress.

cf19drofxots:
Note that this feature is not in any published version yet.

Waiting on @snario to make a PR to update the package version for this method to be exposed.

cf19drofxots:
https://github.com/counterfactual/monorepo/pull/1722

PR #1720, opened 2019-06-19
Make free balance field in StateChannel 1 AppInstance

snario:


packages/node/src/models/state-channel.ts
```diff
@@ -207,22 +197,18 @@ export class StateChannel {
...
-  public getETHFreeBalance(): AppInstance {
-    if (!this.freeBalanceAppIndexes.has(0)) {
-      throw Error(ERRORS.FREE_BALANCE_MISSING);
-    }
-
-    const idx = this.freeBalanceAppIndexes.get(0);
-
-    if (idx === undefined || !this.appInstances.has(idx)) {
-      throw Error(ERRORS.FREE_BALANCE_IDX_CORRUPT(idx!));
+  public getFreeBalanceAppInstance(): AppInstance {
```
cf19drofxots: Make this a getter so other internal retrievals would throw this error (instead of doing `!`) if no app instance is set for the free balance.

packages/node/src/models/state-channel.ts
```diff
@@ -265,23 +261,17 @@ export class StateChannel {
...
-    const freeBalanceAppIndexes = new Map<0, string>([[0, fb.identityHash]]);
-
     return new StateChannel(
       multisigAddress,
       userNeuteredExtendedKeys,
-      appInstances,
+      new Map<string, AppInstance>([]),
       new Map<string, TwoPartyVirtualEthAsLumpInstance>(),
-      freeBalanceAppIndexes,
+      createETHFreeBalance(
```
cf19drofxots: Probably worth renaming this function to just `createFreeBalance` then.

packages/node/src/models/state-channel.ts
```diff
@@ -295,7 +285,7 @@ export class StateChannel {
       userNeuteredExtendedKeys,
       new Map<string, AppInstance>(),
       new Map<string, TwoPartyVirtualEthAsLumpInstance>(),
-      new Map<0, string>(),
+      undefined,
```
cf19drofxots: An empty channel should still have a free balance with the default asset class being ETH (i.e. just invoke `createFreeBalance` here).

packages/node/src/models/state-channel.ts
```diff
@@ -540,7 +530,9 @@ export class StateChannel {
           return [appInstanceEntry[0], appInstanceEntry[1].toJson()];
         }
       ),
-      freeBalanceAppIndexes: Array.from(this.freeBalanceAppIndexes.entries()),
+      freeBalanceAppInstance: !!this.freeBalanceAppInstance
+        ? this.freeBalanceAppInstance.toJson()
+        : undefined,
```
cf19drofxots: The free balance app instance should never be undefined.

packages/node/src/models/state-channel.ts
snario: Before you run setup protocol though you technically do not have a free balance

packages/node/src/models/state-channel.ts
snario: Before you run setup protocol though you technically do not have a free balance

packages/node/src/models/state-channel.ts
cf19drofxots: Got it - seems like this is the exception then and it should be noted.

PR #1721, opened 2019-06-19
Update prototypal/jsonrpc-server

snario:


joelalejandro:
Not really. We could publish it.

joelalejandro:
@snario There's something wrong with the Node not finding the `rpc-server` package? I'll try publishing, but this feels weird.

joelalejandro:
@snario Can you change `package.json` to point to this reference?

```
"rpc-server": "^0.0.1",
```

cf19drofxots:
Are we sure about wanting `^`?

joelalejandro:
@cf19drofxots If you feel better freezing the dependency, I'm OK with it. The `rpc-server` package is pretty much done for now.

PR #1734, opened 2019-06-21
Add reset method on PostgresStoreService

snario:


packages/postgresql-node-connector/test/save-load.spec.ts
```diff
@@ -22,6 +22,12 @@ describe("Postgres store service implementation behavior adheres to interface",
     );
   });
 
+  afterAll(async () => {
+    if (postgresService.reset !== undefined) {
```
cf19drofxots: In the tests this should always be defined, so if it is in fact undefined we should throw an error.

PR #1735, opened 2019-06-21
Replace nonce with versionNumber

snario:
This pull request does a couple of things:

## 1. Rename `nonce` to `versionNumber` for `AppInstance`
Because the term `nonce` is used in many contexts, I think it is clearer to use `versionNumber`. It is especially useful because we are moving to a fixed turn-based mechanism for state updates.

## 2. Remove `challengeNonce`
We used to have a mechanic where we kept track of both (a) the latest unanimously signed state update (the `nonce`) and (b) the number of actions that have been taken with that state as the base on-chain (the `challengeNonce`). This was because you might get in a situation where `nonce` `k` is put on-chain and then some actions are taken on it (reaching `nonce` `k` with `challengeNonce` `j`) and _then_ someone might come along with `nonce` `k+1`. In this case, the newer nonce trumped _all_ action based state updates.

This has been changed in preparation for a larger change which is that when an action is taken, now, the same variable (i.e., `nonce` in the above example, now renamed to `versionNumber`) is incremented. If you end up in a scenario with `nonce` `k+1` being doubly signed and a state at `k` that was progressed via an action to `k+1` then the on-chain version is considered the latest state.


packages/wallet-ui/package.json
```diff
@@ -20,7 +20,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "CI=true react-scripts test",
```
snario: @joelalejandro had to add this because https://github.com/facebook/create-react-app/issues/6899

cf19drofxots:
lgtm to merge if @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll is also good with this

cf19drofxots:
Yeah it'd make more sense actually to decouple the two references more clearly (use `turnNum` where relevant and keep `nonce` where relevant)

PR #1736, opened 2019-06-21
Refactor instruction executor Context objects

snario:
I noticed this part of the code was rather confusing because of a handful of implied variables that were being passed through abstract objects. This makes things much more explicit so that it is clearer what is going on. One property lost is that the instruction executor no longer returns a map of state channels anymore but returns different things depending on whether the protocol affects (1) a single app instance, (2) an entire state channel or (3) multiple state channels (e.g., with an intermediary and metachannel).


snario:
> to more statically limit what parts of the stored state protocols have accessed to, a generic mechanism should be designed and used, instead of a switch statement

Any suggestions?

The two goals are (a) this one; to limit what state is passed into a protocol and (b) to make explicit distinctions between protocols that modify app instance state, a direct channel, and a virtual channel.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
> I claim that the use case of "writing a new protocol" is very uncommon and shouldn't necessarily be one we design for.

We shouldn't design for it, but designing `instructionExecutor` in such a way that new protocols can be added without modifying it makes it much more decoupled and easier to reason about

PR #1773, opened 2019-06-27
make MAppCaller a library

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Since this piece of code does not use any storage from the challenge registry, it should be a library


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
What is "this"?

PR #1774, opened 2019-06-28
Remove MultiSend, move activeApps to FreeBalanceState, upgrade protocols, and more

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This pull request contains several changes rolled into one. They are:

### Removing `MultiSend`
Previously we used the `MultiSend` contract for the purposes of **atomically updating the free balance state and signing a multisignature wallet delegatecall transaction**. We needed this atomicity because we had a design goal of one-round-trip per protocol execution. We've relaxed this design goal and are okay with the protocols requiring multiple round trips, now.

### Add `activeApps` to the Free Balance state
Instead of using `uninstallKey` parameters on the pre-signed transaction made to an `AppInstance` to validate if it is in fact a funded App or not, we instead rely on a master list of "active" (a.k.a. funded) apps. This is an array on the state of the free balance. You can see how this logic plays out here:

https://github.com/counterfactual/monorepo/blob/ec25b2cb25942d1ed4e5eea8fe658f7ac652c30f/packages/contracts/contracts/ConditionalTransactionDelegateTarget.sol#L73-L86

This means the `UninstallKeyRegistry` is also removed.

### Updating protocols to manage effected changes

The following protocols have been upgraded:

- Install
- InstallVirtual
- Uninstall
- UninstallVirtual

In the case of the installation protocols, they now have an extra round trip per user involved. This is because to make an installation secure you need to _first_ see the unanimously signed conditional transaction (the call to the `ConditionalTransactionDelegateTarget`) and _then_ you feel OK with signing the update to the Free Balance (both `activeApps` and the relevant balance updates).

In the case of the uninstall, instead of setting a bit on the `UninstallKeyRegistry` now we set the state of the Free Balance object and remove the app from the `activeApps` list.

- [x] Still need to update the Mermaid diagrams!
- [x] Need to deploy new contract addresses

### Extend `FreeBalanceState`

https://github.com/counterfactual/monorepo/blob/ec25b2cb25942d1ed4e5eea8fe658f7ac652c30f/packages/node/src/models/free-balance.ts#L36-L41

Now it includes `balancesIndexedByToken` and `activeAppsBloomFilter` whereas previously it was just this:

```typescript
export type FreeBalanceState = {
  [tokenAddress: string]: PartyBalance[];
}
```

### Introduce `ConditionalTransaction`
This is a new commitment type which represents a multisignature wallet delegatecall function call to the `ConditionalTransactionDelegateTarget`. Previously this was encoded inside the "install commitment". See here:

Capture d’écran 2019-07-07 à 18 32 15

### Remove `TwoPartyVirtualEthAsLump.sol`

We had a special case contract that handled virtual app agreement interpretation in a hard-coded way for `TwoPartyFixedOutcome` outcome types. Instead, now we re-use `ConditionalTransactionDelegateTarget` as we do for all AppInstances but the clientside code is now hard-coded to the `TwoPartyFixedOutcomeETHInterpreter` for now instead.

### Introduce `IdentityApp`
For test scenarios this is a nice app. It is an app whose `computeOutcome` return value is its `encodedState`.

### Rename `TwoPartyEthAsLump ` to `TwoPartyFixedOutcomeETHInterpreter`
This name was also confusing and so now it references the exact outcome type and asset class it is meant to be an interpreter for. Additionally, the virtual equivalent `TwoPartyVirtualEthAsLump` which manages the distribution of assets between an endpoint and an intermediary was renamed to `TwoPartyFixedOutcomeFromVirtualAppETHInterpreter`.

### Rename `ETHBucket` to `FreeBalanceApp`
Since the Free Balance is taking on a larger role, I decided to give it a more unique name.

### Removing `rollup` from `@counterfactual/chain`
The only consumers of this package were internal to the codebase and didn't require anything other than access to the TypeScript source code, so I removed the rollup dependency and build step.

### Miscellaneous Code Cleanup
- Renamed `latestversionNumber` to `latestVersionNumber`
- Rename several function names to be more clear
- Make several methods in `free-balance.ts` functional instead of imperative
- Add lots of documentation in places
- Rename `validateSignature` to `requireValidSignatureOrThrowError`


Alonski:
Maybe we should wait to merge in Joels fixes before merging this, or anything else, in?

snario:
I'm going to split it up into multiple smaller PRs I think, consider this a draft atm

packages/node/src/models/free-balance.ts
```diff
@@ -8,30 +7,44 @@ import { xkeysToSortedKthAddresses } from "../machine";
...
   amount: {
     _hex: string;
   };
-}
+};
 
-export interface FreeBalanceState {
-  [tokenAddress: string]: PartyBalance[];
-}
+export type ActiveAppsBloomFilter = { [appInstanceIdentityHash: string]: true };
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this is not a bloom filter...please don't call it one

packages/apps/migrations/2_deploy_contracts.js
```diff
@@ -1,15 +1,10 @@
...
-const NimApp = artifacts.require("NimApp");
-const TicTacToeApp = artifacts.require("TicTacToeApp");
-const ETHUnidirectionalTransferApp = artifacts.require("ETHUnidirectionalTransferApp");
-
 const ARTIFACTS = [
-  HighRollerApp,
-  NimApp,
-  TicTacToeApp,
-  ETHUnidirectionalTransferApp
+  artifacts.require("HighRollerApp"),
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/apps/test/nim.spec.ts
```diff
@@ -2,7 +2,6 @@ import { SolidityABIEncoderV2Type } from "@counterfactual/types";
 import chai from "chai";
 import * as waffle from "ethereum-waffle";
 import { Contract } from "ethers";
-import { AddressZero } from "ethers/constants";
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/chain/package.json
```diff
@@ -2,17 +2,14 @@
   "name": "@counterfactual/chain",
   "version": "0.0.1",
   "description": "Utility package to spin up a blockchain with deployed contracts for testing purposes",
-  "main": "dist/index.js",
-  "types": "dist/src/index.d.ts",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/chain/src/contract-deployments.jest.ts
```diff
@@ -1,28 +1,37 @@
...
-import TwoPartyEthAsLump from "@counterfactual/contracts/build/TwoPartyEthAsLump.json";
-import TwoPartyVirtualEthAsLump from "@counterfactual/contracts/build/TwoPartyVirtualEthAsLump.json";
-import UninstallKeyRegistry from "@counterfactual/contracts/build/UninstallKeyRegistry.json";
+import TwoPartyFixedOutcomeETHInterpreter from "@counterfactual/contracts/build/TwoPartyFixedOutcomeETHInterpreter.json";
+import TwoPartyFixedOutcomeFromVirtualAppETHInterpreter from "@counterfactual/contracts/build/TwoPartyFixedOutcomeFromVirtualAppETHInterpreter.json";
 import { NetworkContext } from "@counterfactual/types";
 import { ContractFactory, Wallet } from "ethers";
 
-export async function configureNetworkContext(wallet: Wallet) {
+export type NetworkContextForTestSuite = NetworkContext & {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/contracts/contracts/interpreters/CoinTransferETHInterpreter.sol
```diff
@@ -8,7 +8,7 @@ import "../libs/LibOutcome.sol";
  * This file is excluded from ethlint/solium because of this issue:
  * https://github.com/duaraghav8/Ethlint/issues/261
  */
-contract ETHInterpreter is Interpreter {
+contract CoinTransferETHInterpreter is Interpreter {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Should use a better name...suggest `ETHOfCoinTransferInterpreter`

packages/contracts/contracts/interpreters/TwoPartyFixedOutcomeETHInterpreter.sol
```diff
@@ -5,7 +5,7 @@ import "../interfaces/Interpreter.sol";
 import "../libs/LibOutcome.sol";
 
 
-contract TwoPartyEthAsLump is Interpreter {
+contract TwoPartyFixedOutcomeETHInterpreter is Interpreter {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why the name change? seems slightly worse IMO

packages/chain/src/contract-deployments.jest.ts
```diff
@@ -1,28 +1,37 @@
...
+  TicTacToeApp: string;
+};
+
+export async function deployTestArtifactsToChain(wallet: Wallet) {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this rename can be a separate PR

packages/contracts/contracts/default-apps/IdentityApp.sol
```diff
@@ -0,0 +1,17 @@
+pragma solidity 0.5.10;
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: can be added in a separate PR

packages/contracts/contracts/mixins/MixinChallengeRegistryCore.sol
```diff
@@ -22,6 +22,14 @@ contract MixinChallengeRegistryCore is MChallengeRegistryCore {
     return appChallenges[identityHash];
   }
 
+  function getAppStateHash(bytes32 identityHash)
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this function seems unused

packages/contracts/contracts/default-apps/FreeBalanceApp.sol
```diff
@@ -5,22 +5,23 @@ import "../interfaces/CounterfactualApp.sol";
 import "../libs/LibOutcome.sol";
 
 
-contract ETHBucket is CounterfactualApp {
+contract FreeBalanceApp is CounterfactualApp {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Should be called "LedgerApp" or similar since it now controls both the unassigned balance and the list of active apps

packages/contracts/migrations/2_deploy_contracts.js
```diff
@@ -1,29 +1,16 @@
...
-  ETHBucket,
-  MinimumViableMultisig,
-  MultiSend,
-  UninstallKeyRegistry,
-  ProxyFactory,
-  ConditionalTransactionDelegateTarget,
-  TwoPartyVirtualEthAsLump,
-  ETHInterpreter,
-  TwoPartyEthAsLump
+  artifacts.require("ChallengeRegistry"),
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/contracts/test/challenge-registry-dispute.spec.ts
```diff
@@ -66,7 +66,7 @@ describe("ChallengeRegistry Challenge", () => {
 
   let setStateAsOwner: (versionNumber: number, appState?: string) => Promise<void>;
   let latestState: () => Promise<string>;
-  let latestversionNumber: () => Promise<number>;
+  let latestVersionNumber: () => Promise<number>;
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/contracts/test/challenge-registry.spec.ts
```diff
@@ -40,7 +40,7 @@ describe("ChallengeRegistry", () => {
   let cancelChallenge: () => Promise<void>;
   let sendSignedFinalizationToChain: () => Promise<any>;
   let latestAppState: () => Promise<string>;
-  let latestversionNumber: () => Promise<number>;
+  let latestVersionNumber: () => Promise<number>;
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/node/src/ethereum/virtual-app-set-state-commitment.ts
```diff
@@ -64,7 +64,7 @@ export class VirtualAppSetStateCommitment extends EthereumCommitment {
   }
 
   // overrides EthereumCommitment::Transaction
-  public transaction(
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/node/src/methods/app-instance/propose-install-virtual/operation.ts
```diff
@@ -68,47 +68,48 @@ export function getNextNodeAddress(
...
+  thisAddress: string,
+  intermediaries: string[]
 ): boolean {
   return intermediaries.includes(thisAddress);
 }
 
-export async function getOrCreateVirtualChannel(
-  initiatorIdentifier: string,
-  respondingIdentifier: string,
+export async function getOrCreateStateChannelThatWrapsVirtualAppInstance(
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/node/src/methods/app-instance/propose-install-virtual/operation.ts
```diff
@@ -68,47 +68,48 @@ export function getNextNodeAddress(
...
-  thisAddress: Address,
-  intermediaries: Address[]
+  thisAddress: string,
+  intermediaries: string[]
 ): boolean {
   return intermediaries.includes(thisAddress);
 }
 
-export async function getOrCreateVirtualChannel(
-  initiatorIdentifier: string,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: good catch

packages/node/src/methods/app-instance/propose-install-virtual/operation.ts
```diff
@@ -68,47 +68,48 @@ export function getNextNodeAddress(
...
+  initiatorXpub: string,
+  respondingXpub: string,
   intermediaries: string[],
   store: Store
 ): Promise<StateChannel> {
-  let channel: StateChannel;
+  let stateChannel: StateChannel;
+
   try {
-    channel = await getChannelFromPeerAddress(
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: separate PR

packages/contracts/contracts/interpreters/CoinTransferETHInterpreter.sol
snario: TBH I think `CoinTransferETHInterpreter` is a better name than `ETHOfCoinTransferInterpreter`

packages/contracts/contracts/interpreters/TwoPartyFixedOutcomeETHInterpreter.sol
snario: - `Lump` is never used in any other context
- `ETH` should always be capitalized
- The outcome type is `TwoPartyOutcome` not `TwoParty`
- Standardizes the interpreter contract name as `<OutcomeType><AssetType>Interpreter`

I think it is much better

packages/contracts/contracts/default-apps/FreeBalanceApp.sol
snario: I think FreeBalance is more specific than Ledger because there is no on-chain methods to withdraw from it like there is in Nitro, instead it represents general structural data on the StateChannel. Maybe something like `RootApp` or something like this is clearer but for now I think `FreeBalance` is OK

packages/node/test/machine/integration/install-then-set-state.spec.ts
```diff
@@ -55,7 +58,7 @@ beforeAll(async () => {
 /**
  * @summary Set up a StateChannel and then install a new AppInstance into it.
  *
- * @description We re-use the ETHBucket App (which is the app ETH Free Balance uses)
+ * @description We re-use the FreeBalanceApp App (which is the app ETH Free Balance uses)
```
cf19drofxots: What is `ETH Free Balance` in this case?

packages/node/test/machine/integration/install-then-set-state.spec.ts
```diff
@@ -137,45 +144,68 @@ describe("Scenario: install AppInstance, set state, put on-chain", () => {
...
-        freeBalanceETH.timeout,
-        appInstance.appSeqNo,
-        network.ETHInterpreter,
+        identityAppInstance.identityHash,
+        stateChannel.freeBalance.identityHash,
+        network.CoinTransferETHInterpreter,
         defaultAbiCoder.encode(
-          ["uint256"],
-          [freeBalanceETH.coinTransferInterpreterParams!.limit]
+          ["tuple(uint256 limit)"],
```
cf19drofxots: This encoding should be defined in one place and used throughout the rest of the codebase.

packages/specs/source/peer-protocol.md
```diff
@@ -32,28 +30,21 @@ A protocol consists of the following components:
 
 ## Primitive Types
 
-|   Type    |          Description          |                                                      Sample                                                       |
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: we are using this extension: https://marketplace.visualstudio.com/items?itemName=shuworks.vscode-table-formatter

please do not format it any other way

packages/simple-hub-server/src/middlewares/validate-signature.ts
```diff
@@ -18,10 +18,10 @@ import { AuthenticatedRequest, JsonApiResource } from "../types";
  * `signature_required` error.
  */
 export default function(app: Application) {
-  return validateSignature(app);
+  return requireValidSignatureOrThrowError(app);
 }
 
-function validateSignature(app: Application) {
+function requireValidSignatureOrThrowError(app: Application) {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I prefer the more concise `assertIsValidSignature`

PR #1676, opened 2019-06-12
[node] ERC20 support spike

cf19drofxots:
This implements item 1 in [the ERC20 support issue](https://github.com/counterfactual/monorepo/issues/1668) which is providing the functionality to deposit ERC20 tokens into a channel.

PR #1792, opened 2019-07-02
[node] Update free balance state

cf19drofxots:
This is an extraction of part of the work done in #1676 to support ERC20 calls.

This updates the state model of the free balance to more generally handle other assets, though this PR itself doesn't concern itself with ERC20 modifications, it only paves the way for another upcoming PR that'll do other parts of the work done in #1676 to support ERC20 calls.


packages/contracts/contracts/default-apps/ETHBucket.sol
```diff
@@ -8,15 +8,16 @@ import "../libs/LibOutcome.sol";
 contract ETHBucket is CounterfactualApp {
 
   struct AppState {
-    LibOutcome.CoinTransfer[] transfers;
+    address[] tokens;
+    LibOutcome.CoinTransfer[][] balances;
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: should add a comment specifying what the two indices into `balances` are (I assume they are token address and then recipient)

packages/node/jest.config.js
```diff
@@ -32,6 +32,6 @@ module.exports = {
   "transform": {
     "^.+\\.tsx?$": "ts-jest"
   },
-  "verbose": true,
+  "verbose": false,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I think this suppresses console.log or something...

packages/node/src/ethereum/utils/funds-bucket.ts
```diff
@@ -0,0 +1,75 @@
+import { AppInterface } from "@counterfactual/types";
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: please mark this as moved from `eth-bucket.ts`

packages/node/src/models/free-balance.ts
```diff
@@ -0,0 +1,159 @@
...
+export interface PartyBalanceMap {
+  [to: string]: BigNumber;
+}
+
+export interface PartyBalance {
+  to: string;
+  amount: BigNumber;
+}
+
+export interface PlainPartyBalance {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Recommend naming it `HexPartyBalance` instead

packages/node/src/models/state-channel.ts
```diff
@@ -216,34 +176,37 @@ export class StateChannel {
...
-    const freeBalance = this.freeBalanceAppInstance;
-    const freeBalanceState = freeBalance!.state as ETHBucketAppState;
+  /**
+   * Updates the balances of the parties for the specified asset type.
+   * @param increments
+   * @param tokenAddress
+   */
+  public incrementFreeBalance(
+    increments: PartyBalanceMap,
+    tokenAddress = AddressZero
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I would rather make this non-optional and pass it in every time

packages/node/jest.config.js
cf19drofxots: I _think_ jest behavior has changed since then. I had `verbose` set to true and logs weren't showing up unless `sleep`ed surrounding the relevant code. Putting verbose to `false` displayed the logs without needing to sleep.

PR #1793, opened 2019-07-02
[node] camel-case appversionNumber

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


cf19drofxots:
Looks like this update wasn't applied to all instances?

PR #1795, opened 2019-07-02
[contracts] rename ChallengeStatus enumerated values

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/contracts/contracts/libs/LibStateChannelApp.sol
```diff
@@ -9,8 +9,8 @@ contract LibStateChannelApp {
   // The status of a challenge in the ChallengeRegistry
   enum ChallengeStatus {
     NO_CHALLENGE,
-    CHALLENGE_IS_OPEN,
-    CHALLENGE_WAS_FINALIZED
+    FINALIZES_AFTER_DEADLINE,
```
cf19drofxots: @IIIIllllIIIIllllIIIIllllIIIIllllIIIIll what do you think of `WILL_FINALIZE_AFTER_DEADLINE` and `HAS_FINALIZED` respectively?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I think `WILL_FINALIZE_AFTER_DEADLINE` is slightly inaccurate when the deadline has already passed

PR #1613, opened 2019-06-03
Implements JSON-RPC as a new message format for the Node methods and events

joelalejandro:
Resolves #1614.

While tests are passing in all affected packages, please review this PR with 🦅 eyes, since the scope is quite big and the surface area, albeit not implies altering core functionality (such as controller logic), does modify the way the Node talks to the outer world.

Basically every NodeMessage is now wrapped into a JSONRPC payload, as described in the master issue for this PR. Node-to-node events were left out after discussing with @snario, as there's a plan to rethink that aspect of the messaging layer.

PR #1804, opened 2019-07-05
[playground] [node] [cf-wallet] [cf.js] Fixes most of the messaging layer problems

joelalejandro:
This PR brings Playground's onboarding flow back to working order.

When #1613 was implemented, the messaging layer incurred into lots of changes across the codebase. While tests did pass, they did so because the tests are written with mocks. There was a severe discrepancy between the mocks and the real code.

It all boils down to trying to handle both `procotolMessageEvent`-like payloads with `Rpc`-like payloads with the same methods, which ended up breaking up communication between Nodes.

This is the path a payload must go through in order to reach a `NodeController`. I'll be describing the problems and changes done in each step.

0. `provider.deposit()` is called. This internally derives a method execution to `callRawNodeMethod`.
1. `provider.callRawNodeMethod()` sends a request to execute a Node method following the `Node.RpcMethodName` enumerable, with a given payload. This function converts the payload to a JSONRPC and passes it to the `NodeProvider`.
2. `nodeProvider.sendMessage()` passes the payload with no transformations, as it is a bridge to the `messagePort` the Playground uses to process messages. The NodeProvider sends the message with `messagePort.postMessage`.
3. The Playground `Counterfactual` class listens for the `onMessage` DOM event and received the payload to properly route it through the node.
>  - **Here is the source of the problem**: This function wasn't equipped to deal with RPC payloads. Changes were made to make that work:
>    - If the message is _not_ an RPC payload compatible with `RpcRouter` (i.e. `!rpc.methodName && !rpc.parameters`), it is relayed as it was before, using `relayMessage()` which passes the message using `node.emit()`.
>    - If the message _is_ an RPC payload, it is relayed using `relayJsonRpcMessage()`, which passes the message using `node.router.dispatch()`.
4. The message gets to the Node, and it is handled by the `RequestHandler` if it is a legacy message or by the `RpcRouter` if it is a `JsonRpc`.
5. Either the `RequestHandler` or the `RpcRouter` instantiate a `NodeController` to execute the method and get the result back to the other side.
>  - **Here is the second part of the problem**: `RpcRouter` wasn't emitting events through `RequestHandler.outgoing`, simply returning them asynchronously. There was no possibility for the CF clients to be aware of the results. Also, the result was badly serialized, since the `requestId` was missing and it couldn't be matched in CF.js or CF-Wallet.js' `requestListeners` used by `callRawNodeMethod`.
6. The message is captured by the Playground's Node event listeners set on the `Counterfactual.setupNodeProvider()` and passed via `MessagePort` using the `postToPort()` function.
>  - **There were events we weren't prepared to listen**: `DEPOSIT_CONFIRMED` was not coming back to the originating node because the `notifyCounterparty` was being set to `false` on response. Also, we're now listening to `chan_*`-prefixed methods so the promise handler in `callRawNodeMethod` can capture responses properly.
7. The `NodeProvider` passes back the message, emitting the `message` event to the CF `Provider`. At this point the `callRawNodeMethod` takes control of the response, analyzes it and returns it to the originating method.


Alonski:
Is there a reason that we are still supporting the legacy messaging system? Why not just rip it out and only use the RPC format?

joelalejandro:
@Alonski I'm not confortable playing around with the `protocolMessageEvent`. @snario mentioned there were plans to refactor the entire event system, maybe he can provide better context?

snario:
Basically there are 2 use cases that need to be managed.

**1. Receiving a `ProtocolMessage` for the first time.**

If this is the case, we check the message and dispatch it to the `ProtcolMessageEventController` which just forwards the message to the `runProtocolWithMessage` method on the `InstructionExecutor`. Then, the generator function inside of the `ProtocolFlow` gets triggered and we're good.

**2. Receiving a `ProtocolMessage` for a _mid-execution_ protocol.**

If this is the case, the message must be forwarded to the existing `yield` in the generator function that is in the process of executing. This is handled by an in-memory variable in the `Node` called `ioSendDeferrals` which is just a mapping from `protocolExecutionId` (a unique identifier for the mid-execution protocol) to a deferral object (which is just a promise with an external `resolve` method on it).

**So,** the important thing to note is that when a message gets received by the `Node`, it always goes to the `handleReceivedMessage` method which, depending on if this is scenario (1) or scenario (2), dispatches the received message to either (1) the `ProtocolMessageEventController` or (2) the `resolve` method of the deferral.

As long as the messages received are serialized and deserialized before they enter the `ProtocolMessageEventController` or are passed to `resolve`, we should be fine to use JSONRPC.

@joelalejandro @Alonski

PR #1823, opened 2019-07-08
[chain] remove rollup

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
The only consumers of this package were internal to the codebase and didn't require anything other than access to the TypeScript source code, so I removed the rollup dependency and build step.


packages/chain/package.json
```diff
@@ -2,17 +2,14 @@
...
-  "types": "dist/src/index.d.ts",
+  "main": "src/index.ts",
   "license": "MIT",
   "engines": {
     "yarn": "1.12.3",
     "node": "10.15.3"
   },
   "scripts": {
-    "build": "tsc -b . && rollup -c",
-    "build:contracts": "cd ../contracts && yarn build && cd - && tsc -b . && rollup -c",
```
cf19drofxots: I wouldn't remove this script. It's helpful in the context of dev flow.

PR #1827, opened 2019-07-08
[node] Deposit ERC20

cf19drofxots:
This adds support for depositing ERC20 tokens into a channel.

Mainly, the API for the deposit method has been extended to include an optional parameter: `token` address, allowing the operator to specify which ERC20 token they'd like to deposit into the channel.

https://github.com/counterfactual/monorepo/blob/48f820a0fcc90fb8860a52cfd639d1ca2490cbcb/packages/types/src/node.ts#L151-L156

An upcoming PR will add support with withdrawals, and another PR after that to add support for using deposited ERC20 tokens to fund apps.


packages/contracts/contracts/default-apps/CoinBalanceRefundApp.sol
```diff
@@ -21,7 +24,12 @@ contract ETHBalanceRefundApp {
 
     LibOutcome.CoinTransfer[] memory ret = new LibOutcome.CoinTransfer[](1);
 
-    ret[0].amount = address(appState.multisig).balance - appState.threshold;
+    if (appState.token == address(0x0)) {
+      ret[0].amount = address(appState.multisig).balance - appState.threshold;
+    } else {
+      // solium-disable-next-line max-len
```
snario: Use newline characters

packages/node/src/methods/app-instance/get-free-balance/controller.ts
```diff
@@ -20,14 +28,23 @@ export default class GetFreeBalanceController extends NodeController {
...
   ): Promise<Node.GetFreeBalanceStateResult> {
     const { store } = requestHandler;
-    const { multisigAddress } = params;
+    const { multisigAddress, tokenAddress } = params;
 
     if (!multisigAddress) {
       Promise.reject(NO_STATE_CHANNEL_FOR_MULTISIG_ADDR);
     }
 
+    const tokenAddr = tokenAddress ? tokenAddress : AddressZero;
```
snario: Doesn’t need to be a variable, just place the value in the lookup in fbState. Also clearer variable names like freeBalanceState

snario: Only return at end of function

packages/node/src/methods/state-channel/deposit/controller.ts
```diff
@@ -34,22 +42,34 @@ export default class DepositController extends NodeController {
     const { store, provider } = requestHandler;
     const { multisigAddress, amount } = params;
 
+    params.tokenAddress = params.tokenAddress
```
snario: Don’t mutate the params object

packages/node/src/methods/state-channel/deposit/controller.ts
```diff
@@ -34,22 +42,34 @@ export default class DepositController extends NodeController {
...
+
     const channel = await store.getStateChannel(multisigAddress);
 
     if (
       channel.hasAppInstanceOfKind(
-        requestHandler.networkContext.ETHBalanceRefundApp
+        requestHandler.networkContext.CoinBalanceRefundApp
       )
     ) {
       return Promise.reject(CANNOT_DEPOSIT);
```
snario: Throw an error instead of promise. Stack trace gets shown

packages/node/src/methods/state-channel/deposit/controller.ts
```diff
@@ -34,22 +42,34 @@ export default class DepositController extends NodeController {
...
     }
 
     const address = await requestHandler.getSignerAddress();
 
-    const balanceOfSigner = await provider.getBalance(address);
+    if (params.tokenAddress !== AddressZero) {
```
snario: Use the variable from my PR. CONVENTION_FOR_ETH_TOKEN_ADDRESS instead of AddressZero. It makes it clear what it means

packages/node/src/methods/state-channel/deposit/controller.ts
```diff
@@ -34,22 +42,34 @@ export default class DepositController extends NodeController {
...
+      const contract = new Contract(params.tokenAddress, ERC20.abi, provider);
+      const balance: BigNumber = await contract.functions.balanceOf(address);
```
snario: This could easily fail, catch errors for network issues or chain errors for example what if you get revert

packages/node/src/methods/state-channel/deposit/operation.ts
```diff
@@ -98,33 +101,44 @@ export async function makeDeposit(
...
-        outgoing.emit(NODE_EVENTS.DEPOSIT_FAILED, e);
-        console.error(`${DEPOSIT_FAILED}: ${e}`);
-        return false;
-      }
-
-      retryCount -= 1;
-
-      if (retryCount === 0) {
-        throw new Error(`${DEPOSIT_FAILED}: ${e}`);
+  if (tokenAddress === AddressZero) {
```
snario: Maintain the try again logic for tokens

packages/node/src/methods/state-channel/deposit/operation.ts
```diff
@@ -173,22 +189,31 @@ async function getDepositContext(
...
   publicIdentifier: string,
   provider: BaseProvider,
-  networkContext: NetworkContext
+  networkContext: NetworkContext,
+  tokenAddress: string
 ): Promise<DepositContext> {
   const { multisigAddress } = params;
-  const threshold = await provider.getBalance(multisigAddress);
+  const threshold =
+    tokenAddress === AddressZero
```
snario: You have to do this kind of if else logic a lot in this PR. Try to abstract these situations it into a class of some kind

packages/node/src/models/state-channel.ts
```diff
@@ -529,3 +536,19 @@ export class StateChannel {
     );
   }
 }
+
+function addTokenBalancetoFreeBalance(
```
snario: Capital T

Also see my PR for function naming inspiration

packages/node/src/protocol/uninstall.ts
```diff
@@ -121,7 +126,13 @@ async function proposeStateTransition(
     provider
   );
 
-  const newStateChannel = sc.uninstallApp(appIdentityHash, increments);
+  const newStateChannel = sc.uninstallApp(
+    appIdentityHash,
+    increments,
+    // installing/uninstalling apps with ERC20 is not yet supported
+    // so all installs/uninstalls default to ETH
```
snario: This comment makes no sense

Why does the function take a token parameter

packages/node/test/integration/secure-deposit.spec.ts
```diff
@@ -1,7 +1,11 @@
-import { One } from "ethers/constants";
+import DolphinCoin from "@counterfactual/contracts/build/DolphinCoin.json";
+import { ContractABI } from "@counterfactual/types";
+import { Contract, Wallet } from "ethers";
+import { One, Two, Zero } from "ethers/constants";
 import { JsonRpcProvider } from "ethers/providers";
+import { BigNumber } from "ethers/utils";
 
-import { Node } from "../../src";
+import { INSUFFICIENT_ERC20_FUNDS, Node } from "../../src";
```
snario: Why is this error exported from this path

packages/node/test/integration/secure-deposit.spec.ts
```diff
@@ -39,4 +43,125 @@ describe("Node method follows spec - deposit", () => {
...
+      await erc20Contract.functions.balanceOf(await nodeA.signerAddress())
+    ).toEqual(Zero);
+
+    let error;
+    try {
+      await nodeA.router.dispatch(depositReq);
+    } catch (e) {
+      error = e;
+    } finally {
+      expect(error).toEqual(
```
snario: I think jest has a built in error expect tool you should use that

packages/node/test/integration/secure-deposit.spec.ts
```diff
@@ -39,4 +43,125 @@ describe("Node method follows spec - deposit", () => {
...
+    expect(await erc20Contract.functions.balanceOf(multisigAddress)).toEqual(
+      Two
+    );
+
+    await confirmEthAndERC20FreeBalances(
+      nodeA,
+      multisigAddress,
+      erc20ContractAddress
+    );
+    await confirmEthAndERC20FreeBalances(
```
snario: Always surround multi line commands with white peace

packages/node/test/integration/secure-deposit.spec.ts
```diff
@@ -39,4 +43,125 @@ describe("Node method follows spec - deposit", () => {
...
+      nodeB,
+      multisigAddress,
+      erc20ContractAddress
+    );
+
+    // now deposits ETH
+
+    depositReq = makeDepositRequest(multisigAddress, One);
```
snario: Can you rename this helper method? Like constructDepositParams or something f

packages/node/test/integration/secure-deposit.spec.ts
```diff
@@ -39,4 +43,125 @@ describe("Node method follows spec - deposit", () => {
...
+    await nodeA.router.dispatch(depositReq);
+    await nodeB.router.dispatch(depositReq);
+
+    expect((await provider.getBalance(multisigAddress)).toNumber()).toEqual(
+      preDepositBalance.add(2).toNumber()
+    );
+
+    const freeBalanceState = await getFreeBalanceState(nodeA, multisigAddress);
+    for (const key in freeBalanceState) {
+      expect(freeBalanceState[key]).toEqual(One);
```
snario: Isn’t there some kind of expect that can evaluate all values in an object

packages/node/test/integration/secure-deposit.spec.ts
```diff
@@ -39,4 +43,125 @@ describe("Node method follows spec - deposit", () => {
...
+    }
+  });
 });
+
+/**
+ * @return the ERC20 token balance of the receiver
+ */
+async function transferERC20Tokens(
+  toAddress: string,
+  tokenAddress: string = global["networkContext"]["DolphinCoin"],
```
snario: Is the global object here evaluated at runtime?

packages/node/test/integration/secure-deposit.spec.ts
```diff
@@ -39,4 +43,125 @@ describe("Node method follows spec - deposit", () => {
...
+
+  await contract.functions.transfer(toAddress, amount);
+  const balanceAfter: BigNumber = await contract.functions.balanceOf(toAddress);
+
+  expect(balanceAfter.sub(balanceBefore)).toEqual(amount);
+
+  return balanceAfter;
+}
+
+async function confirmEthAndERC20FreeBalances(
```
snario: I do not recommend putting expect calls inside helper functions. Better to use helper functions to construct Dayan Andy then place the expect in the test function

packages/node/test/machine/integration/virtual-app-set-state-commitment.spec.ts
```diff
@@ -106,7 +106,7 @@ beforeEach(() => {
   );
 });
 
-describe("The virtualAppSetState transaction generated by the commitment", () => {
+describe.skip("The virtualAppSetState transaction generated by the commitment", () => {
```
snario: Fix

packages/types/src/data-types.ts
```diff
@@ -53,5 +53,5 @@ export enum TwoPartyFixedOutcome {
   SPLIT_AND_SEND_TO_BOTH_ADDRS = 2
 }
 
-export const ethBalanceRefundStateEncoding =
-  "tuple(address recipient, address multisig, uint256 threshold)";
+export const coinBalanceRefundStateEncoding =
```
snario: Use multi line string

packages/types/src/simple-types.ts
```diff
@@ -1,10 +1,14 @@
...
+import { BigNumberish, Interface, ParamType } from "ethers/utils";
 
 export type ABIEncoding = string;
 export type AppInstanceID = string;
 export type Address = string;
 export type Bytes32 = string;
 
+// This is copied from the ethers definition of how an ABI is typed.
+// tslint:disable-next-line:prefer-array-literal
+export type ContractABI = Array<string | ParamType> | string | Interface;
```
snario: Does ethers not export it? Why do we need it 

Also why are you disabling the linter

packages/apps/package.json
```diff
@@ -23,8 +23,8 @@
     "lint:ts": "tslint -c tslint.json -p ."
   },
   "devDependencies": {
-    "@counterfactual/contracts": "0.1.4",
-    "@counterfactual/types": "0.0.19",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why bump by 2 versions?

cf19drofxots: This was a mistake - will update

packages/node/src/methods/app-instance/get-free-balance/controller.ts
cf19drofxots: This doesn't make sense. If a token address is specified but it's not in the free balance, just placing the conditional operator inside the lookup the returned value is undefined.

packages/node/src/methods/state-channel/deposit/controller.ts
cf19drofxots: If this is not done here, this check needs to happen in the logic of the next 3 function calls. Also doesn't seem like we'd gain much by copying the params and modifying this member then passing that in.

packages/node/src/methods/state-channel/deposit/controller.ts
cf19drofxots: I wasn't waiting for that to be merged in, guess I'll just do it here.

packages/node/src/methods/state-channel/deposit/controller.ts
cf19drofxots: Addressed.

packages/node/src/methods/state-channel/deposit/operation.ts
cf19drofxots: Addressed.

packages/node/src/protocol/uninstall.ts
cf19drofxots: When an app gets installed, its params will need to indicate what asset it wants to use for collateral. This asset will be used to update the free balance when both doing an install and uninstall.

Since currently installing/uninstalling apps for ERC20 tokens is not supported, this defaults to ETH. When the PR for installing/uninstalling apps for ERC20 tokens is up, this will be modified to take the correct asset into account.

packages/node/test/integration/secure-deposit.spec.ts
cf19drofxots: The errors also get exported by `index.ts` in `/src`

packages/node/test/integration/secure-deposit.spec.ts
cf19drofxots: This is a pretty accurate name for what it does actually. We're not interested in the params, we quite simply want to make a request to deposit.

packages/node/test/integration/secure-deposit.spec.ts
cf19drofxots: The global object itself yes, but the `networkContext` members are not.

packages/node/test/integration/secure-deposit.spec.ts
cf19drofxots: I don't see what's particularly bad about this. The high-level test sets up the context and is very simple about what it expects to happen, while the logic of the more minute checks is carried out one level down.

If all expects were to be expected at the high level then most test suites can become overly verbose and we lose clarity of how the test achieves what it tries to achieve.

packages/types/src/simple-types.ts
cf19drofxots: I thought it did but apparently not.

The linter doesn't like it when `Array` is used instead of an array literal (which honestly seems like a sub-optimal restriction - what's the disadvantage in using `Array`?)

packages/node/test/machine/integration/virtual-app-set-state-commitment.spec.ts
cf19drofxots: Seems like this problem is not relevant to this branch. I pulled master on a fresh clone of the repo and the tests failed on this test.

packages/node/test/integration/secure-deposit.spec.ts
snario: We don’t need the same check in every test that uses the helper
Just write a smaller test

packages/node/src/protocol/uninstall.ts
snario: The function should not take a parameter it does not intend on using

PR #1835, opened 2019-07-09
[monorepo] Bump versions and update yarn lock files

cf19drofxots:


packages/apps/package.json
```diff
@@ -23,8 +23,8 @@
     "lint:ts": "tslint -c tslint.json -p ."
   },
   "devDependencies": {
-    "@counterfactual/contracts": "0.1.5",
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Why two versions?

cf19drofxots: The apps package was already behind before this PR by one version.

PR #1796, opened 2019-07-02
[apps] Simple swap app

rhlsthrm:
# Description
Exchange functionality in app.

### Related issues

- [ ] Deploy preview is functional

PR #1836, opened 2019-07-09
[apps] Simple swap app, "rebased"

cf19drofxots:
This is effectively the same change set as https://github.com/counterfactual/monorepo/pull/1796 but with the noise of the ERC20 commits removed as that branch was a spike.


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Why was the test skip committed?

PR #1838, opened 2019-07-10
[WIP] [node] Event emission symmetry

rhlsthrm:
Attempted to remove the event emission updates from this PR #1736. Unsure if this is all that needs to be done to make the events emit symmetrically. Need some assistance from @snario to continue.


rhlsthrm:
Closing in favor of #1843 which has the correct base.

PR #1843, opened 2019-07-11
[WIP] [node] Event emission symmetry

rhlsthrm:
Attempted to remove the event emission updates from this PR #1736. Unsure if this is all that needs to be done to make the events emit symmetrically. Need some assistance from @snario to continue.

### DIFF
--- a/packages/node/src/events/protocol-message/controller.ts
+++ b/packages/node/src/events/protocol-message/controller.ts
@@ -2,7 +2,6 @@ import {
   Protocol,
   SetupParams,
   TakeActionParams,
-  UninstallParams,
   UninstallVirtualAppParams,
   WithdrawParams
 } from "../../machine";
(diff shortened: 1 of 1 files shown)


snario:
@joelalejandro how should we approach using `router.emit` and `outgoing.emit`?

rhlsthrm:
@joelalejandro any update on this? Need some guidance on how to make the event emission symmetric.

joelalejandro:
@rhlsthrm - sorry for the delayed response. In order to emit events through the new API, you should be using: 

```ts
router.emit(eventName, data, "outgoing");
```

rhlsthrm:
@joelalejandro thanks for the info, I will do that.

In that case, it's unclear what in the PR (#1736) this code was pulled out of actually addresses the issue we are trying to resolve (emitting events symmetrically on both nodes). The only things I saw in that PR that had anything to do with events was what I put in this PR.

cf19drofxots:
What's the status of this PR?

rhlsthrm:
I don't think this is relevant anymore, will close for now.

PR #1844, opened 2019-07-11
Some cleanups to variable naming and code organization

snario:


packages/node/src/network-configuration.ts
```diff
@@ -2,60 +2,71 @@ import RopstenContracts from "@counterfactual/contracts/networks/3.json";
 import RinkebyContracts from "@counterfactual/contracts/networks/4.json";
 import KovanContracts from "@counterfactual/contracts/networks/42.json";
 import {
-  ContractMigration,
-  NetworkContext,
-  networkContextProps
+  DeployedContractNetworksFileEntry,
+  NAMES_OF_DEPLOYED_CONTRACTS_REQUIRED_FOR_COUNTERFACTUAL,
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: The names networkContext and networkContextProps should be kept in sync

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why is it changed to `EXPECTED_CONTRACT_NAMES...`? What's wrong with `networkContextProps`

packages/node/test/unit/install.spec.ts
```diff
@@ -24,18 +28,28 @@ import {
   FreeBalanceStateJSON
 } from "../../src/models/free-balance";
 import { Store } from "../../src/store";
-import { EMPTY_NETWORK } from "../integration/utils";
 import { MemoryStoreService } from "../services/memory-store-service";
 
 import { createAppInstanceProposalForTest } from "./utils";
 
+const NETWORK_CONTEXT_OF_ALL_ZERO_ADDRESSES = EXPECTED_CONTRACT_NAMES_IN_NETWORK_CONTEXT.reduce(
```
cf19drofxots: This should be in a more generic place as it's not specific to this test suite.

packages/node/src/node.ts
```diff
@@ -79,24 +79,19 @@ export class Node {
...
-        this.blocksNeededForConfirmation = blocksNeededForConfirmation;
-      }
+    if (typeof networkContext === "string") {
+      this.networkContext = getNetworkContextForNetworkName(networkContext);
     } else {
-      // Used for testing / ganache
       this.networkContext = networkContext;
     }
+
+    this.blocksNeededForConfirmation = blocksNeededForConfirmation;
```
cf19drofxots: Remove line, this isn't doing anything

packages/node/test/unit/install.spec.ts
snario: It actually is specific to this test suite

cf19drofxots: How? It's a network context object with all zero addresses. I don't see how it's specific to this test suite.

snario: It is never used anywhere else

cf19drofxots: That's not a good reason to place a generic object in a test-specific context, but I digress.

PR #1846, opened 2019-07-11
Rename events folder to message-handling and simplify file structure

snario:
- delete no-ops in `eventNameToImplementation`
- delete `CREATE_CHANNEL` and `DEPOSIT_CONFIRMED` in `eventNameToImplementation`
- move `events/protocol-message` code to `message-handling/handle-protocol-message.ts`
- move all other `events/` folder code into a single file `message-handling/handle-node-message`


packages/node/src/utils.ts
```diff
@@ -21,29 +21,23 @@ export function hashOfOrderedPublicIdentifiers(addresses: Address[]): string {
...
- *        operation.
- * @param store
+ * @param myXpub - first xpub
+ * @param theirXpub - second xpub
+ * @param store - store to search within
  */
-export async function getChannelFromPeerAddress(
-  myIdentifier: string,
-  peerAddress: string,
+export async function getStateChannelBasedOnTwoXpubs(
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `BasedOn` is not a connector used in function names; `getStateChannelWithOwners` is better

packages/node/src/store.ts
```diff
@@ -199,6 +199,8 @@ export class Store {
       ],
       true
     );
+
+    return appInstanceInfo;
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: why does this need to return?

PR #1850, opened 2019-07-11
Remove AppInstanceInfo in Node

snario:
This removes the concept of an `AppInstanceInfo` from the `@counterfactual/node` package.

This was a historical mistake in my opinion, where we had taken a client-side concept (i.e., the `AppInstanceInfo`) from `cf.js` and started using it inside of the `Node`.

Return types from calls such as `GetAppInstance` no longer return an `AppInstanceInfo`, they return the full `AppInstance` (encoded as JSON) instead. This will effect downstream consumers of API calls e.g., Connext. However, the JSON representation of the `AppInstance` is a much richer data structure and it contains all of the information in the `AppInstanceInfo`.


packages/node/src/models/proposed-app-instance-info.ts
```diff
@@ -50,7 +49,7 @@ export interface AppInstanceProposalJSON {
  * This class captures said state for the duration of the proposal being made and
  * the respecting `AppInstance` is installed.
  */
-export class AppInstanceProposal implements AppInstanceInfo {
+export class AppInstanceProposal {
```
cf19drofxots: Is this class needed now that proposed and installed `AppInstance`s share the same model?

packages/cf-wallet.js/src/app-instance.ts
```diff
@@ -32,17 +33,28 @@ export class AppInstance {
...
-    this.myDeposit = info.myDeposit;
-    this.peerDeposit = info.peerDeposit;
-    this.timeout = info.timeout;
-    this.twoPartyOutcomeInterpreterParams =
-      info.twoPartyOutcomeInterpreterParams;
-    this.coinTransferInterpreterParams = info.coinTransferInterpreterParams;
-    this.intermediaries = info.intermediaries;
+
+    if (info["appInterface"] !== undefined) {
+      this.appDefinition = info["appInterface"].addr;
```
joelalejandro: Should we make an effort to make this a bit more type-safe? Something like:

```ts
if ("appInterface" in info) {
  // Replace T with the type that exposes `appInterface`.
  const { appInterface } = <T>info;
  // Assign data...
} else {
  // Replace T2 with the type that exposes all of these properties.
  const { appDefinition, abiEncodings, timeout } = <T2>info;
  // Assign data...
}
```

cf19drofxots: Done

PR #1854, opened 2019-07-12
Remove VirtualAppSetState special-case adjudicator function

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR adds `TimeLockedPassThrough.sol`. By wrapping virtual app instances in instances of this contracts, it should be possible to get rid of `virutalAppSetState`.

## Note

- There is no longer a unique identifier for the `StateChannel` that wraps virtual apps; instead the unique identifier for a channel is _always_ the `create2` computed address of the `MinimumViableMultisig` representing the set of owners.

- The intermediary no longer stores any `StateChannel` that contains the `VirtualApp`. Instead, inside of `signature2` and `signature3` of the `seq: 1` message from the initiating to the intermediary, the initiating passes the `virtualAppInstanceIdentityHash` and the representative `defaultOutcome` based on the `targetAppState`.


packages/contracts/contracts/default-apps/DefaultableProxy.sol
```diff
@@ -0,0 +1,35 @@
...
+import "../ChallengeRegistry.sol";
+
+/*
+Before `switchesOutcomeAt`, this contract should return the exact
+same outcome as the outcome `targetAppIdentityHash` returns. However,
+after `switchesOutcomeAt`, it should return `defaultOutcome`.
+
+`challengeRegistryAddress` is used to look up the outcome.
+*/
+contract DefaultableProxy {
```
snario: use correct natspec

packages/contracts/contracts/default-apps/DefaultableProxy.sol
```diff
@@ -0,0 +1,35 @@
...
+  function computeOutcome(bytes calldata encodedState)
+    external
+    view
+    returns (bytes memory)
+  {
+    AppState memory appState = abi.decode(encodedState, (AppState));
+    if (block.number >= appState.switchesOutcomeAt) {
+      return appState.defaultOutcome;
+    } else {
+      return ChallengeRegistry(appState.challengeRegistryAddress).getOutcome(appState.targetAppIdentityHash);
```
snario: newline

packages/contracts/contracts/default-apps/DefaultableProxy.sol
snario: name is confusing... maybe TimeDependentOutcome or something

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `TimeDependentOutcome` is too general, it includes every function which examines `block.number`, but this does so in a specific way

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Maybe `TimeLockedPassthrough`?

snario:
I noticed that there seems to be a flickering test issue where, sometimes, you will get this error:

```
Attempted to edit an appInstance that does not exist: identity hash = 0x9d62c75de2ddbb3f2ad808d799219e43f1bccb4d36753a21f50adedd1f1ed85c
```

and sometimes you will not.

I am unsure why this is happening exactly.

packages/node/src/methods/app-instance/uninstall-virtual/controller.ts
```diff
@@ -28,18 +28,24 @@ export default class UninstallVirtualController extends NodeController {
     const { store } = requestHandler;
     const { appInstanceId } = params;
 
-    const multisigAddress = await store.getMultisigAddressFromOwnersHash(
+    const multisigAddressForStateChannelWithInterediary = await store.getMultisigAddressFromOwnersHash(
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: mis-spelled

packages/node/src/store.ts
```diff
@@ -32,7 +41,8 @@ import { debugLog, hashOfOrderedPublicIdentifiers } from "./utils";
 export class Store {
   constructor(
     private readonly storeService: Node.IStoreService,
-    private readonly storeKeyPrefix: string
+    private readonly storeKeyPrefix: string,
+    private readonly network: NetworkContext
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Name this networkContext

packages/node/test/machine/integration/install-virtual.spec.ts
```diff
@@ -242,7 +242,8 @@ describe("Scenario: Install virtual app with ERC20, put on-chain", () => {
         beneficiaries,
         capitalProvided: parseEther("10"),
         expiryBlock: (await provider.getBlockNumber()) + 1000,
-        tokenAddress: erc20Contract.address
+        tokenAddress: erc20Contract.address,
+        timeLockedPassThroughIdentityHash: HashZero
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: So this test uses a commitment to the virtual app instance directly, without wrapping it in a TLPC contract... we should note this in a comment in the file

packages/node/test/machine/integration/virtual-app-install.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  network = global["networkContext"];
+
+  appRegistry = new Contract(
+    network.ChallengeRegistry,
+    ChallengeRegistry.abi,
+    wallet
+  );
+});
+
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: is this a rebase error...seems to be the same as install-virtual.spec.ts

packages/specs/source/protocols/uninstall-virtual-app.md
```diff
@@ -8,13 +8,12 @@ Three users run the protocol. They are designated as `initiator`, `responder`, a
 
 ## The `UninstallVirtualAppParams` type
 
-
-|            Field             |   type    |                         description                         |
-| ---------------------------- | --------- | ----------------------------------------------------------- |
+| Field                       | type      | description                                                |
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: Please format this table with the visual studio extension

snario: It is formatted

PR #1900, opened 2019-07-15
[node] Rename roles

cf19drofxots:
Rename `{initiating, responding}` to `{initiator, responder}`


cf19drofxots:
> Should be initiating instead of initiator for consistency

For consistency yes, though `initiating` by itself was flawed to start with (it was only fine when paired with `publicIdentifer`)

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
how come initiating is flawed if not paired with `publicIdentifier`?

cf19drofxots:
I'll rephrase: `initiator` is the entity doing the action. `initiating` by itself is not an entity unless paired with one, such as a public identifier (who'd be _initiating_ the action, or who'd be the _initiator_)

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
oh ok, can you rename `initiating` -> `initiator` across the codebase then

PR #1905, opened 2019-07-16
rename contract

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Is the procedure here that I redeploy the contact to kovan and then merge this?

cf19drofxots:
If it's just the contracts being updated in isolation, then you'd bump the package version as well so the new contracts source and build artifacts get published. This's mean the client (Node) would be using an older version of contracts obviously.

PR #1918, opened 2019-07-18
[contracts,node] Fix mismanaged OutcomeType, Interpreter, and Virtual App code

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
- Redeployed new contracts for all contracts
- Renamed `CoinTransferInterpreter` to `MultiAssetMultiPartyCoinTransferInterpreter`
- Ensured recipient addresses in interpreters were marked as `payable` correctly
- Changed `RefundAppState` from `CoinTransfer[1][1]` to `CoinTransfer[2]`
- Used `SafeMath` everywhere
- Removed `FreeBalanceApp` since it is just `IdentityApp`
- Several code cleanups that don't change function but make code easier to read
- Add `SingleAssetTwoPartyCoinTransferFromVirtualAppInterpreter`
- Add `SingleAssetTwoPartyCoinTransferInterpreter`
- Add unit test cases for `SingleAssetTwoPartyCoinTransferInterpreter` contract functions
- Use TypeScript references for `local-ganache-server`
- Use the correct words for "serialize" and "deserialize" which were mixed up

## Interpreters

- `twoPartyOutcome`
- `multiAssetMultiPartyCoinTransfer`
- `singleAssetTwoPartyCoinTransfer`

For each of the above `OutcomeTypes`, and for both virtual apps and non-virtual-apps:

- The `AppInstance` model now has a field for the interpreter params
- There are typings for the interpreter params
- There are encoding constants in @counterfactual/types for the interpreter params
- There are accessor methods which throw if the `outcomeType` is wrong on `AppInstance`
- The `install`, `uninstall`, `install-virtual-app`, and `uninstall-virtual-app` correct handle each `outcomeType` for installation parameters and computing free balance decrements on uninstallation


packages/apps/test/eth-transfer-app.spec.ts
```diff
@@ -32,7 +32,7 @@ function mkAddress(prefix: string = "0xa"): string {
 
 function decodeBytesToAppState(encodedAppState: string): CoinTransferAppState {
   return defaultAbiCoder.decode(
-    [`tuple(tuple(address to, uint256 amount)[] transfers, bool finalized)`],
+    [`tuple(tuple(address to, uint256 amount)[2] transfers, bool finalized)`],
```
cf19drofxots: DRY this state encoding

packages/contracts/contracts/interpreters/CoinTransferFromVirtualAppInterpreter.sol
```diff
@@ -0,0 +1,59 @@
...
+      params,
+      (Agreement)
+    );
+
+    require(
+      block.number < agreement.expiryBlock,
+      "Virtual App agreement has expired and is no longer valid."
+    );
+
+    ERC20(agreement.tokenAddress).transfer(
```
cf19drofxots: `tokenAddress` might not be an ERC20 asset

packages/node/src/models/app-instance.ts
```diff
@@ -263,7 +264,8 @@ export class AppInstance {
     } catch (e) {
       // TODO: Catch ethers.errors.INVALID_ARGUMENT specifically in catch {}
       console.error(
-        "Attempted to setState on an app with an invalid state object"
+        "Attempted to setState on an app with an invalid state object",
```
cf19drofxots: You can easily format this string to include the app ID, the invalid object and the state encoding that it should adhere to.

packages/contracts/contracts/interpreters/CoinTransferFromVirtualAppInterpreter.sol
snario: Ah, good catch.

packages/node/jest.config.js
```diff
@@ -32,6 +32,6 @@ module.exports = {
   "transform": {
     "^.+\\.tsx?$": "ts-jest"
   },
-  "verbose": false,
+  "verbose": true,
```
snario: Fix.

PR #1919, opened 2019-07-19
[wallet-ui] Fixes `depositConfirmedEvent` on onboarding flow

joelalejandro:
After much debugging, @spersico and I discovered that both `protocolMessageEvent` and `depositConfirmedEvent` were being treated as events handled by the same system, which triggered the `no event handler` error when attempting to process that confirmation event.

We added an extra check to distinguish `protocolMessageEvent` and other legacy events from events currently handled by the `RpcRouter`.

This is a temporary fix until we eliminate the legacy `RequestHandler`.


joelalejandro:
As @cf19drofxots mentioned in Slack, CI is passing a false positive here. If the `secure-deposit` integration test changes to use `notifyCounterparty: true` in the deposit, we end up in a regression.

PR #1922, opened 2019-07-19
[simple-hub-server, bots] Remove subfolder yarn locks

cf19drofxots:
Unblocks https://github.com/counterfactual/monorepo/pull/1856


.circleci/config.yml
```diff
@@ -122,40 +122,6 @@ jobs:
...
-
-      - attach_workspace:
-          at: /home/circleci/project
-
-      - run: apt-get install -y jq # using this to mutate package.json
-
-      - run: npm i -g yarn@1.12.3 # because docker image uses 1.13.0
-
-      - run:
-          name: Check root yarn.lock
```
snario: We should keep this check

cf19drofxots: Ah thought this job was only for the subfolders. will update

PR #1945, opened 2019-07-20
Remove  the `owner` property on an AppIdentity

snario:
The owner property is used for:

1. Added uniqueness on a per `StateChannel` basis of an `AppInstance`
2. Allowing `CALL`s made by a multisignature wallet to override the `signingKeys` property on the adjudicator

This PR removes the ability for both (1) and (2) at the moment.

This means uniqueness for an `AppInstance` is only obtained by unique `signingKeys`, `appDefinition`, and `defaultTimeout` as of the current state of this pull request. It becomes the clients responsibility to never install an `AppInstance` that re-uses all three variables.

Because there is a possibility that `signingKeys` _may_ want to be re-used (although not clear), a new field could be added to the `AppIdentity` called `nonce` or `uniqueKey` or something like this, but it's unclear if that is needed.

### DIFF
--- a/packages/apps/package.json
+++ b/packages/apps/package.json
@@ -23,7 +23,7 @@
     "lint:ts": "tslint -c tslint.json -p ."
   },
   "devDependencies": {
-    "@counterfactual/contracts": "0.1.9",
+    "@counterfactual/contracts": "0.1.10",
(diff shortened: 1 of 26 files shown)


cf19drofxots:
Stay consistent with PR title format

Issue #1949, opened 2019-07-22
[node] initiatorBalanceDecrement is not a bignumber on counterparty's side for deposit app


LayneHaber:
Consistently getting the `initiatorBalanceDecrement.add is not a function` on the deposit app.

This only happens on the counterparty side (i.e. if the user initiates the deposit, these values are bignumbers, whereas on the hub's side they would not be), and they seem to be getting deserialized incorrectly when being pulled from the context.

I believe Nima is already working on this issue, but wanted to make sure there was an actual "issue" to track this. Casting as these are pulled out of the context solves the issue, but inelegantly.

PR #1957, opened 2019-07-22
[node] Casting quickfix in app install

LayneHaber:
### Description

There is a weird issue where the parameters are not being properly casted as a BN on the counterparty's deposit. This is an inelegant, but effective, fix where the parameters are always casted as they are parsed from the context.

### Related issues

Addresses issue #1949 


packages/node/src/protocol/install.ts
```diff
@@ -334,10 +334,10 @@ function computeStateChannelTransition(
   if (initiatorDepositTokenAddress !== responderDepositTokenAddress) {
     tokenIndexedBalanceDecrement = {
       [initiatorDepositTokenAddress]: {
-        [initiatorFbAddress]: initiatorBalanceDecrement
+        [initiatorFbAddress]: bigNumberify(initiatorBalanceDecrement)
```
cf19drofxots: Do we need to recast again here and below if we've done it in the context of this call above already?

snario:
Oh I see the issue.

The `params` aren't getting serialized.

It is the `params` variable here that needs to be serialized:

https://github.com/counterfactual/monorepo/blob/master/packages/node/src/message-handling/handle-protocol-message.ts#L39

I would write a generic serializer that detects `_hex` and serialize at this point.

There is similar code to copy here: https://github.com/counterfactual/monorepo/blob/master/packages/node/src/models/app-instance.ts#L108

cf19drofxots:
@rhlsthrm bump the node version (and its references in consumer packages) and we can merge this

rhlsthrm:
Bumped versions and changed fix to be more general.

PR #1964, opened 2019-07-22
Use name tokenAddress where possible

snario:


cf19drofxots:
Note that the redeployment of the contracts is coming up in a follow up PR.

PR #1975, opened 2019-07-23
Replace numerous toNumber() checks with better jest matcher

snario:


packages/node/test/integration/secure-withdraw.spec.ts
```diff
@@ -17,6 +18,8 @@ import {
   transferERC20Tokens
 } from "./utils";
 
+expect.extend({ toBeEq });
```
cf19drofxots: Why not `extend` in the global setup and replace all `BigNumber` comparison instances as done below?

PR #1977, opened 2019-07-23
Better testing of CoinTransferInterpreter

snario:


cf19drofxots:
This doesn't effect contract functionality so deployed addresses don't need to be updated.

PR #1982, opened 2019-07-23
Make much more explicit different OutcomeTypes

snario:
This is mostly just to illustrate a point

### DIFF
--- a/packages/cf-wallet.js/src/app-instance.ts
+++ b/packages/cf-wallet.js/src/app-instance.ts
@@ -2,7 +2,7 @@ import {
   AppABIEncodings,
   AppInstanceInfo,
   AppInstanceJson,
-  CoinTransferInterpreterParams,
+  CoinTransferListOfListsInterpreterParams,
   TwoPartyFixedOutcomeInterpreterParams
 } from "@counterfactual/types";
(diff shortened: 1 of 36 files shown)


packages/cf-wallet.js/src/app-instance.ts
```diff
@@ -2,7 +2,7 @@ import {
   AppABIEncodings,
   AppInstanceInfo,
   AppInstanceJson,
-  CoinTransferInterpreterParams,
+  CoinTransferListOfListsInterpreterParams,
```
cf19drofxots: Although this is accurate, it doesn't immediately convey why it's a list of lists.

Maybe something like `AssetBasedListOfCoinTransfersInterpreterParams` ?

packages/contracts/contracts/libs/CommonOutcomes.sol
```diff
@@ -2,13 +2,21 @@ pragma solidity 0.5.10;
...
   struct CoinTransfer {
     address to;
     uint256 amount;
   }
 
+  struct CoinTransfersList {
+    CoinTransfer[] transfers;
+  }
+
+  struct CoinTransferListOfLists {
```
cf19drofxots: See other comment

PR #1984, opened 2019-07-24
[node] Update function name that calculates free balance address

cf19drofxots:


Alonski:
@cf19drofxots This missed usages of `node.ethFreeBlanceAddress` in the bots:
https://github.com/counterfactual/monorepo/blob/master/packages/high-roller-bot/src/utils.ts#L58

I will open a PR soon.

PR #1986, opened 2019-07-24
[node] More rigorous testing of the TwoPartyFixedOutcome

cf19drofxots:
This adds tests for using the Node to exercise the different outcomes that are possible with the `TwoPartyFixedOutcome` outcome type.

### DIFF
--- a/packages/node/test/integration/uninstall.spec.ts
+++ b/packages/node/test/integration/uninstall.spec.ts
@@ -1,39 +1,128 @@
+import { One, Two, Zero } from "ethers/constants";
+
 import { Node } from "../../src";
+import { CONVENTION_FOR_ETH_TOKEN_ADDRESS } from "../../src/models/free-balance";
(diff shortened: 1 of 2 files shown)


packages/node/test/integration/uninstall.spec.ts
```diff
@@ -1,39 +1,128 @@
...
+
+      await nodeA.rpcRouter.dispatch(generateUninstallRequest(appInstanceId));
+
+      expect(await getInstalledAppInstances(nodeA)).toEqual([]);
+    });
+
+    it("installs an app with the TwoPartyFixedOutcome outcome and expects Node B to win total", async done => {
+      initialState.winner = 1;
+
+      const appInstanceId = await installTTTApp(
```
snario: I think we should generalize this function to install an `IdentityApp`

cf19drofxots: We should. I'll do this in the next PR.

PR #1997, opened 2019-07-24
[Wallet-UI] Add balance screen & Withdraw component & small fix

### LINKED ISSUE #1953: [wallet-ui] Add "Exchange" screen
Similar to what was introduced in the Playground by #320, the user needs to be capable of adding more funds or withdrawing them from their CF account.

When you click the "Balance" information box located at the header, the user will be faced with two forms:

image

The left one has the same features as the _Deposit_ flow that we've already implemented in #1732.

The right one is the inverse operation of a _deposit_: it's called _Withdraw_.

- It shows the Counterfactual account balance as "Available balance".
- The maximum amount you can withdraw is that of the CF account balance.
- The minimum amount you can withdraw is 0.1 ETH.
- The operation must be handled by a `withdraw` action in the Wallet store module.

### LINKED ISSUE #1983: [wallet-ui] Onboarding is redirecting to `/` instead of `/channels`
We might be missing a call to `getUser()` so the token is part of the available state when we enter `/channels`.

spersico:
### Description

This PR adds a first approach to the balance screen, with a withdraw component marked up and the functions actually working! (but with a giant bug that I'll make an issue later and link here).

It also "fixes" a small bug that joel found #1983 (it's just commented for now).
... and some warnings for unused variables in some mocks that somehow started yelling at me.

### Related issues
related #1983 
closes #1953
related to #2000


packages/wallet-ui/src/pages/account-withdraw/AccountWithdraw.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <Provider store={store}>
+      <Router initialEntries={["/"]}>
+        <AccountWithdraw {...props} />
+      </Router>
+    </Provider>
+  );
+  return { props, component, node: AccountWithdraw };
+}
+
+describe("<AccountRegistration />", () => {
```
joelalejandro: Should probably be `<AccountWithdraw />` :P

packages/wallet-ui/src/pages/account-withdraw/AccountWithdraw.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+  let props: RouteComponentProps;
+
+  beforeEach(() => {
+    const mock = setup();
+    component = mock.component;
+    instance = mock.component.find(Component);
+    props = mock.props;
+  });
+
+  it("should render a Proceed button", () => {
```
joelalejandro: Proceed *or Withdraw button

packages/wallet-ui/src/pages/account-withdraw/AccountWithdraw.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const CTA = component.find(testSelector("deposit-button"));
+    expect(CTA.exists()).toBe(true);
+    expect(["Proceed", "Withdraw"]).toContain(CTA.text());
+  });
+
+  it("should render the form input fields", () => {
+    expect(component.find(testSelector("amount-input")).exists()).toBe(true);
+  });
+
+  it("should trigger User Creation upon click", () => {
```
joelalejandro: should redirect to `/channels` after clicking the button

packages/wallet-ui/src/store/wallet/wallet.ts
```diff
@@ -112,6 +107,60 @@ export const deposit = (
...
+  provider: Web3Provider,
+  history?: History
+): ThunkAction<
+  void,
+  ApplicationState,
+  null,
+  Action<ActionType | WalletWithdrawTransition>
+> => async dispatch => {
+  try {
+    // 1. Ask Metamask to do the deposit. !
```
joelalejandro: withdraw*

packages/wallet-ui/src/store/wallet/wallet.ts
```diff
@@ -112,6 +107,60 @@ export const deposit = (
...
+    log('withdraw', transaction)
```
joelalejandro: Should we keep these logs?

packages/wallet-ui/src/store/wallet/wallet.ts
```diff
@@ -112,6 +107,60 @@ export const deposit = (
...
+    dispatch({ type: WalletWithdrawTransition.CheckWallet });
+    const responsew = await requestWithdraw(transaction); // IMPLEMENT THIS!
```
joelalejandro: typo: `response`

packages/wallet-ui/src/store/wallet/wallet.ts
```diff
@@ -112,6 +107,60 @@ export const deposit = (
...
+    log('withdraw responsew', responsew)
+
+    // 2. Wait until the deposit is completed in both sides. !
```
joelalejandro: withdraw*

packages/wallet-ui/src/store/wallet/wallet.ts
spersico: I don't know, I would leave at least the second one (which has the transaction response, with the hash) for debugging purposes, until we have that working reliably

PR #2001, opened 2019-07-24
Remove unused parameters

snario:


packages/cf.js/src/provider.ts
```diff
@@ -287,14 +287,12 @@ export class Provider {
...
 
-  /**
-   * @ignore
-   */
-  private validateEventType(eventType: EventType) {
-    // if (!this.validEventTypes.includes(eventType)) {
-    //   throw new Error(`"${eventType}" is not a valid event`);
-    // }
-  }
+  // FIXME: Either implement this or remove it
```
cf19drofxots: I'll defer to @joelalejandro on this as there's an issue to properly fix how messages and events are handled

Issue #1910, opened 2019-07-17
[testing] Some tests are not handling promises correctly


snario:
<img width="733" alt="Screen Shot 2019-07-17 at 4 50 53 PM" src="https://user-images.githubusercontent.com/1933029/61410827-15364000-a8b3-11e9-9312-5bde786d930d.png">

snario:
So there are 2 distinct issues:

1. Some promises are returned but the test does not handle them asyncronously

2. The node process throws that error message with the following message because there are exceptions emitting from it that get auto-resolved by the node process into rejections of the promise. We need to handle those.

```
node:7364) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3):
```

PR #2009, opened 2019-07-26
[node] Refactors some test utils and fixes some errors

cf19drofxots:
Fixes most of https://github.com/counterfactual/monorepo/issues/1910

### DIFF
--- a/packages/node/src/methods/app-instance/propose-install/controller.ts
+++ b/packages/node/src/methods/app-instance/propose-install/controller.ts
@@ -38,6 +38,11 @@ export default class ProposeInstallController extends NodeController {
     params: Node.ProposeInstallParams
   ): Promise<Queue[]> {
     const { store, publicIdentifier } = requestHandler;
+    const { initialState } = params;
+
+    if (!initialState) {
+      throw new Error(NULL_INITIAL_STATE_FOR_PROPOSAL);
+    }
 
     const {
       proposedToIdentifier,
@@ -93,11 +98,7 @@ export default class ProposeInstallController extends NodeController {
     params: Node.ProposeInstallParams
   ): Promise<Node.ProposeInstallResult> {
     const { store, publicIdentifier, messagingService } = requestHandler;
-    const { initialState, proposedToIdentifier } = params;
-
-    if (!initialState) {
-      throw new Error(NULL_INITIAL_STATE_FOR_PROPOSAL);
-    }
+    const { proposedToIdentifier } = params;
 
     const appInstanceId = await createProposedAppInstance(
       publicIdentifier,
--- a/packages/node/src/rpc-router.ts
+++ b/packages/node/src/rpc-router.ts
@@ -34,20 +34,24 @@ export default class RpcRouter extends Router {
       throw new Error(`Cannot execute ${rpc.methodName}: no controller`);
     }
 
-    const result = jsonRpcSerializeAsResponse(
-      {
-        result: await new controller.type()[controller.callback](
-          this.requestHandler,
-          rpc.parameters
-        ),
-        type: rpc.methodName
-      },
-      rpc.id as number
-    );
+    try {
+      const result = jsonRpcSerializeAsResponse(
+        {
+          result: await new controller.type()[controller.callback](
+            this.requestHandler,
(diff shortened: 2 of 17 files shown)


packages/node/src/methods/app-instance/propose-install/controller.ts
```diff
@@ -38,6 +38,11 @@ export default class ProposeInstallController extends NodeController {
     params: Node.ProposeInstallParams
   ): Promise<Queue[]> {
     const { store, publicIdentifier } = requestHandler;
+    const { initialState } = params;
+
+    if (!initialState) {
+      throw new Error(NULL_INITIAL_STATE_FOR_PROPOSAL);
+    }
```
snario: This belongs in the `beforeExecution` function.

packages/node/test/integration/install.spec.ts
```diff
@@ -85,8 +87,8 @@ describe("Node method follows spec - install", () => {
         await makeProposeCall(
           nodeA,
           nodeB,
-          (global["networkContext"] as NetworkContextForTestSuite).TicTacToeApp,
-          {},
+          Apps.TicTacToe,
```
snario: This should be named `TicTacToeApp`, the name of the `contract`.

packages/node/test/integration/tic-tac-toe.ts
```diff
@@ -1,10 +1,14 @@
...
+} from "@counterfactual/types";
 
-export const tttStateEncoding =
-  "tuple(uint256 versionNumber, uint256 winner, uint256[3][3] board)";
-
-export const tttActionEncoding =
-  "tuple(uint8 actionType, uint256 playX, uint256 playY, tuple(uint8 winClaimType, uint256 idx) winClaim)";
+export const tttAbiEncodings: AppABIEncodings = {
+  stateEncoding:
+    "tuple(uint256 versionNumber, uint256 winner, uint256[3][3] board)",
```
snario: Useful to make this a multiline string with backtick quotes.

packages/node/test/integration/install.spec.ts
```diff
@@ -10,13 +10,15 @@ import { toBeLt } from "../machine/integration/bignumber-jest-matcher";
 
 import { setup, SetupContext } from "./setup";
 import {
+  Apps,
```
snario: Maybe `TestAppDefinitions` instead of `Apps`?

packages/node/src/rpc-router.ts
```diff
@@ -34,20 +34,24 @@ export default class RpcRouter extends Router {
...
+        },
+        rpc.id as number
+      );
+
+      this.requestHandler.outgoing.emit(rpc.methodName, result);
+
+      return result;
+    } catch (e) {
+      throw new Error(e);
+    }
```
snario: This seems like a bizarre block... `try { ... } catch (e) { throw new Error(e) }` is equivalent to `...`.

packages/node/test/integration/cant-take-action.spec.ts
```diff
@@ -30,7 +30,7 @@ describe("Node method follows spec - fails with improper action taken", () => {
         }
       };
       await createChannel(nodeA, nodeB);
-      const appInstanceId = await installApp(nodeA, nodeB, Apps.TicTacToe);
+      const [appInstanceId] = await installApp(nodeA, nodeB, Apps.TicTacToe);
```
snario: Why does this method now return an array with one element in it?

packages/node/src/methods/app-instance/propose-install/controller.ts
cf19drofxots: Reason for moving it here is so that the request shouldn't even get queued if it would end up throwing anyway.

packages/node/test/integration/cant-take-action.spec.ts
cf19drofxots: The `installApp` function creates the proposal parameters that the proposal & install get called with, so if the caller cares about what those parameters were, it looks at the second element of this returned list.

Some callers care about this to test whether the installed app has the same parameters as the proposed parameters.

Issue #1916, opened 2019-07-18
[node] Allow easy way to register middleware into protocol


rhlsthrm:
We would like to be able to register middleware functions into specific protocols (i.e. before signing).

PR #2034, opened 2019-07-27
[node] Enable app-specific signing keys to be calculated from an externally-provided callback

cf19drofxots:
## Summary

In order for the Node to produce state-update commitments, it needs access to some signing keys.

There are two ways in which this is supported:

1. An extended private key string is provided to the Node at the "EXTENDED_PRIVATE_KEY_PATH" key of the store service that is passed as an argument to the Node. If no such value exists for this key, the Node produces an extended private key and sets it at this key. This extended private key is then used to generate a "public identifer" for the Node (the address by which the Node is known by). It is also used to generate private keys which are specific to AppInstances.

2. Instead of supplying a mnemonic, the Node operator supplies two other arguments:

- an extended public key which will serve as the "public identifier" of the Node, and will be used to generate signer addresses at AppInstance-specific derivation paths for signature verification in the protocols.
- a callback function that offers the generation of a signing key given a specific derivation path. This enables the consumer of the Node (i.e. wallets) to not reveal any mnemonics but provide the ability to sign state isolated to specific AppInstances.

 The `Node` package exposes a reference implementation of the second approach through a function named `generatePrivateKeyGeneratorAndXPubPair` which produces these 2 arguments given an extended private key.

Main function of interest:
https://github.com/counterfactual/monorepo/blob/efd462bdeebaff63d264e109207bf095da285de9/packages/node/src/private-keys-generator.ts#L68-L113

----

This PR addresses signing issue: https://github.com/counterfactual/monorepo/issues/1916

Note: This _does not_  fix a security issue where currently on master, apps sharing the same appSequenceNumber across all channels use the same signing key. The fix is outside the scope of this PR.


packages/node/src/signing-keys-generator.ts
```diff
@@ -0,0 +1,67 @@
+import { Node } from "@counterfactual/types";
+import { Wallet } from "ethers";
+import { hashMessage } from "ethers/utils";
+import { Memoize } from "typescript-memoize";
+
+export class SigningKeysGenerator {
```
snario: Maybe a more explicit name is `PrivateKeyGenerator`?

packages/node/src/signing-keys-generator.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Memoize()
+  /**
+   * @param appSequenceNumber
+   * @param multisigAddress The namespace under which `appSequenceNumber` is
+   *        used since the provided callback function is idempotent for any
+   *        value passed in. This prevents collisions that'd occur on  using the
+   *        same `appSequenceNumber` across different channels from receiving the
+   *        same signing key from the callback.
+   */
+  public async getSigningKey(
```
snario: `getPrivateKey`?

packages/node/src/signing-keys-generator.ts
```diff
@@ -0,0 +1,67 @@
...
+      )!;
+    }
+
+    // FIXME: fallback on mnemonic-based signer if callback isn't specified
+    // in which case the derivation path needs to be converted into a decimal
+    // representation for it to be a valid path
+    const signingKey = await this.privateKeyGenerator(appInstanceUniqueId);
+    try {
+      new Wallet(signingKey);
+    } catch (e) {
```
snario: I'm guessing `ethers` throws some very specific types of errors here, maybe catch those?

packages/node/src/private-keys-generator.ts
```diff
@@ -0,0 +1,154 @@
...
+import { BigNumber } from "ethers/utils";
+import { fromExtendedKey, fromMnemonic } from "ethers/utils/hdnode";
+import log from "loglevel";
+import { Memoize } from "typescript-memoize";
+
+import { CF_PATH } from "./constants";
+
+export const EXTENDED_PRIVATE_KEY_PATH = "EXTENDED_PRIVATE_KEY";
+
+export class PrivateKeysGenerator {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `implements IPrivateKeyGenerator`?

packages/node/src/private-keys-generator.ts
```diff
@@ -0,0 +1,154 @@
...
+    extendedPrvKey = fromMnemonic(Wallet.createRandom().mnemonic).extendedKey;
+    await storeService.set([
+      { key: EXTENDED_PRIVATE_KEY_PATH, value: extendedPrvKey }
+    ]);
+  } else {
+    log.info("Using extended private key found in the store.");
+  }
+  const [
+    privKeyGenerator,
+    pubExtendedKey
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `extendedPubKey` (or `xpub`) for consistency

packages/node/src/node.ts
```diff
@@ -59,9 +62,22 @@ export class Node {
     nodeConfig: NodeConfig,
     provider: BaseProvider,
     networkOrNetworkContext: "ropsten" | "kovan" | "rinkeby" | NetworkContext,
+    publicExtendedKey?: string,
+    privateKeyGenerator?: NodeTypes.IPrivateKeyGenerator,
     blocksNeededForConfirmation?: number
   ): Promise<Node> {
+    const [
+      privateKeysGenerator,
+      pubExtendedKey
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `extendedPubKey` or `xpub`

packages/node/src/private-keys-generator.ts
cf19drofxots: This class itself doesn't implement the interface, it's just a wrapper invoking the function that implements the interface and does the needed error checking for whether the generated private key is valid.

cf19drofxots: I'll rename the class, this is incorrectly named.

rhlsthrm:
This has bumped up in priority for us. Can we get an update on if there's anything blocking this? Besides of course the master update.

PR #2054, opened 2019-07-29
[node] check undefined

LayneHaber:
### Description

Fails if any value in the object is undefined, just added a check in the function to look for this case.

### Related issues


snario:
Makes sense, although technically this shouldn't ever happen unless the value `undefined` is stored in the place where serialized data is stored? I'm not sure in which scenario we would have deserialized data that is also not bigNumberified.

PR #2063, opened 2019-07-29
[node] catch all failing IEs in mininode tests

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Currently there is a bug in the mininode test harness. Spec files like `protocols.spec.ts` have a pointer to a single IE which they use to initiate protocols, and the outcome of the protocol initiated is awaited, hence errors are caught. However, initiating protocols causes other mininodes to receive messages and hence run protocols as well, but the spec file did not have a pointer to those outcomes. This bug can be reproduced by throwing an error after this line. https://github.com/counterfactual/monorepo/blob/aa3a7b5f28f1c1eb2a561b71cd9179f9b14b8dda/packages/node/src/protocol/setup.ts#L99

This PR fixes this bug.

### DIFF
--- a/packages/node/test/machine/integration/message-router.ts
+++ b/packages/node/test/machine/integration/message-router.ts
@@ -4,32 +4,47 @@ import { Opcode } from "../../../src/machine";
 import { MiniNode } from "./mininode";
 
 export class MessageRouter {
+  // mapping from a mininode's xpub to the mininode
   private nodesMap: Map<string, MiniNode>;
+
+  // mapping from a mininode's xpub to a promise representing the future value
+  // of an IO_SEND_AND_WAIT call. It is expected that the protocol is awaiting
+  // on this promise.
   private deferrals: Map<string, Deferred<any>>;
 
+  // when a message from a mininode causes a protocol to run in another node,
+  // a promise representing completion of the second protocol is added here.
+  private pendingPromises: Set<Promise<void>>;
+
   constructor(nodes: MiniNode[]) {
-    this.nodesMap = new Map<string, MiniNode>();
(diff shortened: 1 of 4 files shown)


packages/node/test/machine/integration/message-router.ts
```diff
@@ -32,6 +42,11 @@ export class MessageRouter {
     }
   }
 
+  private maybePush(v: Promise<void> | null) {
```
snario: Rename to something more explicit like `appendToPendingPromisesIfNotNull`

packages/node/test/machine/integration/protocols.spec.ts
```diff
@@ -64,9 +64,13 @@ describe("Three mininodes", () => {
       multisigAddress: multisigAB
     });
 
-    // todo: if nodeB/nodeC is still busy doing stuff, we should wait for it
+    // todo(xuanji): if nodeB/nodeC is still busy doing stuff, we should wait for it
 
-    mr.assertNoPending();
+    await mr.waitAll();
+
+    done();
```
snario: Why is this being added here? This seems like it ignores the entire test.

packages/node/test/machine/integration/message-router.ts
```diff
@@ -44,14 +59,18 @@ export class MessageRouter {
...
+      // finishes
+      return toNode.dispatchMessage(message);
     }
 
     deferred.resolve(message);
+    return null;
   }
 
-  public assertNoPending() {
+  public async waitAll() {
```
snario: Rename to more explicit `waitForAllPendingPromises`

PR #2068, opened 2019-07-30
Add more protocol tests

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR adds protocol tests run via the mininode test harness for installing and uninstalling direct apps with the 3 different outcome types and using either ETH or an ERC20 token


packages/node/src/models/state-channel.ts
```diff
@@ -212,6 +212,23 @@ export class StateChannel {
...
+
+    const ret = {} as any;
+
+    for (const tokenAddress of Object.keys(balances)) {
+      const ret2 = {} as any;
+      for (const coinTransfer of balances[tokenAddress]) {
+        ret2[coinTransfer.to] = coinTransfer.amount;
+      }
+      ret[tokenAddress] = ret2;
+    }
```
snario: I think this function is doing a similar transform for something we have a utility function for within `free-balance.ts`.

packages/node/test/machine/integration/protocols2.spec.ts
```diff
@@ -120,9 +110,9 @@ describe("Three mininodes", () => {
         responderXpub: mininodeC.xpub,
         defaultTimeout: 100,
         initialState: appState,
-        initiatorBalanceDecrement: Zero,
-        responderBalanceDecrement: Zero,
-        tokenAddress: CONVENTION_FOR_ETH_TOKEN_ADDRESS,
+        initiatorBalanceDecrement: One,
+        responderBalanceDecrement: One,
+        tokenAddress: getAddress("88a5c2d9919e46f883eb62f7b8dd9d0cc45bc290"),
```
snario: 1. You should not need to checksum a string literal in the code, instead either checksum it once and hard code it in the file.
2. Why magic constant? Perhaps instead use a dynamically generated string and put that in the test suite context?

packages/node/test/machine/integration/two-party-protocols.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+  for (const sc of mininodeB.scm) {
+    mininodeB.scm.set(
+      sc[0],
+      sc[1].incrementFreeBalance({
+        [CONVENTION_FOR_ETH_TOKEN_ADDRESS]: {
+          [sc[1].getFreeBalanceAddrOf(mininodeA.xpub)]: One,
+          [sc[1].getFreeBalanceAddrOf(mininodeB.xpub)]: One
+        },
+        [getAddress("0x88a5c2d9919e46f883eb62f7b8dd9d0cc45bc290")]: {
```
snario: Again try to avoid hard coding magic constants multiple times

packages/node/test/machine/integration/two-party-protocols.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+let identityApp: Contract;
+let mininodeA: MiniNode;
+let mininodeB: MiniNode;
+let multisigAB: string;
+let mr: MessageRouter;
+let participants: string[];
+
+expect.extend({ toBeEq });
+
+beforeEach(async () => {
```
snario: Can you add some documentation to describe what this `beforeEach` does?

packages/node/test/machine/integration/two-party-protocols.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        [getAddress("0x88a5c2d9919e46f883eb62f7b8dd9d0cc45bc290")]: {
+          [sc[1].getFreeBalanceAddrOf(mininodeA.xpub)]: One,
+          [sc[1].getFreeBalanceAddrOf(mininodeB.xpub)]: One
+        }
+      })
+    );
+  }
+});
+
+describe("I/U tests", () => {
```
snario: What is an I/U?

packages/node/test/machine/integration/two-party-protocols.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    await mininodeA.ie.initiateProtocol(Protocol.Uninstall, mininodeA.scm, {
+      appIdentityHash: key,
+      initiatorXpub: mininodeA.xpub,
+      responderXpub: mininodeB.xpub,
+      multisigAddress: multisigAB
+    });
+
+    await mr.waitForAllPendingPromises();
+  });
+  // todo
```
snario: Marking this as we need to do these

packages/types/src/simple-types.ts
```diff
@@ -19,9 +19,7 @@ export type SolidityValueType =
 // around in protocol messages and include this in transaction data in challenges,
 // we impose some restrictions on the type; they must be serializable both as
 // JSON and as solidity structs.
-export type SolidityABIEncoderV2Type =
-  | SolidityABIEncoderV2Struct
-  | SolidityABIEncoderV2SArray;
+export type SolidityABIEncoderV2Type = SolidityValueType
```
snario: Do we need both of these types? Maybe pick one and re-use across code.

tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "experimentalDecorators": true,
     "noImplicitAny": false,
     "noUnusedParameters": true,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
snario: Change back.

packages/node/src/models/state-channel.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: it's similar but not the same

packages/types/src/simple-types.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: fixed in #2082

packages/node/src/protocol/utils/get-outcome-increments.ts
```diff
@@ -150,18 +150,15 @@ function handleMultiAssetMultiPartyCoinTransfer(
...
-  );
+  const ret = {} as any;
+
+  interpreterParams.tokenAddresses.forEach((tokenAddress, index) => {
+    ret[tokenAddress] = convertCoinTransfersToCoinTransfersMap(
+      decodedTransfers[index]
+    );
+  });
+
+  return ret;
```
snario:
```typescript
return interpreterParams.tokenAddresses.reduce(
    (acc, tokenAddress, index) => ({
        ...acc,
        [tokenAddress]: convertCoinTransfersToCoinTransfersMap(
            decodedTransfers[index]
        )
    }),
    {}
)
```

packages/node/test/machine/integration/protocols/install-uninstall.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+import { Two, Zero } from "ethers/constants";
+
+import { CONVENTION_FOR_ETH_TOKEN_ADDRESS } from "../../../../src/constants";
+import { toBeEq } from "../bignumber-jest-matcher";
+
+import { Participant, TestRunner } from "./test-runner";
+
+expect.extend({ toBeEq });
+
+export enum TokenType {
```
snario: This enum seems to be for the test suite only, maybe give it a more explicit name

packages/node/test/machine/integration/protocols/install-uninstall.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    tr.assertFB(Participant.A, tokenAddress, Zero);
+    tr.assertFB(Participant.B, tokenAddress, Zero);
+    await tr.uninstall();
+    tr.assertFB(Participant.A, tokenAddress, Two);
+    tr.assertFB(Participant.B, tokenAddress, Zero);
+  }
+}
+
+describe("Install-then-uninstall in a direct channel", () => {
+  for (const outcomeType of [
```
snario: I think this entire for lop would be better off written  as explicit it statements

packages/node/test/machine/integration/protocols/install-uninstall.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+describe("Install-then-uninstall of a virtual app", () => {
+  for (const outcomeType of [
+    OutcomeType.TWO_PARTY_FIXED_OUTCOME,
+    OutcomeType.SINGLE_ASSET_TWO_PARTY_COIN_TRANSFER
+  ]) {
+    for (const tokenAddress of [
+      CONVENTION_FOR_ETH_TOKEN_ADDRESS,
+      TestRunner.TEST_TOKEN_ADDRESS
+    ]) {
+      it("test", async () => {
```
snario: Needs an explicit name

packages/node/test/machine/integration/protocols/protocols.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  appWithAction = await new ContractFactory(
+    AppWithAction.abi,
+    AppWithAction.bytecode,
+    wallet
+  ).deploy();
+});
+
+describe("Three mininodes", () => {
+  it("Can run all the protocols", async () => {
```
snario: This  test can be broken up by protocol and edit `describe` level variables along the way; then if any particular part fails we will have a more clear idea as to what the failure was

packages/node/test/machine/integration/protocols/test-runner.ts
```diff
@@ -0,0 +1,414 @@
...
+export enum Participant {
+  A,
+  B,
+  C
+}
+
+export class TestRunner {
+  static readonly TEST_TOKEN_ADDRESS = getAddress(
+    "88a5c2d9919e46f883eb62f7b8dd9d0cc45bc290"
+  );
```
snario: You should not need to run `getAddress` on a string literal.

snario: Also is this just a random string?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: It is a checksummed string

packages/node/test/machine/integration/protocols/install-uninstall.spec.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I disagree, it's much more LOC

packages/node/test/machine/integration/protocols/test-runner.ts
snario: `getAddress(
    "88a5c2d9919e46f883eb62f7b8dd9d0cc45bc290"
  )`

is equal to 

`"0x88a5c2d9919e46f883eb62f7b8dd9d0cc45bc290"`

seems like unnecessary complexity

packages/node/test/machine/integration/protocols/install-uninstall.spec.ts
snario: but less readable -- I prefer tests that are readable

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I don't think it will be more readable - the main worry is that you would have to scrutinize the 7 `it` statements, mentally reconstruct the 3x3 matrix of which features are tested (and which of those combinations are disabled), and then make sure that the 7 `it` statements actually implement the matrix properly

PR #2069, opened 2019-07-30
[node] pass tokenAddress parameter for install-virtual operation

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/node/src/methods/app-instance/install-virtual/operation.ts
```diff
@@ -24,6 +23,15 @@ export async function installVirtual(
 
   let updatedStateChannelsMap: Map<string, StateChannel>;
 
+  if (
+    proposal.initiatorDepositTokenAddress !==
```
snario: Instead of doing `proposal.<property>` multiple times in this function body, you can destructure it like:

```typescript
const {
  initiatorDepositTokenAddress,
  responderDepositTokenAddress,
  initialState,
  ...
}
```

LayneHaber:
FWIW this works on our side for erc20 transfers in virtual apps 👍

PR #2070, opened 2019-07-30
[node] use builder pattern

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR creates a `build` method on StateChannel that returns a new copy with selected fields overridden. It also splits out the functions `removeActiveApp` and `incrementFreeBalance`.

### DIFF
--- a/packages/node/src/models/state-channel.ts
+++ b/packages/node/src/models/state-channel.ts
@@ -212,40 +212,66 @@ export class StateChannel {
     return topLevelKey;
   }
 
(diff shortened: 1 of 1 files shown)


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
> Bump node version

wdym

cf19drofxots:
https://github.com/counterfactual/monorepo/blob/master/packages/node/package.json#L3

PR #2075, opened 2019-07-31
[node] Remove need for mnemonic, use extended private key

cf19drofxots:
This is a middle step to better implementing https://github.com/counterfactual/monorepo/pull/2034. That PR will introduce another execution path of using a _neutered_ extended key, and this PR incrementally paves the way for more consistent usage of extended keys throughout and minimizing usage of mnemonics.


packages/node/test/machine/integration/install-then-set-state.spec.ts
```diff
@@ -77,12 +80,11 @@ beforeAll(async () => {
  */
 describe("Scenario: install AppInstance, set state, put on-chain", () => {
   it("returns the funds the app had locked up for both ETH and ERC20", async done => {
-    const xkeys = getRandomHDNodes(2);
+    const xprivs = getRandomExtendedKeys(2);
```
snario: `xprv` vs `xpriv`

packages/node/test/machine/integration/random-signing-keys.ts
```diff
@@ -12,13 +13,22 @@ export function getSortedRandomSigningKeys(length: number) {
     );
 }
 
-export function getRandomHDNodes(length: number) {
-  // tslint:disable-next-line:prefer-array-literal
+export function getXPubsFromExtendedKeys(xkeys: string[]): string[] {
```
snario: I'd make `xkeyToXpub` a function and then instead of doing `getXPubsFromExtendedKeys(xkeys)` in the code, just do `xkeys.map(neuterExtendedKey)`

packages/node/test/machine/integration/random-signing-keys.ts
```diff
@@ -12,13 +13,22 @@ export function getSortedRandomSigningKeys(length: number) {
...
+  return getRandomExtendedKeys(length).map(
+    x => fromExtendedKey(x).neuter().extendedKey
+  );
+}
+
+export function getRandomExtendedKeys(length: number): string[] {
   return Array(length)
     .fill(0)
-    .map(_ => fromMnemonic(Wallet.createRandom().mnemonic));
+    .map(_ => computeRandomExtendedKey());
```
snario:
```suggestion
    .map(computeRandomExtendedKey);
```

packages/local-ganache-server/src/index.ts
```diff
@@ -24,7 +24,7 @@ export class LocalGanacheServer {
   server: any;
   networkContext: NetworkContextForTestSuite;
 
-  constructor(mnemonics: string[], initialBalance: string = "1000") {
+  constructor(extendedKeys: string[], initialBalance: string = "1000") {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: we should avoid the phrase "extended Key" because it doesn't specify private or public keys. I suggest this argument be named "xprvs"

packages/node/src/machine/xkeys.ts
```diff
@@ -10,6 +11,10 @@ import { fromExtendedKey, HDNode } from "ethers/utils/hdnode";
  * BIP-44 specifies that if the purpose is 44, then the format is "m / purpose' / cointype' / account' / change / index"
  */
 
+export function computeRandomExtendedKey(): string {
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `computeRandomXpriv`

packages/node/src/signer.ts
```diff
@@ -1,19 +1,24 @@
 import { Node } from "@counterfactual/types";
-import { Wallet } from "ethers";
-import { fromMnemonic, HDNode } from "ethers/utils/hdnode";
+import { fromExtendedKey, HDNode } from "ethers/utils/hdnode";
 
-export const MNEMONIC_PATH = "MNEMONIC";
+import { computeRandomExtendedKey } from "./machine/xkeys";
+
+export const EXTENDED_KEY_PATH = "EXTENDED_KEY";
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `EXTENDED_PRIVATE_KEY`

packages/node/src/machine/xkeys.ts
snario: I prefer ExtendedKey, or NonNeuteredExtendedKey

packages/local-ganache-server/src/index.ts
snario: In some cases though, we can operate on either one

packages/node/src/machine/xkeys.ts
cf19drofxots: Alright well these are opposing styles and we need to pick one. I'm leaning towards `xprv` simply because it's much shorter than `NonNeuteredExtendedKey`

packages/local-ganache-server/src/index.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: yes, but (unless I'm mistaken) this function must take an extended private key, not an extended public key

cf19drofxots: In the scope of this PR, that understanding is correct. The referenced PR changes this behavior though.

packages/node/src/machine/xkeys.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: to be clearer: "`createRandomExtendedKey`" is ambiguous because it doesn't specify whether the output will begin with `xpub` or `xprv` and it's important for the caller of this function to know. `createNonNeuteredExtendedKey` is not ambiguous, hence acceptable to me on that point. however the phrase "non-neutered extended key" is not commonly used AFAICT

I think using the terminology from https://bitcoin.org/en/wallets-guide#hierarchical-deterministic-key-creation is best, it has all the terms we need and doesn't use the word "neuter"

packages/local-ganache-server/src/index.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: you mean you can pass an array which contains both extended private keys and extended public keys as this argument? why would you do that?

packages/node/src/signer.ts
```diff
@@ -1,19 +1,24 @@
...
+  if (!extendedKey) {
+    extendedKey = computeRandomExtendedKey();
+    await storeService.set([{ key: EXTENDED_KEY_PATH, value: extendedKey }]);
   }
 
-  // 25446 is 0x6366... or "cf" in ascii, for "Counterfactual".
-  return fromMnemonic(mnemonic).derivePath("m/44'/60'/0'/25446");
+  try {
+    // 25446 is 0x6366... or "cf" in ascii, for "Counterfactual".
+    return fromExtendedKey(extendedKey).derivePath("m/44'/60'/0'/25446");
```
snario: Add "m/44'/60'/0'/25446" to constants file

packages/local-ganache-server/src/index.ts
cf19drofxots: Nope.

It's explained in the PR's description.

packages/node/src/machine/xkeys.ts
cf19drofxots: Inspired by the referenced guide, "extended {private,public} key" is used as of this PR.

packages/node/test/test-constants.jest.ts
```diff
@@ -1,8 +1,8 @@
-export const A_MNEMONIC =
-  "maximum guide cement kiwi lobster skate immense expand spot crew zoo ring";
+export const A_EXTENDED_KEY =
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I still think this shoud be `A_EXTENDED_PRIVATE_KEY` or `A_XPRIV`, etc

packages/node/README.md
```diff
@@ -20,6 +20,12 @@ Some specific examples of this include:
 
 We have [some diagrams](./docs/diagram.md) explaining the Node's architecture and control flow.
 
+## Node Address and Signing Keys
+
+The Node is expected to have access to an [extended key](https://bitcoin.org/en/glossary/extended-key) which it uses to derive a "public identifier" used as the address of a Node instance. This key is also used to produce private keys at app-specific derivation paths.
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: access to an extended public key (even the linked page says "extended key" is ambiguous)

packages/node/test/test-constants.jest.ts
cf19drofxots: oh good catch, it should be

packages/node/README.md
cf19drofxots: Updated README

PR #2078, opened 2019-07-31
[node] Add function to get supported asset types

cf19drofxots:


packages/node/src/node.ts
```diff
@@ -27,6 +27,10 @@ export interface NodeConfig {
   STORE_KEY_PREFIX: string;
 }
 
+export function getSupportedAssetTypes(): string[] {
+  return ["ETH", "ERC20"];
```
joelalejandro: @cf19drofxots We'd need also the token address for each asset.

cf19drofxots:
Misunderstood requirement.

PR #2083, opened 2019-07-31
[node] Remove unreasonable circular dependencies

cf19drofxots:


packages/node/src/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { AddressZero } from "ethers/constants";
```
cf19drofxots: We might not want this but having a place to start gathering all the constants across the codebase might be a reasonable idea.

packages/node/rollup.config.js
```diff
@@ -40,6 +40,27 @@ const external = [
   })
 ];
 
+const onwarn = warning => {
+  // Silence circular dependency warnings specifically for reasonable
```
snario: What is a reasonable circular dependency?

cf19drofxots: Updated the documentation

PR #2092, opened 2019-08-01
Handle asymmetric event emission + better concurrency handling

snario:
Concurrency fixes:

1. Install virtual app now has queuing. It did not before.
2. Uninstall virtual app now queues on both channels with counterparties, not just one.
3. Update and take action now queues on app instance identity hash, not multisig address

### DIFF
--- a/packages/node/src/message-handling/handle-protocol-message.ts
+++ b/packages/node/src/message-handling/handle-protocol-message.ts
@@ -17,9 +17,8 @@ import { StateChannel } from "../models";
 import { UNASSIGNED_SEQ_NO } from "../protocol/utils/signature-forwarder";
 import { RequestHandler } from "../request-handler";
 import RpcRouter from "../rpc-router";
-import { Store } from "../store";
 import { NODE_EVENTS, NodeMessageWrappedProtocolMessage } from "../types";
-import { bigNumberifyJson, hashOfOrderedPublicIdentifiers } from "../utils";
+import { bigNumberifyJson, getCreate2MultisigAddress } from "../utils";
 
 /**
  * Forwards all received NodeMessages that are for the machine's internal
@@ -48,7 +47,6 @@ export async function handleReceivedProtocolMessage(
   const queueNames = await getQueueNamesListByProtocolName(
     protocol,
     params,
-    store,
     requestHandler
   );
 
@@ -189,12 +187,22 @@ const getSetupEventData = (
 async function getQueueNamesListByProtocolName(
   protocol: string,
   params: ProtocolParameters,
-  store: Store,
   requestHandler: RequestHandler
 ): Promise<string[]> {
-  const { publicIdentifier } = requestHandler;
+  const { publicIdentifier, networkContext } = requestHandler;
+
+  function multisigAddressFor(xpubs: string[]) {
+    return getCreate2MultisigAddress(
+      xpubs,
+      networkContext.ProxyFactory,
+      networkContext.MinimumViableMultisig
+    );
+  }
 
   switch (protocol) {
+    /**
+     * Queue on the multisig address of the direct channel.
+     */
     case Protocol.Install:
     case Protocol.Setup:
     case Protocol.Uninstall:
@@ -207,12 +215,18 @@ async function getQueueNamesListByProtocolName(
 
       return [multisigAddress];
 
+    /**
+     * Queue on the appInstanceId of the AppInstance.
+     */
     case Protocol.TakeAction:
     case Protocol.Update:
       const { appIdentityHash } = params as TakeActionParams | UpdateParams;
 
       return [appIdentityHash];
 
+    /**
+     * Queue on the multisig addresses of both direct channels involved.
+     */
     case Protocol.InstallVirtualApp:
     case Protocol.UninstallVirtualApp:
       const { initiatorXpub, intermediaryXpub, responderXpub } = params as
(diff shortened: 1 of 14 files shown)


packages/node/src/message-handling/handle-protocol-message.ts
```diff
@@ -36,138 +37,213 @@ export async function handleReceivedProtocolMessage(
     router
   } = requestHandler;
 
-  const deserializedMsg = bigNumberifyJson(msg);
-
-  const { data } = deserializedMsg;
+  const { data } = bigNumberifyJson(msg) as NodeMessageWrappedProtocolMessage;
```
cf19drofxots: Based on offline discussion we should avoid type casting like this.

snario: I have to type case in this case because `bigNumberifyJson` destroys type information.

cf19drofxots:
Can you describe in the PR's description how this achieves better concurrency?

snario:
> Can you describe in the PR's description how this achieves better concurrency?

see above

packages/node/src/methods/app-instance/uninstall-virtual/controller.ts
```diff
@@ -50,16 +50,14 @@ export default class UninstallVirtualController extends NodeController {
   }
 
   protected async beforeExecution(
+    // @ts-ignore
```
cf19drofxots: The `RequestHandler` should be optional on the `beforeExecution` fn signature to avoid having to ignore it like this.

packages/node/src/methods/controller.ts
```diff
@@ -4,35 +4,21 @@ import { Controller } from "rpc-server";
...
-    let ret;
-
-    if (shardedQueues.length > 0) {
-      for (const queue of shardedQueues) queue.add(executeCached);
-      for (const queue of shardedQueues) await queue;
-      ret = await promise;
-    } else {
-      ret = await executeCached();
-    }
+    const ret = await executeFunctionWithinQueues(
```
cf19drofxots: Better variable name.

packages/node/src/message-handling/handle-protocol-message.ts
```diff
@@ -36,138 +37,213 @@ export async function handleReceivedProtocolMessage(
...
+
+      return [appIdentityHash];
+
+    case Protocol.InstallVirtualApp:
+    case Protocol.UninstallVirtualApp:
+      const { initiatorXpub, intermediaryXpub, responderXpub } = params as
+        | UninstallVirtualAppParams
+        | InstallVirtualAppParams;
+
+      if (publicIdentifier === intermediaryXpub) {
```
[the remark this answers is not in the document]
snario: That is not possible as the initiator never receives a prototol message here

packages/node/src/message-handling/handle-protocol-message.ts
```diff
@@ -36,138 +37,213 @@ export async function handleReceivedProtocolMessage(
...
+        )
+      };
+    default:
+      throw new Error(
+        `handleReceivedProtocolMessage received invalid protocol message: ${protocol}`
       );
+  }
+}
+
+const getStateUpdateEventData = (
```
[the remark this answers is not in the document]
snario: Code minimization; no margin for adding mutation or anything, just returns precisely one thing in a very simple statement

packages/node/src/message-handling/handle-protocol-message.ts
```diff
@@ -36,138 +37,213 @@ export async function handleReceivedProtocolMessage(
...
+  protocol: string,
+  params: ProtocolParameters,
+  publicIdentifier: string,
+  stateChannelsMap: Map<string, StateChannel>
+) {
+  const baseEvent = { from: publicIdentifier };
+
+  switch (protocol) {
+    case Protocol.Install:
+      // TODO: Have to take an InstallParams object and somehow compute the
```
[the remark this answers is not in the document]
snario: No

packages/node/src/methods/app-instance/uninstall-virtual/controller.ts
snario: Doesn't actually help because `beforeExecution` gets called by the super class implementation of `enqueueByShard`, but I agree something might be better

packages/node/src/methods/controller.ts
snario: There actually isn't a better name IMO, it's literally an `any` type response

packages/node/src/message-handling/handle-protocol-message.ts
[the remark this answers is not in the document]
cf19drofxots: Let's make a note of that here to provide more clarity for others with less _immediate_ context.

packages/node/src/message-handling/handle-protocol-message.ts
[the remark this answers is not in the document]
cf19drofxots: An implementation with the function syntax wouldn't add mutation either. I'm just advocating for that to keep the style consistent, not because I think what this is doing is not simple.

packages/node/src/message-handling/handle-protocol-message.ts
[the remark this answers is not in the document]
cf19drofxots: Made an issue for these: https://github.com/counterfactual/monorepo/issues/2108

packages/node/src/methods/app-instance/uninstall-virtual/controller.ts
cf19drofxots: https://github.com/counterfactual/monorepo/issues/2109

packages/node/src/methods/controller.ts
cf19drofxots: Wasn't referring to the type, just the name. Not a blocker.

packages/node/test/integration/uninstall-with-postgres.spec.ts
```diff
@@ -37,6 +37,9 @@ describe("Node method follows spec - uninstall", () => {
 
       await createChannel(nodeA, nodeB);
 
+      // FIXME: There is some timing issue with postgres @snario noticed
```
cf19drofxots: https://github.com/counterfactual/monorepo/issues/2110

cf19drofxots:
Bump package versions to pick up the new changes.

PR #2094, opened 2019-08-01
[node] Replace `free-balance` API with `FreeBalanceClass`

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Previously, the helper file `free-balance.ts` exported 6 data structures and 5 functions:

### Data Structures

- `FreeBalanceState`
- `FreeBalanceStateJSON`
- `CoinTransfer`
- `CoinTransferJSON`
- `CoinTransferMap`
- `TokenIndexedCoinTransferMap`

### Functions

- `getBalancesFromFreeBalanceAppInstance`
- `deserializeFreeBalanceState`
-  `serializeFreeBalanceState`
- `convertCoinTransfersMapToCoinTransfers` 
- `convertCoinTransfersToCoinTransfersMap`

This PR creates a new class, `FreeBalanceClass`, with a smaller conversion API - it has 1 method each to convert to and from from an `AppInstance`, and no other methods should be used for conversion. This allows us to de-export most of the data structures and conversion functions listed above.

### DIFF
--- a/packages/node/src/ethereum/utils/free-balance-app.ts
+++ b/packages/node/src/ethereum/utils/free-balance-app.ts
@@ -1,10 +1,9 @@
 import { AppInterface } from "@counterfactual/types";
 import { Zero } from "ethers/constants";
-import { BigNumber, defaultAbiCoder } from "ethers/utils";
+import { BigNumber } from "ethers/utils";
 
 import {
   CoinTransferMap,
-  FreeBalanceStateJSON,
   TokenIndexedCoinTransferMap
 } from "../../models/free-balance";
 
@@ -25,10 +24,6 @@ export function getFreeBalanceAppInterface(addr: string): AppInterface {
     stateEncoding: freeBalanceAppStateEncoding,
     actionEncoding: undefined // because no actions exist for FreeBalanceApp
   };
-}
-
-export function encodeFreeBalanceAppState(state: FreeBalanceStateJSON) {
-  return defaultAbiCoder.encode([freeBalanceAppStateEncoding], [state]);
 }
 
 export function flipTokenIndexedBalances(
@@ -59,8 +54,8 @@ export function flip(coinTransferMap: CoinTransferMap): CoinTransferMap {
 /**
  * Returns the first base mapping, but incremented by values specified in the
  * second increment. Passing increments whose keys are not present in the base
- * is an error. Keys in the base mapping which are not explicitly incremented
- * are returned unchanged.
+ * sets them to the increment. Keys in the base mapping which are not explicitly
+ * incremented are returned unchanged.
  */
 export function merge(
   base: { [s: string]: BigNumber },
@@ -68,22 +63,13 @@ export function merge(
 ) {
   const ret = {} as { [s: string]: BigNumber };
 
-  for (const key of Object.keys(base)) {
-    if (increments[key]) {
-      ret[key] = base[key].add(increments[key]);
-      if (ret[key].lt(Zero)) {
-        throw new Error("Underflow in merge");
-      }
-    } else {
-      ret[key] = base[key];
-    }
-  }
+  const s1 = new Set(Object.keys(base));
+  const s2 = new Set(Object.keys(increments));
 
-  for (const key of Object.keys(increments)) {
-    if (!base[key]) {
-      throw new Error(
-        `mismatch: ${Object.keys(increments)} ⊄ ${Object.keys(base)}`
-      );
+  for (const key of new Set([...s1, ...s2])) {
+    ret[key] = (base[key] || Zero).add(increments[key] || Zero);
+    if (ret[key].lt(Zero)) {
+      throw new Error("Underflow in merge");
     }
   }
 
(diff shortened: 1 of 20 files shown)


packages/node/src/methods/app-instance/get-free-balance/controller.ts
```diff
@@ -36,15 +31,14 @@ export default class GetFreeBalanceController extends NodeController {
 
     const stateChannel = await store.getStateChannel(multisigAddress);
 
-    const freeBalanceState = deserializeFreeBalanceState(stateChannel
-      .freeBalance.state as FreeBalanceStateJSON);
+    const ret = stateChannel
```
joelalejandro: Can we please maintain concrete, specific variable names? `ret` feels a bit too generic, IMHO.

joelalejandro: Sorry, just saw this was in WIP, didn't mean to early-review :sweat_smile:

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: What do you suggest the name be? I use `ret` if the string `const ret =` can be replaced with `return` (except for the error handling after)

packages/node/src/ethereum/utils/free-balance-app.ts
```diff
@@ -59,31 +54,22 @@ export function flip(coinTransferMap: CoinTransferMap): CoinTransferMap {
...
 
-  for (const key of Object.keys(increments)) {
-    if (!base[key]) {
-      throw new Error(
-        `mismatch: ${Object.keys(increments)} ⊄ ${Object.keys(base)}`
-      );
+  for (const key of new Set([...s1, ...s2])) {
+    ret[key] = (base[key] || Zero).add(increments[key] || Zero);
+    if (ret[key].lt(Zero)) {
+      throw new Error("Underflow in merge");
```
cf19drofxots: Since this error is very specific to this context, can it be more descriptive about said context?

i.e. if this is thrown, a Node operator is unsure of what is being merged and furthermore is unsure of where to start with investigating an "underflow" (another term that could use updating, maybe mentioning relative to 0 or something?)

packages/node/src/models/free-balance.ts
```diff
@@ -52,14 +55,122 @@ export type TokenIndexedCoinTransferMap = {
...
 };
 
+// todo(xuanji): replace with Set
 export type ActiveAppsMap = { [appInstanceIdentityHash: string]: true };
 
-export type FreeBalanceState = {
+export class FreeBalanceClass {
+  private constructor(
+    private readonly activeAppsMap: ActiveAppsMap,
+    private readonly balancesIndexedByToken: {
```
cf19drofxots: Why not use `tokenIndexedBalances`?

packages/node/src/models/free-balance.ts
```diff
@@ -52,14 +55,122 @@ export type TokenIndexedCoinTransferMap = {
...
+      [tokenAddress: string]: CoinTransfer[];
+    }
+  ) {}
+  public toFreeBalanceState(): FreeBalanceState {
+    return {
+      activeAppsMap: this.activeAppsMap,
+      balancesIndexedByToken: this.balancesIndexedByToken
+    };
+  }
+  public toTokenIndexedCoinTransferMap() {
```
cf19drofxots: Maybe `reduce` here?

packages/node/src/models/free-balance.ts
```diff
@@ -52,14 +55,122 @@ export type TokenIndexedCoinTransferMap = {
...
+      freeBalanceState.balancesIndexedByToken
+    );
+  }
+  public getBalance(tokenAddress: string, beneficiary: string) {
+    try {
+      return convertCoinTransfersToCoinTransfersMap(
+        this.balancesIndexedByToken[tokenAddress]
+      )[beneficiary];
+    } catch {
+      return Zero;
```
cf19drofxots: Maybe not entirely necessary but to be explicit it could be useful to clarify in the API doc that calling `getFreeBalance` for a specific token address will (intuitively) return `Zero` even if the channel hasn't ever had any deposits/withdrawals of that token.

packages/node/src/models/free-balance.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I will probably do this in a followup PR, but this class was meant to replicate `FreeBalanceState` since that was the closest thing to a good data structure being passed around

packages/node/src/models/free-balance.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: I dislike our use of `reduce`

packages/node/src/models/free-balance.ts
cf19drofxots: Ok

packages/node/src/models/free-balance.ts
cf19drofxots: I'm indifferent whether we use `reduce` or not, I just prefer a consistent style.

packages/node/src/ethereum/utils/free-balance-app.ts
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: moved to caller

PR #2105, opened 2019-08-02
[node] Remove owners hash DB field and related functions

snario:


packages/node/src/methods/app-instance/propose-install/controller.ts
```diff
@@ -66,17 +67,12 @@ export default class ProposeInstallController extends NodeController {
...
 
-    const multisigAddress = await store.getMultisigAddressFromOwnersHash(
-      hashOfOrderedPublicIdentifiers([myIdentifier, proposedToIdentifier])
+    const multisigAddress = getCreate2MultisigAddress(
+      [myIdentifier, proposedToIdentifier],
+      networkContext.ProxyFactory,
+      networkContext.MinimumViableMultisig
     );
 
-    // TODO: DRY this at top level of most calls that query a channel
```
cf19drofxots: Make an issue to address this and leave it here until the issue implements this better. Removing this makes the software possibly more error-prone, not less.

packages/node/src/methods/app-instance/propose-install-virtual/operation.ts
```diff
@@ -92,13 +91,12 @@ export async function getOrCreateStateChannelBetweenVirtualAppParticipants(
...
       responderXpub,
+      network,
       store
     );
   } catch (e) {
     if (
-      e
-        .toString()
-        .includes(NO_CHANNEL_BETWEEN_NODES(initiatorXpub, responderXpub)) &&
+      e.toString().includes("Call to getStateChannel failed") &&
```
cf19drofxots: Remove the use of string literals when its value is expected to appear in multiple places, makes comparison less error-prone.

PR #2111, opened 2019-08-02
Ensure token balances are withdraw-able from free balance

snario:


packages/cf-funding-protocol-contracts/contracts/ConditionalTransactionDelegateTarget.sol
```diff
@@ -35,22 +40,35 @@ contract ConditionalTransactionDelegateTarget {
...
-    bytes memory payload = abi.encodeWithSignature(
-      "interpretOutcomeAndExecuteEffect(bytes,bytes)",
-      abi.encode(outcome),
-      multiAssetMultiPartyCoinTransferInterpreterParams
+    uint256[] memory limits = new uint256[](
+      freeBalanceAppState.tokenAddresses.length
     );
 
+    for (uint256 i = 0; i < freeBalanceAppState.tokenAddresses.length; i++) {
+      limits[i] = MAX_UINT256;
```
cf19drofxots: The limits might not be the max limit though?

snario: For the free balance app specifically, it is gifted the permission to have no limits (because its values changed frequently)

cf19drofxots: I mean the limits are known client-side, so why can't they be passed in along with this other data?

Maybe I'm wrong, but this seems like it relaxes security for convenience when it's not that hard to avoid doing so.

snario: The transaction's interpreter parameters are determined at the time of creation of the free balance; hence we cannot know how much will be deposited into it all-time. So, we either rely on the app state (unsafe), or we just give it full permissions.

cf19drofxots: If the limit is the _max_ allowed limit, how is relying on app state any less safe?

In either case, a comment around this code would be helpful to clarify this reasoning. Otherwise the setting the limit to this max without context is a bit strange.

Issue #1852, opened 2019-07-12
[wallet-ui] Features and test suites


joelalejandro:
After work on #1713, the idea is to finish migrating some pending features from the Playground, that are relevant to the Wallet UI experience. 

### Features to add
- [ ] Exchange screen
  - [ ] Deposit form: _This is the same as the onboarding deposit flow, except there's no multisig deployment, it's just a regular deposit to a pre-existing multisig._
  - [ ] Withdraw form: _This is the opposite of the deposit operations, it transfers funds from the CF account to the Ethereum wallet._

#### State channels screen
> InVision: https://projects.invisionapp.com/share/9PQ8DASYM3W#/screens/361728445

The _State channels_ screen shows a tree view containing three types of leaves: the Hub (aka `simple-hub-server` or `connext`), the User (a real person or a bot), and the Dapp itself.

There are some actions a user can perform:

- *Launch*: An user can re-open a Dapp if it has a state channel open with it.
  - [ ] How do we know what's its URL? Where do we store that information?
- *Debug*: An advanced user (such as a developer) could want to inspect the state of the channel.
  - [ ] Should we render the output of a `GET_STATE` operation here?
  - [ ] What data is relevant to present? How can we arrange it properly?
  - [ ] Should this be our own version of an "inspector" feature?
- *Uninstall*: An user can end the state channel by uninstalling it, triggering the `UNINSTALL` operation.

### Testing

Learning from previous mistakes, we're building tests for every component and looking into integration flows. Now that we're working with React, we have better supported tools to do this. With this in mind, we can resume the PoC we did a couple of months ago with the WebDriver that auto-configures the Metamask extension, allowing us to test a dapp interacting with our Wallet.

_This list will change as we analyze in more detail what we need to test_.

#### Test suites

_By component/module_
- [ ] Root-level App component
- [ ] UI components
  - [ ] AccountContext
  - [ ] ChannelNode
  - [ ] ChannelTree
  - [ ] FormButton
  - [ ] FormInput
  - [ ] LayoutHeader
  - [ ] WidgetCard
  - [ ] WidgetErrorMessage
  - [ ] WidgetHeader
  - [ ] WidgetLogo
  - [ ] WidgetScreen
  - [ ] WidgetTooltip
- [ ] Store modules
  - [ ] Channels
  - [ ] User
  - [ ] Wallet
- [ ] Utils
- [ ] Hub API Client

_By scenario_
[...]

PR #2134, opened 2019-08-05
[greenboard] Adds Login scenario

joelalejandro:
This PR implements the following scenario:

> A registered user should be able to login to their account and see their CF balance.

Contributes to #1852.


joelalejandro:
This test is currently unstable, apparently the StateCollector isn't persisting correctly the LocalStorage from one session to the next.

joelalejandro:
Test has been stabilized.

PR #2102, opened 2019-08-02
[Wallet-UI] ERC20 Token Support

spersico:
### Description

This PR, when merged will add support for withdraw and deposit transactions inside the wallet.
The necessary middleware modifications have been added in the MM integration branch.

There are some details to polish, but deposits and withdraws behave like ETH now, and the balance component now shows the current state of the balance for all currencies.

### Related issues
closes #1955

PR #2135, opened 2019-08-05
[wip] [greenboard] Adds Deposit scenario

joelalejandro:
This PR implements the following scenario:

> A registered user should be able to make deposits.

Currently it's not working due to an error in the "Deposit" screen.


joelalejandro:
Waiting on #2102 to be merged to prevent test selector conflicts.

joelalejandro:
@spersico can you pick up where I left off on this PR? It'd be great to add a second flow where the user attempts to deposit with an ERC20 token instead of ETH.

spersico:
Yes! Sure!

PR #2142, opened 2019-08-07
[node] Withdrawal confirmed event

LayneHaber:
Emit a withdrawal confirmed event.

Would like a confirmation on types emitted here, I just went with what was easiest.


cf19drofxots:
> Would like a confirmation on types emitted here, I just went with what was easiest.

seems good

PR #2145, opened 2019-08-07
[monorepo] Fixes CI configuration

joelalejandro:
Checks are getting stuck with `run-tests` not returning a result. This PR attempts to fix that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,3 +161,27 @@ workflows:
           filters:
             branches:
               only: master
+
+      - heroku/deploy-via-git:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -161,27 +161,3 @@ workflows:
           filters:
             branches:
               only: master
-
-      - heroku/deploy-via-git:
-          app-name: simple-hub-server-staging
```
cf19drofxots: Is this also being extracted?

joelalejandro: Oops! These weren't meant to be removed, reverting.

PR #2146, opened 2019-08-07
[node] Withdraw Commitment API

cf19drofxots:
## **Problem**

The current CF Node API for performing a withdraw restricts the call such that the withdrawal made on a channel is made from the free balance of the party initiating the withdrawal.

This prevents the ability for two parties to unanimously agree on a withdraw such that the counter party of the party doing the withdrawal is the one submitting the withdrawal transaction to the network and incurring the gas costs of this action.

## Desired Effect

A party within a channel should be able to submit a counter party's withdrawal transaction to the network so the party withdrawing from the channel can withdraw without spending gas. This is ultimately a UX improvement, particularly useful in the context of parties being in channels with a hub, where the hub incurs the cost of the withdrawal gas when the parties want to withdraw.

This PR introduces an API method to reflect the desired effect and generate a withdraw commitment without sending it to the network.

The following code snippet of the test case demonstrates expected functionality: 
https://github.com/counterfactual/monorepo/blob/a3c37c3af75703707dcb14a8ff6b02233edb7faa/packages/node/test/integration/secure-withdraw.spec.ts#L165-L184

Note: we also desire that a deposit not be able to occur during a withdraw commitment production, but implementing this restriction is not within the scope of this PR.


packages/types/src/node.ts
```diff
@@ -28,6 +28,15 @@ export namespace Node {
     type: EventName;
   };
 
+  // This is used instead of the ethers `Transaction` because that type
+  // requires the nonce and chain ID to be specified, when sometimes those
+  // arguments are not known at the time of creating a transaction.
+  export type MinimalTransaction = {
```
snario: Aren't all properties optional on the ethers transaction?

packages/types/src/node.ts
```diff
@@ -115,7 +125,8 @@ export namespace Node {
     TAKE_ACTION = "chan_takeAction",
     UNINSTALL = "chan_uninstall",
     UNINSTALL_VIRTUAL = "chan_uninstallVirtual",
-    WITHDRAW = "chan_withdraw"
+    WITHDRAW = "chan_withdraw",
+    WITHDRAW_COMMITMENT = "chan_withdraw_commitment"
```
snario: Does not follow camelcase pattern

cf19drofxots:
> I don't see any locking mechanism on deposits and withdrawals until the withdrawal commitment has been observed to occur on-chain

From description:
> Note: we also desire that a deposit not be able to occur during a withdraw commitment production, but implementing this restriction is not within the scope of this PR.

packages/types/src/node.ts
cf19drofxots: Apparently not.

packages/types/src/node.ts
cf19drofxots: Fixed

cf19drofxots:
@snario any other comments?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
I assume the new method will return a transaction with `to` set to the multisig address, with the `data` encoding a call to the `execTransaction` method? If so can we specify that in the API doc?

I also think a cleaner / more debuggable API would be to return the arguments to `execTransaction` (i.e., return `(address to,uint256 value,bytes memory data,Operation operation,bytes[] signatures)`)

packages/node/test/integration/utils.ts
```diff
@@ -190,6 +201,24 @@ export function makeDepositRequest(
   });
 }
 
+export function makeWithdrawCommitmentRequest(
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: `request` is not defined as part of the API, this should be called `makeWithdrawCommitmentRpc`

cf19drofxots: Agree this could be more accurately named, but `makeWithdrawCommitmentRpc` would also be inaccurate because the call itself isn't being made by this function.

If anything, this should be named `constructWithdrawCommitmentRpc` because the function is constructing the RPC to be made.

cf19drofxots:
> I assume the new method will return a transaction with to set to the multisig address, with the data encoding a call to the execTransaction method? If so can we specify that in the API doc?

Yeah. The doc doesn't mention it with this much detail because that's more of an implementation detail, whereas the API caller only cares about the effect of the call, right?

> I also think a cleaner / more debuggable API would be to return the arguments to execTransaction (i.e., return (address to,uint256 value,bytes memory data,Operation operation,bytes[] signatures))

The caller only cares about receiving a transaction to pass off (as described in the PR description). The use case of just sending the transaction seems more frequent than wanting to debug it, while retaining the ability to debug it by just decoding the `data` field.

packages/node/test/integration/utils.ts
cf19drofxots: I'll do the renaming in a separate PR to reduce the noise here.

cf19drofxots:
Merging as this has been reviewed and comments addressed. If any follow up comments remain they can be addressed as a follow up.

packages/node/test/integration/utils.ts
cf19drofxots: https://github.com/counterfactual/monorepo/pull/2175

PR #2150, opened 2019-08-08
[node] do getAddress on token addresses

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
There are multiple controllers that check a certain token address and default it to ETH if it is falsy.
- Create a helper function to do that
- Do a `ethers.getAddress` normalization in addition

### DIFF
--- a/packages/node/src/methods/app-instance/get-free-balance/controller.ts
+++ b/packages/node/src/methods/app-instance/get-free-balance/controller.ts
@@ -1,7 +1,6 @@
 import { Node } from "@counterfactual/types";
 import { jsonRpcMethod } from "rpc-server";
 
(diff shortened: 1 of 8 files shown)


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Why two functions? These are always done together

snario:
Because they are distinct operations; one is for a default parameter on the RPC of the Node and the other is argument sanitization

cf19drofxots:
The `normalization` can still happen while accommodating for validity checksum via a behavior like `normalizeTokenAddressOrThrow` where it's expected that it'll throw if the token address format is invalid.

So perhaps all that's missing is the rename and a try/catch around the `getAddress` call?

PR #1001, opened 2019-03-11
[wip][metamask-integration] Post Messages to MetaMask

Alonski:
### Description

This implements messaging to and from MetaMask background script for use with the Node and the Playground UI running inside of MetaMask,

### Related issues

- [ ] Deploy preview is functional

PR #2151, opened 2019-08-08
[node-provider] Adds NodeProviderEthereum

joelalejandro:
This is an extraction + enhancement PR based on #1001.

It'll add the `NodeProviderEthereum` class to the `NodeProvider` package, as per @Alonski's original commits, plus adding some better types + unit tests for the new class.

### DIFF
--- a/packages/node-provider/src/node-provider-ethereum.ts
+++ b/packages/node-provider/src/node-provider-ethereum.ts
@@ -103,7 +103,7 @@ export default class NodeProviderEthereum implements INodeProvider {
     }
 
     ethereum
-      .send("counterfactual:nodeProvider", [message])
+      .send("counterfactual:nodeProvider:request", [message])
       .then(({ result }: { result: Node.Message }) => {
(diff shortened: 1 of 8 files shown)


packages/node-provider/src/node-provider-ethereum.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    this.startEthereumEventListeners();
+    this.notifyNodeProviderIsConnected();
+
+    return Promise.resolve(this);
+  }
+
+  private startEthereumEventListeners() {
+    const NODE_EVENTS = [
```
snario: Could these be pulled from somewhere else?

joelalejandro: Maybe, but it's hardly mission-critical at the moment. Right now the purpose of breaking up #1001 is to try to figure out what's broken between this, Metamask, the Hub, the bots and the dapps. I'm happy to leave a comment in the code to revisit this later.

PR #2153, opened 2019-08-09
[cf.js] Allows usage of NodeProviderEthereum

joelalejandro:
Extracted out of #1001.

### DIFF
--- a/packages/cf.js/src/index.ts
+++ b/packages/cf.js/src/index.ts
@@ -1,4 +1,6 @@
-import NodeProvider from "@counterfactual/node-provider";
+import NodeProvider, {
+  NodeProviderEthereum
+} from "@counterfactual/node-provider";
 
 import { AppFactory } from "./app-factory";
 import { Provider } from "./provider";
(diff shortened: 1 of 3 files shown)


packages/cf.js/src/index.ts
```diff
@@ -1,4 +1,6 @@
-import NodeProvider from "@counterfactual/node-provider";
+import NodeProvider, {
```
cf19drofxots: Seems like a weird pattern to import these from another package then export them?

joelalejandro: That's because we want CF.js users _not_ to install NodeProvider. CF.js bundles this package so any dapp developers only need to install CF.js instead of both packages. It's for better developer ergonomics, IMO.

cf19drofxots: In that case, what's the disadvantage of folding the node-provider code to be a subset of cf.js package?

joelalejandro: That would prevent potential reuse of this package from `cf-wallet.js`.

cf19drofxots: Hmm fair enough. I'll put up a PR to introduce a README for the `node-provider` package in that case to clarify this pattern.

Issue #2172, opened 2019-08-12
node: FreeBalance check returns 0 when non-zero


bohendo:
Think there's a reversion in latest CF code, getting the following when trying to swap eth -> dai:
```
Uncaught (in promise) Error: Error: Node with public identifier xpub6E3tjd9js7QMrBtYo7f157D7MwauL6MWdLzKekFaRBb3bvaQnUPjHKJcdNhiqSjhmwa6TcTjV1wSDTgvz52To2ZjhGMiQFbYie2N2LZpNx6 has insufficient funds in channel 0xa5017a0949aF5264D86575F43a1db92f4c2ceebA
  for token 0xeec918d74c746167564401103096d45bbd494b74 to allocate towards an AppInstance. Current free balance for token is 0,
  attempted allocation amount: 9179640000000000000 
```
Even though the hub (xpub id ends with `...Nx6` btw) logs the following before trying to swap:
```
[Nest] 66   - 08/12/2019, 10:19 AM   [ChannelService] Collateralizing 0xa5017a0949aF5264D86575F43a1db92f4c2ceebA with 9179640000000000000, token: 0xeec918d74c746167564401103096d45bbd494b74
```
And the client, before swapping logs:
```
Available collateral: 9.17964 tokens, 9.17964 needed
```

Using: `@counterfactual/node@0.2.42` and `@counterfactual/types/0.0.32`
W same connext code, this error does not appear when using `@counterfactual/node@0.2.41` and `@counterfactual/types/0.0.31`

PS @rhlsthrm noticed the same error while trying to do ERC20 transfers

PR #2174, opened 2019-08-12
[node] add normalization in deserializeFreeBalanceState

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Fixes https://github.com/counterfactual/monorepo/issues/2172


cf19drofxots:
We should bump the version so the fix can be immediately consumed

PR #2178, opened 2019-08-13
[node] add customData field to ProtocolMessage

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
`ProtocolMessage` had two issues:

- `signature{1,2,3,4}` are not meaningful labels used by the spec
- the install-virtual-app protocol was misusing one of the signatures to pass data

This PR fixes both issues by creating a `customData` field with arbitrary keys allowed and mechanically moving `signature{1,2,3,4}` into it. A follow-up PR will give the signatures better names.

### DIFF
--- a/packages/node/src/machine/instruction-executor.ts
+++ b/packages/node/src/machine/instruction-executor.ts
@@ -109,7 +109,8 @@ export class InstructionExecutor {
       protocol: protocolName,
       protocolExecutionID: uuid.v1(),
       seq: 0,
-      toXpub: params[firstRecipientFromProtocolName(protocolName)]
+      toXpub: params[firstRecipientFromProtocolName(protocolName)],
+      customData: {}
     });
   }
 
@@ -123,7 +124,8 @@ export class InstructionExecutor {
         params,
         protocolExecutionID: uuid.v1(),
         seq: 0,
-        toXpub: params.responderXpub
+        toXpub: params.responderXpub,
+        customData: {}
       }
     );
   }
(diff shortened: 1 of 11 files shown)


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
It is also used to pass some non-signature data: https://github.com/counterfactual/monorepo/pull/2178/files#diff-04891d9781c7500549ac5219b5551ad1R112

cf19drofxots:
> It is also used to pass some non-signature data: https://github.com/counterfactual/monorepo/pull/2178/files#diff-04891d9781c7500549ac5219b5551ad1R112

oh jeez, that's so hacky...

Maybe it makes sense to have a signatures field in the `ProtocolMessage` struct and have another `metadata` field to pass other data?

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
I don't see why `signatures` and another non-singnature field should be separated

cf19drofxots:
They don't necessarily need to be. It's just one approach to structure the data in a less hacky way. The other approach is naming the fields more purposefully. A slight bias towards having `signatures` be its own field is because all protocols have them, so it's less of an arbitrary data struct being passed around (as opposed to the case for the other data you pointed to).

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
They all have them, but for each protocol the signatures mean different things, so I don't think they should be placed together

cf19drofxots:
Hmm ok. In that case this reasoning should be noted where `customData` gets used because at first glance (similar to my thinking process) seems like the signatures can be grouped.

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
added a comment to `customData`

PR #2182, opened 2019-08-14
[node] IStoreService: rename key to path

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This brings the interface in line with Firebase terminology and with how we actually treat variables with this name

### DIFF
--- a/packages/firebase-client/src/index.ts
+++ b/packages/firebase-client/src/index.ts
@@ -156,15 +156,15 @@ class FirebaseStoreService implements Node.IStoreService {
     private readonly storeServiceKey: string
   ) {}
 
-  async get(key: string): Promise<any> {
+  async get(path: string): Promise<any> {
     let result: any;
     await this.firebase
       .ref(this.storeServiceKey)
-      .child(key)
+      .child(path)
       .once("value", (snapshot: firebase.database.DataSnapshot | null) => {
         if (snapshot === null) {
           console.debug(
-            `Failed to retrieve value at ${key}: received a "null" snapshot`
+            `Failed to retrieve value at ${path}: received a "null" snapshot`
           );
           return;
         }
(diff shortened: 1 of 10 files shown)


cf19drofxots:
Seems like this broke master but the fix should be easy?

```
remote:        src/node.ts(236,7): error TS2345: Argument of type 'import("/tmp/build_26031d705add4dfcfa39b6758d16c2bd/node_modules/@counterfactual/types/dist/src/node").Node.IStoreService' is not assignable to parameter of type 'import("/tmp/build_26031d705add4dfcfa39b6758d16c2bd/node_modules/@counterfactual/node/node_modules/@counterfactual/types/dist/src/node").Node.IStoreService'.        
remote:          Types of property 'set' are incompatible.        
remote:            Type '(pairs: { path: string; value: any; }[], allowDelete?: Boolean | undefined) => Promise<void>' is not assignable to type '(pairs: { key: string; value: any; }[], allowDelete?: Boolean | undefined) => Promise<void>'.        
remote:              Types of parameters 'pairs' and 'pairs' are incompatible.        
remote:                Type '{ key: string; value: any; }[]' is not assignable to type '{ path: string; value: any; }[]'.        
remote:                  Property 'path' is missing in type '{ key: string; value: any; }' but required in type '{ path: string; value: any; }'
```

circleci.com/gh/counterfactual/monorepo/41332

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
I think it's from the old version of types

cf19drofxots:
But it also broke on master after the types PR was merged

https://www.circleci.com/gh/counterfactual/monorepo/41326

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
That job runs after publish to npm

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
wait, I am wrong...

cf19drofxots:
This is now causing tests to fail on other branches not relevant to this change

https://circleci.com/gh/counterfactual/monorepo/41357

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
that circleci task shouldn't be producing a `packages/simple-hub-server/node_modules/@counterfactual/node/node_modules/@counterfactual/types` folder

cf19drofxots:
I know, found that to be odd

PR #2188, opened 2019-08-15
[greenboard] Uses node-gyp-powered sqlite3 if no sqlite3 binary is installed in the system

### LINKED ISSUE #2139: [testing] Add fallback in case sqlite3 cli is not installed
## Description
Currently if the user doesn't have sqlite3 installed, the greenboard is not able to clean the DB automatically, and after the first test, all tests start to fail (because the wallet address already has a related user).

It would be great that this is not required, and that we simply have a clean copy of the DB each time that we run the tests.

joelalejandro:
Resolves #2139.

### DIFF
--- a/e2e-run.sh
+++ b/e2e-run.sh
@@ -2,27 +2,29 @@
 
 echo "(1/4) Running preflight checks and cleanups..."
 
-if [ -z "$NVM_DIR" ];
-then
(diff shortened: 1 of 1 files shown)


spersico:
Looking good! Seems ok. I'll test it tomorrow, now go to sleep!

PR #2189, opened 2019-08-15
[greenboard] Fallbacks to $(which python2.7) if no PYTHON env var exists

### LINKED ISSUE #2138: [testing] Replace $PYTHON variable
## Description

During the setup process of the E2E suite, the Greenboard E2E project asks the user for the $PYTHON environment variable.
I believe that we should use something more general, like `which python` or `which python2.7`, as the $PYTHON var it's not pre-installed in Ubuntu (and consequentially none of the related distros).

joelalejandro:
Resolves #2138.

### DIFF
--- a/e2e-prepare.sh
+++ b/e2e-prepare.sh
@@ -6,11 +6,18 @@ echo "(1/7) Running preflight checks and cleanups..."
 
 if [ -z "$PYTHON" ];
 then
-  echo "  > Error: Please set the PYTHON environment variable."
-  exit
+  echo "  > Python env-var not found, attempting detection via which"
+  PYTHON="$(which python2.7)"
+fi
+
+if [ -z "$PYTHON" ];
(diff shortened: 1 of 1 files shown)


e2e-prepare.sh
```diff
@@ -6,12 +6,19 @@ echo "(1/7) Running preflight checks and cleanups..."
...
+fi
+
+if [ -z "$PYTHON" ];
+then
+  echo "  > ERROR: You need to install python2.7 to compile node-gyp related dependencies."
 else
   echo "  > Found Python at: $PYTHON"
 fi
 
+exit
```
spersico: Didn't run this yet but.... Are you sure that this goes here and not in line 15 inside the*then*?

joelalejandro: This shouldn't be here, lol.

Issue #2016, opened 2019-07-26
[node] Show zero free balance instead of error for token without free balance


rhlsthrm:
Right now, we get an error when we query free balance for a token that does not have any: `Error: No free balance exists for the specified token: 0xdDA6327139485221633A1FcD65f4aC932E60A2e1`

Instead of an error that we have to catch on the caller side, it would be more beneficial to display a zero balance. What do you think?

joelalejandro:
@spersico This might be related to what we experienced today.

joelalejandro:
@rhlsthrm I agree it's something we should consider. Althought it might be more accurate to reflect that the token does not have a free balance contract deployed, it adds complexity and nuances that could be easily handled by the Node.

cf19drofxots:
Returning a zero balance seems fine. I'll modify the API.

PR #2190, opened 2019-08-15
[node] Return zero balance for token pre-deposit

cf19drofxots:
fixes: https://github.com/counterfactual/monorepo/issues/2016

### DIFF
--- a/packages/node/src/methods/app-instance/get-free-balance/controller.ts
+++ b/packages/node/src/methods/app-instance/get-free-balance/controller.ts
@@ -29,10 +29,6 @@ export default class GetFreeBalanceController extends NodeController {
 
     const stateChannel = await store.getStateChannel(multisigAddress);
 
-    const balances = stateChannel
-      .getFreeBalanceClass()
-      .withTokenAddress(tokenAddress);
-
-    return balances;
+    return stateChannel.getFreeBalanceClass().withTokenAddress(tokenAddress);
   }
 }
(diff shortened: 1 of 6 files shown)


packages/node/src/models/free-balance.ts
```diff
@@ -273,6 +282,9 @@ function serializeFreeBalanceState(
 export function convertCoinTransfersToCoinTransfersMap(
   coinTransfers: CoinTransfer[]
 ): CoinTransferMap {
+  if (!coinTransfers) {
```
joelalejandro: @cf19drofxots Instead of throwing an error, wouldn't it be best to default `coinTransfers` to `[]`?

packages/node/src/methods/app-instance/get-free-balance/controller.ts
```diff
@@ -30,14 +29,10 @@ export default class GetFreeBalanceController extends NodeController {
 
     const stateChannel = await store.getStateChannel(multisigAddress);
 
-    const ret = stateChannel
+    const balances = stateChannel
```
IIIIllllIIIIllllIIIIllllIIIIllllIIIIll: this can be returned directly

cf19drofxots: you are correct. done.

packages/node/src/models/free-balance.ts
cf19drofxots: The expectation is that this function should always be called with a list of length non-zero, so even if we default to `[]` that's a silent ignoring of whoever called this having the wrong argument value, hence the throw.

PR #2220, opened 2019-08-19
Ensure wait before confirming bytecode correctness

snario:


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
Might be good to limit the wait (https://github.com/counterfactual/monorepo/blob/master/packages/node/src/node.ts#L198)

PR #2234, opened 2019-08-20
[node] print success message

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -175,6 +175,12 @@ export default class CreateChannelController extends NodeController {
           continue;
         }
 
+        // else
+
+        if (tryCount > 0) {
+          console.log(`success on try #${tryCount}`);
```
snario: If you're going to add a log, make it more verbose and explain the context

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -175,6 +175,14 @@ export default class CreateChannelController extends NodeController {
           continue;
         }
 
+        // else
```
snario: If you're gonna write code, write code, not a comment

PR #2242, opened 2019-08-21
[contracts] Mainnet migrations

LayneHaber:
### Description

I deployed to mainnet. Copy pasted in address for the `ProxyFactory` in the `1.json`. Can also commit the artifacts there (I copy pasted output from `manual-deploy.js` to a file on my compy)

### Related issues

- [ ] Needs to work on mainnet


packages/cf-funding-protocol-contracts/manual-deploy.js
```diff
@@ -11,6 +11,8 @@ const getContractAddress = ethers.utils.getContractAddress;
 const provider = new InfuraProvider("rinkeby", process.env.INFURA_API_KEY);
 const wallet = Wallet.fromMnemonic(process.env.ETH_ACCOUNT_MNENOMIC).connect(provider);
 
+console.log(wallet.address);
```
cf19drofxots: Remove?

LayneHaber: easier to check the pending transactions when its sent from the wallet

cf19drofxots: Makes sense - there wasn't a comment or anything around it so I thought it was for debugging purposes mostly

PR #2248, opened 2019-08-21
[contracts] pass waffle flags

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
#2224 changed some waffle flags to make ProxyFactory.json more reproducible; this PR applies the same change to all other waffle configs


snario:
We could perhaps put this in the root of the monorepo and inherit from it but nbd

PR #2254, opened 2019-08-23
[greenboard] Runs E2E tests in CircleCI using a Docker Container

joelalejandro:
This PR creates a new Docker image called `counterfactual/greenboard-environment` that sets up everything we need to run E2E tests on CircleCI.


joelalejandro:
Do not merge - e0c70df is a false positive.

PR #2250, opened 2019-08-21
[monorepo] reproducible contract builds

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
https://www.notion.so/elfour/Post-Mortem-Why-did-withdrawals-not-work-0382d5a610f84c17b37e2433f030224e#0cd2367a52c645a28f77ec0901529836

- [ ] upstream the `legacyOutput` patch
- [x] commit `expected-build` and yarn commands for verifying / generating them
- [x] import `expected-build`
- [ ] switch deploy script to use `expected-build`
- [ ] delete truffle

PR #2276, opened 2019-08-26
[monorepo] commit contract build artifacts

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
This PR 
- commits the waffle build output into source control under `expected-build`
- adds a `verify-solc` command to rebuild and compare to the committed build output and fail if they differ
- runs `verify-solc` in CI
- makes the node import artifcats from `expected-build` instead of `build`

Note: deploying the output is being done separately in #2250


cf19drofxots:
https://github.com/counterfactual/monorepo/pull/2290 this effectively caused the expected bytecode (for the `MinimumViableMultisig` contract) introduced in this PR to be stale.

cf19drofxots:
@IIIIllllIIIIllllIIIIllllIIIIllllIIIIll are there steps documented somewhere as to how these expected builds are supposed to be generated?

PR #2296, opened 2019-08-28
Throw a revert if the create2 opcode fails

snario:


packages/cf-funding-protocol-contracts/contracts/proxies/ProxyFactory.sol
```diff
@@ -60,6 +60,8 @@ contract ProxyFactory {
     // solium-disable-next-line security/no-inline-assembly
     assembly {
       proxy := create2(0x0, add(0x20, deploymentData), mload(deploymentData), salt)
+      let codeSize := extcodesize(proxy)
```
cf19drofxots: Looks like `codeSize` isn't even being used?

Issue #2286, opened 2019-08-28
[cf-adjudicator-contracts] A redundant line was included unnecessarily


rhlsthrm:
From Decenter audit:

File name:​ MixinRespondToChallenge.sol
Lines 46 and 51 are basically the same and the first one makes the second one redundant. We recommend removing one of the two in order to keep the code as clean as possible.

rhlsthrm:
Fixed with #2299

PR #2299, opened 2019-08-29
[contracts] Remove duplicate line

rhlsthrm:
Feedback from audit. Quick one line fix. Addresses #2286


IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:
can you run `yarn run patch-package` before updating the expected builds

PR #2320, opened 2019-09-03
[node] Add appSeqNo to engine input parameters

snario:
This pull request adds `appSeqNo` to the input parameters of the `Install` and `InstallVirtualApp` protocols. For the `Install` protocol it is the `appSeqNo` to be used for the newly added app channel under the ledger channel between the participants. In the `InstallVirtualApp` case it is used for the virtual app that is created between the sender and responder.

This changes means that it is the caller of the engine (i.e., the `InstructionExecutor`) that is responsible for ensuring uniqueness, not the protocol.

The proposals API of the `node` now bumps the `monotonicNumProposedApps` value on the stored `StateChannel` object when making a new proposal, not when the installation is complete as it was before.

This should help with the error case of proposing two apps and then installing then in that order. There is a test for this.

This is not a perfect fix as it simply puts more responsibility on the `node` to ensure it does not propose two `appSeqNo` of the same value. There could be some code inside the protocols to ensure this does not happen by storing a `Set` of all used `appSeqNo` in the `StateChannel` object but that has not been added in this pull request.


packages/node/src/store.ts
```diff
@@ -245,7 +245,9 @@ export class Store {
     );
 
     if (!appInstanceProposal) {
-      throw Error(NO_PROPOSED_APP_INSTANCE_FOR_APP_INSTANCE_ID(appInstanceId));
+      throw new Error(
```
cf19drofxots: `new` is not needed

snario: It is the convention we follow

cf19drofxots: Do another search. We mostly don't use `new`.

PR #2329, opened 2019-09-04
Expose expected-build-artifacts

snario:


cf19drofxots:
The above comment will be done in a follow up PR to unblock this from working again.

PR #2338, opened 2019-09-05
Upgrade p-queue

snario:


kthomas:
@snario @rhlsthrm should probably prioritize merging this one

PR #2347, opened 2019-09-06
Move requestHandler.getShardedQueue to single location

snario:
Removes complexity from controllers.


snario:
Because of #2348 we should not need this

PR #2348, opened 2019-09-06
Add tests to queueing helper function

snario:
Third one fails, isolating an issue.

### DIFF
--- a/packages/node/src/methods/queued-execution.ts
+++ b/packages/node/src/methods/queued-execution.ts
@@ -19,9 +19,11 @@ export async function executeFunctionWithinQueues(
     return promise;
   }
 
+  const p: Promise<any>[] = [];
   if (queueList.length > 0) {
     for (const queue of queueList) queue.add(executeCached);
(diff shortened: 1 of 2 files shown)


LayneHaber:
fixed + added an additional test

packages/node/src/methods/queued-execution.ts
```diff
@@ -12,19 +12,20 @@ export async function executeFunctionWithinQueues(
...
-    for (const promise of promiseList) await promise;
-    return executionPromise;
+    for (const queue of queueList) queue.add(executeCached);
+    for (const queue of queueList) p.push(queue.onIdle());
+    await Promise.all(p);
+    return await executeCached();
   }
 
-  return executeCached();
+  return await f();
```
rhlsthrm: I think their convention is not to `return await <Promise>` just `return <Promise>`. Seen this in other places.

LayneHaber: @snario wanna give the changes i made and tests a once over for consistency?

PR #2349, opened 2019-09-06
Make all calls use a single global queue

snario:
This gets the `uninstall-concurrent-mixed` test to the expected case that I mentioned earlier (where both sides are in a stalemate).


snario:
Because of #2348 we should not need this

PR #2357, opened 2019-09-07
Add SafeMath for setState calls

snario:


kthomas:
This actually does not fix the overflow vulnerability. Adding `SafeMath` no longer works to protect the `finalizesAt` from overflowing because of this commit: https://github.com/counterfactual/monorepo/commit/4265a14998b50744b6ea32241b439216dc24675a

The gas optimizations in the above commit properly pack the `AppChallenge` values 👍 

But we need to make protect ourselves against overflowing `uint248` now instead of `uint256`. `SafeMath` protects us from ourselves when working with `uint256`.

kthomas:
Here is a proposed solution which does what `SafeMath` does, but for `uint248`:

```
uint248 finalizesAt = uint248(block.number + req.timeout);
require(finalizesAt >= req.timeout, "uint248 addition overflow");
challenge.finalizesAt = finalizesAt
```

kthomas:
To make things easier: https://github.com/counterfactual/monorepo/pull/2363

I recommend closing this PR.

PR #2366, opened 2019-09-09
[node] Concurrency tests

rhlsthrm:
Add tests for concurrency in various situations. `packages/node/test/integration/uninstall-concurrent-mixed.spec.ts` is still not working, and is skipped for now.


packages/node/test/integration/uninstall-concurrent-mixed.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+        case nodeB.publicIdentifier:
+          return "Node B";
+        case nodeC.publicIdentifier:
+          return "Node C";
+        default:
+          return "unknown";
+      }
+    };
+    nodeA.on(NODE_EVENTS.PROTOCOL_MESSAGE_EVENT, (msg: ProposeMessage) => {
+      console.log(
```
LayneHaber: we should probably remove all this logging before putting these tests into master

packages/node/test/integration/uninstall-concurrent-mixed.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+  makeProposeCall
+} from "./utils";
+
+expect.extend({ toBeLt });
+
+jest.setTimeout(15000);
+
+// TODO: this test isnt working bc of the concurrency issues that are being investigated now.
+// reenable this test when it works
+describe.skip("Concurrently uninstalling virtual and regular applications without issue", () => {
```
LayneHaber: do we want to merge skipped tests into master?

snario: I think it's okay as we intend for this to pass

rhlsthrm:
> Majority of these tests also have weak assertions (assertions are primarily that they will finish without throwing an explicit error. Do we want to have stronger assertions in any of these tests besides "code completes", or can we leave that to other integration/unit tests that aren't explicitly concurrency tests?

Since the other unit tests exist, I think these are fine as "weak" assertions. @LayneHaber what are the changes requested? Hesitant to remove the console logging now on the skipped test since it's being skipped anyways.

LayneHaber:
> 
> Hesitant to remove the console logging now on the skipped test since it's being skipped anyways.

sure we can leave the `console.log`s until the test isnt skipped and were done debugging

PR #2369, opened 2019-09-09
[node] Uninstall install concurrent test

LayneHaber:
### Description

Adds two additional concurrency tests:
1. A and B have an app installed. A tries to concurrently install separate app and uninstall existing app

2. A and B have an app installed. A tries to install another app while B uninstalls existing app

Both pass


LayneHaber:
closing in favor of #2385

PR #2379, opened 2019-09-10
[node] Add lock service interface to node

cf19drofxots:


rhlsthrm:
Let's remove the `x.x.x-lock` from the packages, as per @snario's agreement that the concurrency protection should be in the code.

packages/node/test/services/memory-lock-service.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class MemoryLockService implements Node.ILockService {
+  private readonly locks: Map<string, Lock> = new Map();
+  constructor() {}
+  async get(path: string): Promise<Lock> {
+    if (this.locks.has(path)) {
+      return this.locks.get(path)!;
+    }
+    return {
+      operation: "",
```
snario: What is an operation?

packages/types/src/node.ts
```diff
@@ -64,6 +64,16 @@ export namespace Node {
     reset?(): Promise<void>;
   }
 
+  export interface ILockService {
+    get(path: string): Promise<Lock>;
+    set(path: string, value: Lock): Promise<void>;
```
snario: Rename to `tryToGrabLock` or something more indicative of what is does.

Also the error case should be explicit.

packages/types/src/node.ts
```diff
@@ -64,6 +64,16 @@ export namespace Node {
     reset?(): Promise<void>;
   }
 
+  export interface ILockService {
+    get(path: string): Promise<Lock>;
```
snario: `checkIfLockExists`?

packages/node/src/methods/queued-execution.ts
```diff
@@ -8,18 +14,93 @@ import Queue, { Task } from "p-queue";
...
    * This promise will get run `n` times for `n` queues (since it
    * will be called in every queue) and so to ensure it only runs
    * once overall we memoize it.
    */
-  function runTaskWithMemoization() {
-    if (!promise) promise = task();
+  async function runTaskWithMemoization() {
+    let error;
+    let blockingQueue;
+    for (let i = 0; i < 5; i += 1) {
```
snario: 5?

packages/types/src/node.ts
kthomas: ^^ This.

Copying my suggestion for `ILock` interface from discord here:

```
export interface ILockService {
  acquireLock(timeout: number): Promise<any>;
}
```

The promise is rejected if after `timeout` millis if the lock isn't able to be acquired.

`ILock` would be implemented on objects. If we are using `ILockService` (i.e., if we add a `RedlockService` impl) perhaps the interface should look like this:
```
export interface ILockService {
  acquireLock(callback: (...args: any[]) => any, timeout: number): Promise<any>;
}
```
`callback` would be invoked if the lock was successfully acquired. The promise is rejected if after `timeout` millis if the lock isn't able to be acquired. In this case, `callback` is dropped on the floor.

packages/types/src/node.ts
kthomas: See above comment; we do not need a method to check. We should just attempt to acquire a lock and execute code if successful. Implementations can decide to attempt to acquire the lock multiple times as business logic necessitates...

kthomas: ...meaning, if they need retry logic.

packages/types/src/node.ts
kthomas: Also for clarity, the `any` retval on the `callback` in this example allows us to return stuff after we get the lock!

We should not return a `Promise<Lock>`. We should return a `Promise<any>` that returns the retval from `callback` and reject the promise if we can't acquire the lock.

packages/node/src/methods/queued-execution.ts
kthomas: I need a walkthru of this file when possible 🙏

PR #2383, opened 2019-09-10
Update yarn version on Dockerfile

snario:


snario:
Cat and mouse situation. Need to merge this in such that the [Docker Hub](https://cloud.docker.com/u/counterfactual/repository/docker/counterfactual/circleci-environment/builds) will use it, but can't merge because current one has 1.12.3. Will fix later

PR #2386, opened 2019-09-11
Add an ILockService concept to ProcessQueue

snario:
This is more of a Proof-of-Concept, I think.

It adds a locking mechanism to the `ProcessQueue` that adheres to the interface suggested by @kt. There is also an in-memory implementation of this interface that is default used by the `Node`. All the tests seem to pass which indicates to me it is a harmless addition to `master` now.

The next thing to try is to implement a version of `ILockService` that communicates to a central registry of locks, as this one is only using a local in-process registry like I mentioned. The central registry could just be an in-memory-of-the-jest-test-suite one for the next iteration. Eventually Connext intends to use Redis.

I'm not 100% confident this solves the issues we see with deadlock scenarios yet, as like I mentioned above, this is _not_ actually a central registry yet (i.e., it is just an elegant code change to the existing `ProcessQueue` object).

I think, also, that we can remove the call to `ProcessQueue::addTask` from the `handle-protocol-message.ts` file once we try to implement this in a central registry context.

### DIFF
--- a/packages/node/src/lock/lock.ts
+++ b/packages/node/src/lock/lock.ts
@@ -1,0 +1,49 @@
+import Queue from "p-queue";
+import uuid from "uuid";
+
+import { Deferred } from "../deferred";
+
+export class Lock {
+  private currentLockHandle: Deferred<any> = new Deferred();
+  private unlockKey: string = "";
+  private readonly requestsForLock: Queue;
+
+  constructor(public readonly lockName: string) {
+    this.requestsForLock = new Queue({ concurrency: 1 });
+    this.currentLockHandle.resolve();
+  }
+
+  async acquireLock(timeout: number): Promise<string> {
+    const unlockKey = uuid.v1();
+    const lockAvailableNow = this.requestsForLock.add(() => {});
+    this.requestsForLock.add(() =>
+      this.acquireLockInternal(unlockKey, timeout)
+    );
+    await lockAvailableNow;
+    return unlockKey;
+  }
+
+  async releaseLock(unlockKey: string) {
+    this.verifyLockKey(unlockKey);
+    this.currentLockHandle.resolve();
+  }
+
+  private acquireLockInternal(
+    unlockKey: string,
+    timeout: number
+  ): Promise<any> {
+    const claim = new Deferred();
+    this.currentLockHandle = claim;
+    this.unlockKey = unlockKey;
+    setTimeout(() => claim.reject("Request timed out."), timeout);
+    return claim.promise;
+  }
+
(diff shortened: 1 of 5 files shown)


packages/node/src/lock/lock.ts
```diff
@@ -0,0 +1,49 @@
+import Queue from "p-queue";
+import uuid from "uuid";
+
+import { Deferred } from "../deferred";
+
+export class Lock {
```
snario: NOTE: This is just a helper class for the `MemoryLockService` example.

packages/node/src/lock/memory-lock-service.ts
```diff
@@ -0,0 +1,47 @@
+import { Lock } from "./lock";
+import { ILockInterface } from "./types";
+
+export class MemoryLockService implements ILockInterface {
```
snario: REMINDER: This is a _local_ implementation. It is not a central lock. It's just a dummy local lock.

packages/node/src/lock/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface ILockInterface {
```
snario: As suggested by @kt, however I added `lockName`.

packages/node/src/process-queue.ts
```diff
@@ -1,9 +1,38 @@
...
+    private readonly lockingService: ILockInterface,
+    ...args: any[]
+  ) {
+    super(...args);
+  }
+
+  async add(task: Task<any>) {
+    return super.add(() =>
+      this.lockingService.acquireLock(this.queueName, task, 30_000)
+    );
```
snario: This is the important bit. It's meant to add the process to the queue, then when it gets executed locally, request the lock before continuing, otherwise error. I think this is the right order...

kthomas:
`:%s/kt/kthomas/g` 🎸 

Reviewing soon!

packages/node/src/lock/types.ts
kthomas: Ah yes, `lockName` makes good sense here. Originally I had conceptualized the interface `ILock`, which internally would supply its own lock key. In `ILockService` `lockName` would be provided as you suggest 👍

packages/node/src/lock/lock.ts
kthomas: Makes sense to have an in-memory implementation for unit tests.

kthomas: I still need to review the `Lock` object. It seems unnecessary to me conceptually but I still need to dig into that class.

PR #2388, opened 2019-09-11
Add MemoryLockService to Jest tests, begin un-skipping concurrency tests

snario:
WIP


packages/node/src/message-handling/handle-protocol-message.ts
```diff
@@ -49,24 +49,17 @@ export async function handleReceivedProtocolMessage(
     requestHandler
   );
 
-  const postProtocolStateChannelsMap = await requestHandler.processQueue.addTask(
```
snario: As mentioned, I think we can remove these, not 100% sure yet

packages/node/src/message-handling/handle-node-message.ts
```diff
@@ -29,26 +29,24 @@ export async function handleReceivedProposalMessage(
     networkContext.MinimumViableMultisig
   );
 
-  await requestHandler.processQueue.addTask([multisigAddress], async () => {
```
snario: As mentioned, I think we can remove these, not 100% sure yet

packages/node/src/methods/queued-execution.ts
```diff
@@ -34,19 +36,21 @@ export async function addToManyQueues(queues: Queue[], task: Task<any>) {
...
    */
-  const waitForEveryQueueToFinish = Promise.all(
-    queues.map(q => q.add(() => {}))
-  );
+  const deferreds = Array(queues.length).fill(new Deferred());
+  const waitForEveryQueueToFinish = Promise.all(deferreds);
 
   await Promise.all(
-    queues.map(q =>
+    queues.map((q, i) =>
```
snario: #2387 is included in this branch atm

Issue #2342, opened 2019-09-06
[node] Clean up Connext specific tests


rhlsthrm:
Address the issues in the above PR:

https://github.com/counterfactual/monorepo/pull/2339

LayneHaber:
See #2390

PR #2390, opened 2019-09-12
[node] Clean connext test

LayneHaber:
### Description

Cleans up the `connext-issue.spec.ts` file, also creates a dedicated spot to insert connext specific utils


LayneHaber:
Addresses #2342

PR #2413, opened 2019-09-13
[node] Modify tests to repro yet another issue.

rhlsthrm:


snario:
Can you remove the console logs

PR #2433, opened 2019-09-15
[node] Fix RPC router issue

rhlsthrm:
When we run our stack with local CF code (i.e. not from NPM), we see an issue sometimes where this function is accessed before the class has been initialized. This should fix those cases.


rhlsthrm:
> This isn't really a fix though. It gives false info about the listener count when the count may not actually be 0, right?

@joelalejandro also had a look with us on this, and we decided if the service is not initialized, there aren't any listeners anyways, so it's not really false info.

Issue #2435, opened 2019-09-15
[node] Symmetrical FINISHED events for all parties after protocol execution completes


rhlsthrm:
Add support for `FINISHED` events to all protocols which fire when the protocol completes. These should also be passed in a message to other channel counterparties so that everyone knows when all stores have unlocked at the same time.

- [x] `SETUP_FINISHED` @ArjunBhuptani @cf19drofxots #2437 
- [ ] `DEPOSIT_FINISHED`
- [x] `INSTALL_FINISHED` - PR https://github.com/counterfactual/monorepo/pull/2438 (@LayneHaber started it, @cf19drofxots getting it over the line)
- [x] `UNINSTALL_FINISHED` @ArjunBhuptani #2440
- [ ] `INSTALL_VIRTUAL_FINISHED`
- [ ] `UNINSTALL_VIRTUAL_FINISHED`
- [x] `WITHDRAW_FINISHED` @cf19drofxots #2441

~P.s. Listening for the INSTALL event does not yield the event from the initiator.~ resolved via #2439

rhlsthrm:
So initiatior actually does emit the event, but it emits it with the RPC method name, not the event name:

https://github.com/counterfactual/monorepo/blob/master/packages/node/src/rpc-router.ts#L48

Updated the title to reflect.

ArjunBhuptani:
`SETUP_FINISHED` PR here: https://github.com/counterfactual/monorepo/compare/setup-finished-event?expand=1

rhlsthrm:
This issue got changed a lot from the original. The original issue that I reported is this:

> So initiatior actually does emit the event, but it emits it with the RPC method name, not the event name:
>
> https://github.com/counterfactual/monorepo/blob/master/packages/node/src/rpc-router.ts#L48
>
> Updated the title to reflect.

Currently this issue as described isn't relevant, but what I mentioned here is still an issue/

PR #2437, opened 2019-09-16
[Node] `SETUP_FINISHED` event

ArjunBhuptani:
Part of the work needed to complete the meta issue: https://github.com/counterfactual/monorepo/issues/2435


packages/node/src/machine/types.ts
```diff
@@ -43,6 +44,12 @@ export type ProtocolMessage = {
   customData: { [key: string]: any };
 };
 
+export type FinMessage = {
+  toXpub: string;
+  processID: string;
+  eventName: Node.EventName;
```
LayneHaber: should this be the event name or the protocol name?

packages/node/src/node.ts
```diff
@@ -40,7 +45,7 @@ export class Node {
 
   private readonly ioSendDeferrals = new Map<
     string,
-    Deferred<NodeMessageWrappedProtocolMessage>
+    Deferred<NodeMessageWrappedProtocolMessage | NodeMessageWrappedFinMessage>
```
LayneHaber: the typing seems like it could be combined into a declared union type within the types file, but if were only using it here it doesn't really make too much of a difference. something to keep in mind if we end up using this in other places though

packages/node/src/node.ts
```diff
@@ -338,10 +388,15 @@ export class Node {
     const isProtocolMessage = (msg: NodeTypes.NodeMessage) =>
       msg.type === NODE_EVENTS.PROTOCOL_MESSAGE_EVENT;
 
-    const isExpectingResponse = (msg: NodeMessageWrappedProtocolMessage) =>
-      this.ioSendDeferrals.has(msg.data.processID);
+    const isFinishedMessage = (msg: NodeTypes.NodeMessage) =>
+      msg.type.endsWith("FinishedEvent");
+
+    const isExpectingResponse = (
+      msg: NodeMessageWrappedProtocolMessage | NodeMessageWrappedFinMessage
```
LayneHaber: okay looks like its being used in a few places, so may be worth adding this as a union type. i can make the change in follow on PRs though so its not blocking, but dont want the types package to have to be published twice

packages/node/src/node.ts
```diff
@@ -338,10 +388,15 @@ export class Node {
     const isProtocolMessage = (msg: NodeTypes.NodeMessage) =>
       msg.type === NODE_EVENTS.PROTOCOL_MESSAGE_EVENT;
```
LayneHaber: wont this return true for a `NodeMessageWrappedFinMessage` as well? Ref:

```javascript
await this.messagingService.send(to, {
        data,
        from: fromXpub,
        type: NODE_EVENTS.PROTOCOL_MESSAGE_EVENT
      } as NodeMessageWrappedFinMessage);
    });
```

cf19drofxots: Yeah it will - the `FinMessage` can be thought of as a very specific type of `ProtocolMessage` as its purpose is to conclude a protocol, but its definition is different from the data that gets included in a `ProtocolMessage`

packages/node/src/machine/types.ts
cf19drofxots: Event name. We don't care about the protocol name here because the protocol for which this event is being sent around is inherently linked to the `processID`.

PR #2438, opened 2019-09-17
[node] Implements emission of INSTALL_FINISHED event across parties

LayneHaber:
### Description

Initiator sends message to counterparty in install to emit event once the protocol has finished for all involved parties.

### Related issues

#2435

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@counterfactual/node",
-  "version": "0.2.76",
+  "version": "0.2.77",
   "main": "dist/index.js",
   "iife": "dist/index.iife.js",
   "types": "dist/src/index.d.ts",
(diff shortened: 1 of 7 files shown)


LayneHaber:
NOTE: these opcodes were inspired by bugs noticed by the connext team. specifically, the protocols would finish for the responders, they would go offline, and the initiator would emit `IO_SEND_AND_WAIT` errors.

This may *not* be the best way to address the problem. (Why is the initiator timing out anyways? If the protocol is done for the responder, it should be okay for them to go offline.)

I am currently testing that the theory that this has to do with the NATS messaging implementation used by the client/hub (something is queued but not actually sent, or theres some asynchronous delay there). Don't merge this PR until it's verified this is not a NATS issue.

cf19drofxots:
@LayneHaber 

> If the protocol is done for the responder, it should be okay for them to go offline.

The current implementation of the protocol takes this into consideration.

PR #2442, opened 2019-09-17
Flip event emission order

ArjunBhuptani:
Minor fix to stop potential race

### DIFF
--- a/packages/node/src/node.ts
+++ b/packages/node/src/node.ts
@@ -175,13 +175,13 @@ export class Node {
       const fromXpub = this.publicIdentifier;
       const to = data.toXpub;
 
-      this.requestHandler.outgoing.emit(data.eventName, data);
-
       await this.messagingService.send(to, {
         data,
         from: fromXpub,
         type: NODE_EVENTS.PROTOCOL_MESSAGE_EVENT
       } as NodeMessageWrappedFinMessage);
+
+      this.requestHandler.outgoing.emit(data.eventName, data);
     });
 
     protocolRunner.register(Opcode.IO_SEND, async (args: [ProtocolMessage]) => {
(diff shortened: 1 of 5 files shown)


packages/node/src/node.ts
```diff
@@ -175,13 +175,13 @@ export class Node {
...
 
-      this.requestHandler.outgoing.emit(data.eventName, data);
-
       await this.messagingService.send(to, {
         data,
         from: fromXpub,
         type: NODE_EVENTS.PROTOCOL_MESSAGE_EVENT
       } as NodeMessageWrappedFinMessage);
+
+      this.requestHandler.outgoing.emit(data.eventName, data);
```
cf19drofxots: How would this change prevent a potential race?

ArjunBhuptani: Maybe race is the wrong word:

Right now, it seems like the bot farm responder shuts off when it hears the `installFinalEvent` which is *before* it's actually sending the final message to the initiator. Just trying this out to see if it helps fix the issue.

packages/node/src/node.ts
```diff
@@ -277,7 +277,7 @@ export class Node {
 
       if (!msg || !("data" in (msg as NodeMessageWrappedFinMessage))) {
         throw Error(
-          `IO_SEND_AND_WAIT timed out after 30s waiting for counterparty reply in ${data.eventName}`
+          `IO_WAIT timed out after 30s waiting for counterparty reply in ${data.eventName}`
```
LayneHaber: we have to change these timeouts or these logs at some point. its waiting 60s

packages/node/test/integration/uninstall.spec.ts
```diff
@@ -174,16 +174,23 @@ describe("Node A and B install apps of different outcome types, then uninstall t
...
       let nodeBProcessID;
 
+      const verifyId = () => {
+        if (nodeAProcessID && nodeBProcessID) {
+          expect(nodeBProcessID).toEqual(nodeAProcessID);
+          done();
+        }
+      };
+
       nodeA.on(NODE_EVENTS.UNINSTALL_FINISHED, async (msg: FinMessage) => {
```
LayneHaber: is ordering important here? failed the previous test because it emits on nodeB then nodeA. the changes i made the tests un-opinionated on this

PR #2444, opened 2019-09-20
Revert _FINISHED protocol messages

snario:


packages/cf-adjudicator-contracts/package.json
```diff
@@ -32,7 +32,7 @@
     "solidity"
   ],
   "devDependencies": {
-    "@counterfactual/types": "0.0.40",
+    "@counterfactual/types": "0.0.39",
```
cf19drofxots: Change this to `0.0.41` - if it's `0.0.39` then the next change that bumps this to `0.0.40` will conflict with the current published npm package.

packages/node/src/methods/app-instance/propose-install/controller.ts
```diff
@@ -53,6 +47,12 @@ export default class ProposeInstallController extends NodeController {
     params: Node.ProposeInstallParams
   ) {
     const { store, publicIdentifier, networkContext } = requestHandler;
+    const { initialState } = params;
```
cf19drofxots: This change should not be reverted as we wouldn't want to queue an invalid request.

snario: It should be in `beforeExecution`

cf19drofxots: That doesn't make sense though - why would you queue a request that's invalid to begin with?

snario: `beforeExecution` runs before the queue is added

PR #2445, opened 2019-09-20
[node] Replace 'ProtocolRunner' with 'Engine'

cf19drofxots:
This is so that the codebase is consistent with upcoming documentation that will reference specific code blocks referring to how protocols are executed.


snario:
Actually the only thing I have against this is that as of right now, Engine is not implemented by this class. This class does something different than what Engine specifies.

cf19drofxots:
> Engine is not implemented by this class

Wouldn't this count as version 1? With version 2 coming?

cf19drofxots:
Closing this because the classes the implement the protocol execution logic don't reflect what the `Engine` actually does.

PR #2448, opened 2019-09-20
Remove proposals from Store DB

snario:
The store should just be a blob you can save to disk and retrieve when necessary and in-memory during node operation. This begins that work.

### DIFF
--- a/packages/node/src/store.ts
+++ b/packages/node/src/store.ts
@@ -91,7 +91,8 @@ export class Store {
         return sc.multisigAddress;
       }
     }
-    return ""; // FIXME
+
+    throw new Error(NO_MULTISIG_FOR_APP_INSTANCE_ID);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


packages/node/src/store.ts
```diff
@@ -85,9 +82,16 @@ export class Store {
...
+    for (const sc of (await this.getStateChannelsMap()).values()) {
+      if (
+        sc.proposedAppInstances.has(appInstanceId) ||
+        sc.appInstances.has(appInstanceId) ||
+        (sc.hasFreeBalance && sc.freeBalance.identityHash === appInstanceId)
+      ) {
+        return sc.multisigAddress;
+      }
+    }
+    return ""; // FIXME
```
cf19drofxots: Don't return empty string here - throw instead if there's no multisig address to be found for some app instance ID.

Issue #2065, opened 2019-07-30
[node] Deploy multisig on withdrawal


LayneHaber:
Deploy the multisig on withdrawal instead of deposit

PR #2467, opened 2019-09-25
[node] Deploy multisig on withdraw

rhlsthrm:
Addresses issue #2065.

### DIFF
--- a/packages/node/src/methods/state-channel/create/controller.ts
+++ b/packages/node/src/methods/state-channel/create/controller.ts
@@ -40,9 +40,6 @@ export default class CreateChannelController extends NodeController {
       networkContext.MinimumViableMultisig
     );
 
-    // always will be HashZero since we are deploying on withdrawal
-    const tx = { hash: HashZero } as TransactionResponse;
-
     // Check if the database has stored the relevant data for this state channel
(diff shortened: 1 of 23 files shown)


snario:
Maybe a better strategy would be to make a separate endpoint for deploying the multisig independently? Then, just alias on the Connext client a method `DEPLOY_AND_WITHDRAW` whcih runs them in sequence.

packages/node/src/methods/state-channel/create/controller.ts
```diff
@@ -50,39 +31,28 @@ export default class CreateChannelController extends NodeController {
...
     const multisigAddress = getCreate2MultisigAddress(
       owners,
       networkContext.ProxyFactory,
       networkContext.MinimumViableMultisig
     );
 
-    // By default, if the contract has been deployed and
-    // DB has records of it, controller will return HashZero
-    let tx = { hash: HashZero } as TransactionResponse;
+    // always will be HashZero since we are deploying on withdrawal
```
cf19drofxots: Alternatively we can update the API for this method.

IMO I don't think backward compatibility is a concern atm.

rhlsthrm: API updated, this line is deleted.

PR #2487, opened 2019-09-29
Adds a PROPOSE protocol to replace Node's propose mechanic

snario:
The Node handled proposals before, but now it is is something that the `protocolRunner` manages.


snario:
This was a first step towards making the `store` in-memory and easily extensible for adding things like an adjudicator cache such that the challenge methods are easy to implement.

PR #2517, opened 2019-10-18
Make CounterfactulApp methods public

snario:
To implement `CounterfactualAdapterApp` in the `nitro-protocol` package on `@statechannels/monorepo` I need to be able to inherit from `CounterfactualApp`, so these methods must be public.


Alonski:
@snario Can this be merged and released?

PR #2563, opened 2019-10-28
Basic approach to one off migration

snario:


rhlsthrm:
This looks good. I am currently testing out the migration process to make sure the newly migrated channels work properly. Can we export this function from the node? I'd like to be able to import the function into a script we can call from the CLI.

packages/node/src/store.ts
```diff
@@ -15,6 +15,7 @@ import {
 import { prettyPrintObject } from "./utils";
 
 interface SharedData {
+  version: 1; // TODO: Add better versioning & migrations tooling
```
cf19drofxots: From looking around it seems like there are various ways of versioning a db schema.

Though I'm kinda confused as to why the `version` number is being put in `SharedData`? I'm also confused how this handles migrations overall if there are no typed schemas between the versions?

Maybe we should have a separate blob that holds both the version number and the schema at the corresponding version number. The migration script would then just diff `n-1` and `n` to apply the migration.

snario: `StoreData = { version: number, data: SharedData }` ?

PR #2566, opened 2019-10-29
Consolidate some utility files

snario:


packages/node/src/utils/create2-address.ts
```diff
@@ -1,27 +1,13 @@
 import MinimumViableMultisig from "@counterfactual/cf-funding-protocol-contracts/expected-build-artifacts/MinimumViableMultisig.json";
```
cf19drofxots: So this becomes a file _just_ for calculating a create2 address? Seems excessive to have a file just for this and probably more appropriate to put this in a `utils` file (as it was before).

snario: I think it's fine to have this in its own file, it's sufficiently unique compared to the other utility functions which are pretty basic things that don't relate to this at all.

PR #2586, opened 2019-11-07
[node] proposed app removal

LayneHaber:
### Description

Apps were not being removed from the proposed apps list in the state channel when the protocol completed for the non-initiating party.

Fixed by defaulting to original accessing of `appInstanceId`, but adding another check in case it was in the `params` of the `data` obj.

May want to clean up the casting, or the added place for the test stubs, but functional

### DIFF
--- a/packages/node/test/integration/install.spec.ts
+++ b/packages/node/test/integration/install.spec.ts
@@ -15,10 +15,10 @@ import {
   getAppContext,
   getBalances,
   getInstalledAppInstances,
+  getProposedAppInstances,
   makeAndSendProposeCall,
   makeInstallCall,
-  transferERC20Tokens,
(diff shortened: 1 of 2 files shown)


packages/node/test/integration/install.spec.ts
```diff
@@ -62,12 +63,27 @@ describe("Node method follows spec - install", () => {
           makeInstallCall(nodeB, msg.data.appInstanceId);
         });
 
+        // FIXME: still no symmetric events -- nodeB will never emit an
```
snario: This event should emit

rhlsthrm: It does not at the moment, this test fails if this is commented out.

PR #2590, opened 2019-11-07
Remove getCreate2MultisigAddress call in handle-protocol-message

snario:


LayneHaber:
Bump versions on node

